Issue #6, opened 2024-06-17
Storage Layer implementations


patflynn:
- [x] #21
- [x] #22
- [x] #23
- [x] #24

PR #43, opened 2024-07-09
Add queue mix-in for storage implementations

AlCutter:
This PR adds a buffering queue which storage implementations may use to batch up calls to their Add methods.

If multiple identical entries are added to the queue, it will silently de-deduplicate them and return the same assigned index (or error) for each.

Supports #6


AlCutter:
GH auto-closed this for reasons best known to itself, re-raised in #44

PR #44, opened 2024-07-09
Add queue mix-in for storage implementations

AlCutter:
This PR adds a buffering queue which storage implementations may use to batch up calls to their Add methods.

If multiple identical entries are added to the queue, it will silently de-deduplicate them and return the same assigned index (or error) for each.

Supports https://github.com/transparency-dev/trillian-tessera/issues/6


storage/queue.go
```diff
@@ -0,0 +1,157 @@
...
+	inFlightMu sync.Mutex
+	inFlight   map[string]*entry
+}
+
+// IndexFunc is a function which returns an assigned log index, or an error.
+type IndexFunc func() (idx uint64, err error)
+
+// FlushFunc is the signature of a function which will receive the slice of queued entries.
+// It should return the index assigned to the first entry in the provided slice.
+type FlushFunc func(ctx context.Context, entries [][]byte) (index uint64, err error)
```
phbnf, line 50: This assumes that all the entries in a queue will be sequenced one after other one. If the flush function does some deduplication, this property might not hold. Could you return an index for each entry instead please?

storage/queue.go
```diff
@@ -0,0 +1,157 @@
...
+	if err := q.buf.Push(entry); err != nil {
+		entry.assign(0, err)
+		close(entry.c)
+	}
+	return entry.index
+}
+
+// doFlush handles the queue flush, and sending notifications of assigned log indices.
+//
+// To prevent blockin the queue longer than necessary, the notifications happen in a
```
phbnf, line 101: nit, blocking

storage/queue.go
```diff
@@ -0,0 +1,157 @@
...
+// When the buffered queue grows past a defined size, or the age of the oldest entry in the
+// queue reaches a defined threshold, the queue will call a provided FlushFunc with
+// a slice containing all queued entries in the same order as they were added.
+//
+// If multiple identical entries are added to the queue between flushes, the queue will deduplicate them by
+// passing only the first through to the FlushFunc, and returning the index assigned to that entry to all
+// duplicate add calls.
+// Note that this deduplication only applies to "in-flight" entries currently in the queue; entries added
+// after a flush will not be deduped against those added before the flush.
+type Queue struct {
```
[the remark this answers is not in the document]
AlCutter: I think of those as "sequenced entries", there's a process migrates them to a different storage and integrates them into the tree but I'm not really thinking of that as a queue.

I don't want to call this struct `Batch` since that's not what it is, I thought about using `Pool` or `Buffer` but those names lack the connotation of FIFO which `Queue` carries.

storage/queue.go
AlCutter: That is a requirement for using this mix-in, yup.

If there's a reason why a given storage implementation cannot meet this requirement then it's fine for it to do its own queueing, but we're explicitly targeting dedup as "best effort" for performance reasons, so I'd expect it should be fine (unless there's a storage implementation which physically _cannot_ store dupes?)

Returning an index for each entry is something I'd rather avoid if possible (what would the API look like - return a `map[string([]byte)]uint64` and force the caller to iterate over it and reconcile what's happened to the missing entries?)

phbnf: You could return a map, or []uint64, that is the same length that entries is. It still allows deduplication as best effort, but has the merit of pushing all the sequencing logic to the flush function rather than then computing the index for each entry after Flush has returned.

AlCutter: I think we should wait until there's actually a concrete storage implementation which _requires_ this before making that change - this is an internal component and we're nowhere near a tagged release yet so it'll be easy to make.

storage/queue.go
```diff
@@ -0,0 +1,157 @@
...
+	buf   *buffer.Buffer
+	flush FlushFunc
+
+	inFlightMu sync.Mutex
+	inFlight   map[string]*entry
```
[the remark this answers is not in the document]
phbnf: SGTM, thanks for trying!

storage/queue.go
phbnf: Al and I had a long conversation over VC about this.

Summary:
I ran into scaling problems because of deduplication twice already, both with DynamoDB and Aurora. We need to do more tests to see how expensive, scalable and fast it is to do deduplication using S3/GCS for CT deployments. In the interest of moving forward let's keep deduplication in S3/GCS for now, even if we know that it will have a hit on latency, which means that batch deduplication is not a thing for now, and come back to this later. I have both S3 and Aurora deduplication experimental code ready, I'll try and run both with the same QPS to see what's the cost difference over a few hours.

phbnf (approved):

PR #45, opened 2024-07-09
Add outline of GCP spanner implementation

AlCutter:
This PR adds Spanner support to the GCP storage implementation, along with sequencing support.

TODO: relies on #44, merge & rebase.

Towards #23

### DIFF
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -38,10 +38,12 @@ import (
 	"net/http"
 	"os"
 	"slices"
+	"time"
 
 	"cloud.google.com/go/spanner"
 	gcs "cloud.google.com/go/storage"
 	"github.com/transparency-dev/trillian-tessera/api/layout"
+	"github.com/transparency-dev/trillian-tessera/storage"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
 	"google.golang.org/grpc/codes"
@@ -55,14 +57,22 @@ type Storage struct {
 	projectID string
 	bucket    string
 
-	dbPool   *spanner.Client
-	objStore objStore
+	dbPool    *spanner.Client
+	sequencer sequencer
+	objStore  objStore
+
+	queue *storage.Queue
 }
 
 // objStore describes a type which can store and retrieve objects.
 type objStore interface {
 	getObject(ctx context.Context, obj string) ([]byte, int64, error)
 	setObject(ctx context.Context, obj string, data []byte, cond *gcs.Conditions) error
+}
+
+// coord describes a type which knows how to sequence entries.
+type sequencer interface {
+	assignEntries(ctx context.Context, entries [][]byte) (uint64, error)
 }
 
 // Config holds GCP project and resource configuration for a storage instance.
@@ -85,19 +95,20 @@ func New(ctx context.Context, cfg Config) (*Storage, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to create GCS storage: %v", err)
 	}
-
-	dbPool, err := spanner.NewClient(ctx, cfg.Spanner)
+	seq, err := newSpannerSequencer(ctx, cfg.Spanner)
 	if err != nil {
-		klog.Exitf("Failed to connect to Spanner: %v", err)
+		return nil, fmt.Errorf("failed to create Spanner sequencer: %v", err)
 	}
 
 	r := &Storage{
 		gcsClient: c,
 		projectID: cfg.ProjectID,
 		bucket:    cfg.Bucket,
-		dbPool:    dbPool,
 		objStore:  gcsStorage,
+		sequencer: seq,
 	}
+	// TODO(al): make queue options configurable:
+	r.queue = storage.NewQueue(time.Second, 256, r.sequenceEntries)
 
 	if err := r.initDB(ctx); err != nil {
 		return nil, fmt.Errorf("failed to init DB: %v", err)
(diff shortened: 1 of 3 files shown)


storage/gcp/gcp.go
```diff
@@ -154,58 +245,53 @@ func (s *Storage) initDB(ctx context.Context) error {
...
-func (s *Storage) setTile(ctx context.Context, level, index uint64, tile [][]byte) error {
-	tileSize := uint64(len(tile))
-	klog.V(2).Infof("StoreTile: level %d index %x ts: %x", level, index, tileSize)
-	if tileSize == 0 || tileSize > 256 {
-		return fmt.Errorf("tileSize %d must be > 0 and <= 256", tileSize)
+// Entries are allocated contiguous indices, in the order in which they appear in the entries parameter.
+// This is achieved by storing the passed-in entries in the Seq table in Spanner, keyed by the
+// index assigned to the first entry in the batch.
+func (s *spannerSequencer) assignEntries(ctx context.Context, entries [][]byte) (uint64, error) {
+	// Flatted the entries into a single slice of bytes which we can store in the Seq.v columne
```
phbnf, line 254: nit column

storage/gcp/gcp.go
```diff
@@ -154,58 +245,53 @@ func (s *Storage) initDB(ctx context.Context) error {
...
+		}
 
-// getTile returns the tile at the given tile-level and tile-index for the specified log size, or
-// an error if it doesn't exist.
-func (s *Storage) getTile(ctx context.Context, level, index, logSize uint64) ([][]byte, error) {
-	tileSize := layout.PartialTileSize(level, index, logSize)
+		next := uint64(next) // Shadow next with a uint64 version of the same value to save on casts.
+		m := []*spanner.Mutation{
+			// Insert our newly sequenced batch of entries into Seq,
+			spanner.Insert("Seq", []string{"id", "seq", "v"}, []interface{}{0, int64(next), data}),
```
phbnf, line 279: Would it be worth splitting data bundles, and aligning their end on a multiple of 256?
I expect that integration will work in batches as well "integrate X bundles at most" rather than "integrate all available bundles in one go"? In which case, splitting data bundles will allow to align integration with full leaf bundles and limit the number of partial leaf bundles / L0 tiles.

phbnf: Alternatively - did you try sequencing entries into bundles directly?

AlCutter: Yeah, that's (aligning to tile size) worth exploring, I'll pop a TODO to come back and try that out later when we have something working.

Sequencing to bundles is something I tried very early on, the main problem you run into with that approach (on a non-transactional storage), is that given a tree of size N, it's very likely that two frontends will want to add a different number of leaves to the sequenced bundles at the same time. At which point you can't rely on GCS preconditions to save you, since the objects have different keys (because partial suffix), so then you need to either a) somehow create a global lock, or b) rely on some transactional infra to manage the sequencing.

Each has trade-offs, for now a transactional "coordination" is the right way to go (for scale), although I am keeping a back-of-mind thread thinking about how to also scale down for "Tessera serverless" use cases.

storage/gcp/gcp.go
```diff
@@ -85,25 +96,95 @@ func New(ctx context.Context, cfg Config) (*Storage, error) {
...
+// The location to which the tile is written is defined by the tile layout spec.
+func (s *Storage) setTile(ctx context.Context, level, index uint64, tile [][]byte) error {
+	tileSize := uint64(len(tile))
+	klog.V(2).Infof("StoreTile: level %d index %x ts: %x", level, index, tileSize)
+	if tileSize == 0 || tileSize > 256 {
+		return fmt.Errorf("tileSize %d must be > 0 and <= 256", tileSize)
+	}
+	t := slices.Concat(tile...)
+
+	// Pass an empty rootDir since we don't need this concept in GCS.
```
phbnf, line 136: I'd just remove this, we can't even see where the emptyness is, this is only relevant if you've worked on a different implementation before.

storage/gcp/gcp.go
```diff
@@ -85,25 +96,95 @@ func New(ctx context.Context, cfg Config) (*Storage, error) {
...
+
+	return s.objStore.setObject(ctx, tPath, t, &gcs.Conditions{DoesNotExist: true})
+}
+
+// getTile returns the tile at the given tile-level and tile-index for the specified log size, or
+// an error if it doesn't exist.
+func (s *Storage) getTile(ctx context.Context, level, index, logSize uint64) ([][]byte, error) {
+	tileSize := layout.PartialTileSize(level, index, logSize)
+
+	// Pass an empty rootDir since we don't need this concept in GCS.
```
phbnf, line 147: Same here

storage/gcp/gcp.go
```diff
@@ -85,25 +96,95 @@ func New(ctx context.Context, cfg Config) (*Storage, error) {
...
+	// TODO(al): make queue options configurable:
+	r.queue = storage.NewQueue(time.Second, 256, r.sequencer.assignEntries)
 
-	if err := r.initDB(ctx); err != nil {
-		return nil, fmt.Errorf("failed to init DB: %v", err)
+	return r, nil
+}
+
+// tileSuffix returns either the empty string or a tiles API suffix based on the passed-in tile size.
+func tileSuffix(s uint64) string {
```
phbnf, line 118: Should this belong in layout? Can be moved later, in which case add a TODO.

phbnf (approved):

storage/gcp/gcp.go
AlCutter: oops, that shouldn't be there, ta.

storage/gcp/gcp.go
AlCutter: same

storage/gcp/gcp.go
AlCutter: Yeah, this'll go away when Martin's API stuff is in.

PR #49, opened 2024-07-10
Added client for tlog-tiles

mhutchinson:
Added client for tlog-tiles logs.

This was largely copied from serverless-log and minimally changed to
bind this to the tlog-tiles API which has a few differences from the
serverless API. The obvious changes are the URL paths, but that is
largely hidden behind the `layout` API. The more significant delta is
that serverless returns fully populated tiles, where tlog-tiles returns
only the leaf nodes and thus the client needs to perform some
repopulation of the internal nodes in order to get at all node hashes
for proof generation.

The test data is largely the same as the serverless-log test data in
shape, though it needed to be completely regenerated due to the above
changes. This required the posix log implementation to be ported across
to the tlog-tiles API too. I have this in a local branch and will get it
up in its own PR after this one.

This client is intended to be a minimal foundation on which we can build.
It has known inefficiencies such as only returning a single leaf from an 
entry bundle, and not storing the results of computed internal hashes
within a tile. These will be addressed as I write the hammer tool that uses
this client.

Towards #66.


AlCutter:
Would it be possible to insert a first "import" commit that just copies over the original code verbatim, and then rebase this commit on top? It'll make it a lot easier to spot the diffs looking at it commit-by-commit

mhutchinson:
> Would it be possible to insert a first "import" commit that just copies over the original code verbatim, and then rebase this commit on top? It'll make it a lot easier to spot the diffs looking at it commit-by-commit

Done. Right on the limits of my git-fu, but it's done.

AlCutter (approved):
Discussed IRL: this is intended to be a quick client which unblocks the hammer/larger testing rather than an optimised client. The intention is for us to double back at a later date to improve it.

mhutchinson:
> Discussed IRL: this is intended to be a quick client which unblocks the hammer/larger testing rather than an optimised client. The intention is for us to double back at a later date to improve it.

Yeah and this is now clear in this PR description and in #66, which will stay open until the known rough edges are resolved.

PR #50, opened 2024-07-10
Add read tile in MySQL storage implementation

roger2hk:
https://github.com/transparency-dev/trillian-tessera/issues/21

### DIFF
--- a/cmd/example-mysql/main.go
+++ b/cmd/example-mysql/main.go
@@ -121,6 +121,7 @@ func main() {
 }
 
 // parseTileLevelIndex takes level and index in string, validates and returns them in uint64.
+// Note that this doesn't return any of the partial tile component.
 //
 // Examples:
 // "/tile/0/x001/x234/067" means level 0 and index 1234067 of a full tile.
--- a/cmd/example-mysql/main.go
+++ b/cmd/example-mysql/main.go
@@ -73,10 +73,10 @@ func main() {
 	})
 
 	http.HandleFunc("GET /tile/{level}/{index...}", func(w http.ResponseWriter, r *http.Request) {
-		level, index, err := parseTileLevelIndex(r.PathValue("level"), r.PathValue("index"))
+		level, index, width, err := parseTileLevelIndexWidth(r.PathValue("level"), r.PathValue("index"))
 		if err != nil {
 			w.WriteHeader(http.StatusBadRequest)
-			if _, werr := w.Write([]byte(err.Error())); werr != nil {
+			if _, werr := w.Write([]byte(fmt.Sprintf("Malformed URL: %s", err.Error()))); werr != nil {
 				klog.Errorf("/tile/{level}/{index...}: %v", werr)
 			}
 			return
@@ -94,7 +94,12 @@ func main() {
 			return
 		}
 
-		w.Header().Set("Cache-Control", "public, max-age=31536000, immutable")
+		// Only add the immutable Cache-Control header when the number of hash in the returned tile match the requested tile width.
+		// This ensures the response will not be cached when returning a partial tile on a full tile request.
+		if len(tile)/32 == int(width) {
+			w.Header().Set("Cache-Control", "public, max-age=31536000, immutable")
+		}
+
 		if _, err := w.Write(tile); err != nil {
 			klog.Errorf("/tile/{level}/{index...}: %v", err)
 			return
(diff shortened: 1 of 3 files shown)


cmd/example-mysql/main.go
```diff
@@ -87,3 +119,33 @@ func main() {
...
+// "/tile/0/x001/x234/067" means level 0 and index 1234067 of a full tile.
+// "/tile/0/x001/x234/067.p/8" means level 0, index 1234067 and width 8 of a partial tile.
+func parseTileLevelIndex(level, index string) (uint64, uint64, error) {
+	l, err := strconv.ParseUint(level, 10, 64)
+	if err != nil {
+		return 0, 0, fmt.Errorf("malformed URL: failed to parse tile level")
+	}
+
+	var i uint64
+	switch splittedIndexPath := strings.Split(index, "/"); len(splittedIndexPath) {
```
mhutchinson, line 135: nit: splitted isn't a word because English is weird. `I split the path` can be the past tense or the present tense. 🙃 

`splitIndexPath` works I think? Or `indexParts` to shorten further.

cmd/example-mysql/main.go
```diff
@@ -87,3 +119,33 @@ func main() {
 		klog.Exitf("ListenAndServe: %v", err)
 	}
 }
+
+// parseTileLevelIndex takes level and index in string, validates and returns them in uint64.
+//
+// Examples:
+// "/tile/0/x001/x234/067" means level 0 and index 1234067 of a full tile.
+// "/tile/0/x001/x234/067.p/8" means level 0, index 1234067 and width 8 of a partial tile.
```
mhutchinson, line 127: Add a comment that this doesn't return any of the partial tile component?

cmd/example-mysql/main.go
```diff
@@ -69,6 +72,35 @@ func main() {
...
+				w.WriteHeader(http.StatusNotFound)
+				return
+			}
+
+			klog.Errorf("/tile/{level}/{index...}: %v", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		w.Header().Set("Cache-Control", "public, max-age=31536000, immutable")
```
mhutchinson, line 97: This could be dangerous here; if someone requests a partial tile without the partial index, you will return the partial tile. When this tile grows and eventually becomes full, someone going via the cache will not be able to see it.

I think we need to look at only setting this header when the number of leaves exactly matches the tile that was requested (i.e. a partial tile was requested and exactly that many leaves were present, or a full tile was requested and we returned 256 leaves).

This should be calculable because we know that the hashes are 32 bytes long, so we can determine here how many hashes are in the returned tile. You'll have to make the path parser pull out the partial tile size though and return it (I would consider making it return the number of expected hashes, so return 256 if there is no partial specification).

cmd/example-mysql/main.go
roger2hk: Done.

cmd/example-mysql/main.go
roger2hk: Ooops! Updated to `indexParts`.

cmd/example-mysql/main.go
roger2hk: Good catch. PTAL, thanks!

cmd/example-mysql/main.go
```diff
@@ -120,19 +125,18 @@ func main() {
...
 
-// parseTileLevelIndex takes level and index in string, validates and returns them in uint64.
-// Note that this doesn't return any of the partial tile component.
+// parseTileLevelWidthIndex takes level and index in string, validates and returns the level, index and width in uint64.
 //
 // Examples:
 // "/tile/0/x001/x234/067" means level 0 and index 1234067 of a full tile.
 // "/tile/0/x001/x234/067.p/8" means level 0, index 1234067 and width 8 of a partial tile.
-func parseTileLevelIndex(level, index string) (uint64, uint64, error) {
+func parseTileLevelIndexWidth(level, index string) (uint64, uint64, uint64, error) {
```
mhutchinson, line 133: Can we have some unit tests for this? I'll approve this and leave it up to you whether you want to put them in this PR, or a follow-up.

mhutchinson (approved):

cmd/example-mysql/main.go
roger2hk: Sure, let me do that in a follow-up pull request today.

Issue #21, opened 2024-06-17
vanilla MySQL storage implementation


patflynn:
- [x] Add sequenced entries to entry bundles.
- [x] Refactor the way to fully utilise the MySQL for fetching multiple tiles in one query with the same ordering.
- [x] Add Go tests.
- [x] Add integration tests.

PR #56, opened 2024-07-10
Fix incorrect `parseTileLevelIndexWidth` implementation and add tests

roger2hk:
#21

### DIFF
--- a/cmd/example-mysql/main.go
+++ b/cmd/example-mysql/main.go
@@ -132,32 +132,34 @@ func main() {
 // "/tile/0/x001/x234/067.p/8" means level 0, index 1234067 and width 8 of a partial tile.
 func parseTileLevelIndexWidth(level, index string) (uint64, uint64, uint64, error) {
 	l, err := strconv.ParseUint(level, 10, 64)
-	if err != nil {
+	if l > 63 || err != nil {
 		return 0, 0, 0, fmt.Errorf("failed to parse tile level")
 	}
 
-	var i, w uint64
-	switch indexPaths := strings.Split(index, "/"); len(indexPaths) {
-	// Full tile = 3
-	// Partial tile = 4
-	case 3, 4:
-		indexPath := strings.Join(indexPaths[0:3], "")
-		indexPath = strings.ReplaceAll(indexPath, "x", "")
-		indexPath = strings.ReplaceAll(indexPath, ".p", "")
-		i, err = strconv.ParseUint(indexPath, 10, 64)
-		if err != nil {
+	i := uint64(0)
+	w := uint64(256)
+	indexPaths := strings.Split(index, "/")
+
+	if strings.Contains(index, ".p") {
+		w, err = strconv.ParseUint(indexPaths[len(indexPaths)-1], 10, 64)
+		if err != nil || w < 1 || w > 255 {
 			return 0, 0, 0, fmt.Errorf("failed to parse tile index")
 		}
+		indexPaths[len(indexPaths)-2] = strings.TrimSuffix(indexPaths[len(indexPaths)-2], ".p")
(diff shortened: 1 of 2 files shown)


cmd/example-mysql/main.go
```diff
@@ -132,33 +132,35 @@ func main() {
 // "/tile/0/x001/x234/067.p/8" means level 0, index 1234067 and width 8 of a partial tile.
 func parseTileLevelIndexWidth(level, index string) (uint64, uint64, uint64, error) {
 	l, err := strconv.ParseUint(level, 10, 64)
-	if err != nil {
+	if l > 63 || err != nil {
```
mhutchinson, line 135: Maybe I should have had more caffeine this morning, but why 63? A comment or a commented constant would be useful.

cmd/example-mysql/main.go
```diff
@@ -132,33 +132,35 @@ func main() {
...
-	switch indexPaths := strings.Split(index, "/"); len(indexPaths) {
-	// Full tile = 3
-	// Partial tile = 4
-	case 3, 4:
-		indexPath := strings.Join(indexPaths[0:3], "")
-		indexPath = strings.ReplaceAll(indexPath, "x", "")
-		indexPath = strings.ReplaceAll(indexPath, ".p", "")
-		i, err = strconv.ParseUint(indexPath, 10, 64)
-		if err != nil {
+	i := uint64(0)
```
mhutchinson, line 139: nit: move this declaration down to above for `for` loop on 156 to increase locality? I was reading the loop down there and then had to search for where `i` was defined.

cmd/example-mysql/main.go
```diff
@@ -132,33 +132,35 @@ func main() {
...
-		}
-	default:
+	if strings.Count(index, "x") != len(indexPaths)-1 || strings.HasPrefix(indexPaths[len(indexPaths)-1], "x") {
 		return 0, 0, 0, fmt.Errorf("failed to parse tile index")
 	}
 
+	for _, indexPath := range indexPaths {
+		indexPath = strings.TrimPrefix(indexPath, "x")
+		n, err := strconv.ParseUint(indexPath, 10, 64)
+		if err != nil || n >= 1000 || len(indexPath) > 3 {
```
mhutchinson, line 159: Is the `len` part needed? The actual definition says that `It MUST be a non-negative integer encoded into zero-padded 3-digit path elements`, so this should actually be `len(indexPath) != 3`, but I think it's extra validation we don't strictly need? As in, if we can get a number out of it between [0,1000) then we're probably good to go.

mhutchinson (approved):
Thanks for taking the time to do the tests :-)

There are a couple of comments to be addressed, but I trust that the fixes won't need a re-review. Feel free to ask if you'd like one though.

cmd/example-mysql/main.go
roger2hk: I do agree that it's an extra validation. The spec said it is a MUST, so that's why I added it here. Updated to `len(indexPath) != 3`.

cmd/example-mysql/main.go
roger2hk: Done.

cmd/example-mysql/main.go
roger2hk: Added a comment. The maximum level is 63 as specified in the tlog-tiles specification.

cmd/example-mysql/main.go
roger2hk: I'll go ahead to merge the PR. Let me know if you think that this should be removed.

mhutchinson: There are arguments both ways. I'm happy with this approach and it can be changed later if needed.

PR #60, opened 2024-07-12
Consume previously sequenced entries

AlCutter:
This PR adds support for getting sequenced entries back from the sequencer portion of the GCP implementation, and is a precursor for adding the integration support.

Toward #23

### DIFF
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -72,8 +72,16 @@ type objStore interface {
 
 // coord describes a type which knows how to sequence entries.
 type sequencer interface {
+	// assignEntries should durably allocate contiguous index numbers to the provided entries,
+	// and return the lowest assigned index.
 	assignEntries(ctx context.Context, entries [][]byte) (uint64, error)
+	// consumeEntries should call the provided function with up to limit previously sequenced entries.
+	// If the call to consumeFunc returns no error, the entries should be considered to have been consumed.
+	consumeEntries(ctx context.Context, limit uint64, f consumeFunc) (bool, error)
 }
+
+// consumeFunc is the signature of a function which can consume entries from the sequencer.
+type consumeFunc func(ctx context.Context, from uint64, entries [][]byte) error
 
 // Config holds GCP project and resource configuration for a storage instance.
 type Config struct {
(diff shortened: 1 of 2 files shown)


storage/gcp/gcp.go
```diff
@@ -73,9 +72,17 @@ type objStore interface {
 
 // coord describes a type which knows how to sequence entries.
```
roger2hk, line 73: Should `coord` be `sequencer` in the doc comment?

storage/gcp/gcp.go
```diff
@@ -73,9 +72,17 @@ type objStore interface {
...
 type sequencer interface {
+	// assignEntries should durably allocate contiguous index numbers to the provided entries,
+	// and return the lowest assigned index.
 	assignEntries(ctx context.Context, entries [][]byte) (uint64, error)
+	// consumeEntries should call the provided function with up to limit previously sequenced entries.
+	// If the call to consumeFunc returns no error, the entries should be considered to have been consumed.
+	consumeEntries(ctx context.Context, limit uint64, f consumeFunc) (bool, error)
```
roger2hk, line 80: The doc comment doesn't tell the purpose of the returned `bool` although it was mentioned in the actual implementation doc comment.

storage/gcp/gcp.go
AlCutter: d'oh, fixed

storage/gcp/gcp.go
AlCutter: Good spot, added - thanks!

storage/gcp/gcp.go
```diff
@@ -302,6 +340,92 @@ func (s *spannerSequencer) assignEntries(ctx context.Context, entries [][]byte)
 	return uint64(next), nil
 }
 
+// consumeEntries calls f with previously sequenced entries.
+//
+// Once f returns without error, the entries it was called with are considered to have been consumed and are
+// removed from the Seq table.
+//
+// Returns true if some entries were consumed as a weak signal that there may be further entries waiting to be consumed.
+func (s *spannerSequencer) consumeEntries(ctx context.Context, limit uint64, f consumeFunc) (bool, error) {
```
roger2hk, line 349: What if the logic retrieves `limit + 1` sequenced entries to provide a strong signal?

roger2hk (approved):

storage/gcp/gcp.go
AlCutter: At that point it may as well just do the work (and so on), but I'm trying to keep the chunks of work bounded. I figure that the cost of an "false positive" signal here is not really that great.

This is all open to change in the near future though, if it turns out to perform poorly.

PR #61, opened 2024-07-12
Add integrate mix-in

AlCutter:
This PR adds a storage mix-in for performing the integration of new entries into the log.

Towards #6 & #23

### DIFF
--- a/storage/integrate.go
+++ b/storage/integrate.go
@@ -34,8 +34,13 @@ import (
 type getFullTileFunc func(ctx context.Context, tileID TileID, treeSize uint64) (*fullTile, error)
 
 // TreeBuilder constructs Merkle trees.
+//
+// This struct it indended to be used by storage implementations during the integration of entries into the log.
+// TreeBuilder caches data from tiles to speed things up, but has no mechanism for evicting from its internal cache,
+// so while it _may_ be possible to use the same instance across a number of integration runs (e.g. if the same job
+// is responsible for integrating entries for a number of contiguous trees), the lifetime should be bounded so as not
+// to leak memory.
 type TreeBuilder struct {
-	sync.Mutex
 	getTile getFullTileFunc
 	rf      *compact.RangeFactory
 }
@@ -97,9 +102,6 @@ func (t *TreeBuilder) newRange(ctx context.Context, treeSize uint64) (*compact.R
 }
 
 func (t *TreeBuilder) Integrate(ctx context.Context, fromSize uint64, entries [][]byte) (newSize uint64, rootHash []byte, tiles map[TileID]*api.HashTile, err error) {
-	t.Lock()
-	defer t.Unlock()
-
 	baseRange, err := t.newRange(ctx, fromSize)
 	if err != nil {
 		return 0, nil, nil, fmt.Errorf("failed to create range covering existing log: %w", err)
--- a/storage/integrate.go
+++ b/storage/integrate.go
@@ -26,7 +26,7 @@ import (
 	"github.com/transparency-dev/merkle/rfc6962"
 	"github.com/transparency-dev/trillian-tessera/api"
 	"github.com/transparency-dev/trillian-tessera/api/layout"
-	"golang.org/x/sync/errgroup"
+	"golang.org/x/exp/maps"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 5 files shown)


storage/integrate.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	klog.V(1).Infof("Loaded state with roothash %x", r)
+	// Create a new compact range which represents the update to the tree
+	newRange := t.rf.NewEmptyRange(fromSize)
+	if len(entries) == 0 {
+		klog.V(1).Infof("Nothing to do.")
+		// Nothing to do, nothing done.
+		return fromSize, r, nil, nil
+	}
```
mhutchinson, line 123: I don't know how common this will be, but I'd stick it at the top of the method to save all the work you don't need to do.

storage/integrate.go
```diff
@@ -0,0 +1,337 @@
...
+	e, ok := r.entries[k]
+	return e, ok
+}
+
+// Set associates the given tileID with a tile.
+func (r *tileReadCache) Set(tileID TileID, treeSize uint64, t *populatedTile) {
+	r.Lock()
+	defer r.Unlock()
+	k := layout.TilePath(uint64(tileID.Level), tileID.Index, treeSize)
+	if e, ok := r.entries[k]; ok && !e.Equals(t) {
```
mhutchinson, line 186: You should be able to drop this equality check if the read-once semantics for any given tile are fixed.

storage/integrate.go
```diff
@@ -0,0 +1,337 @@
...
+// Err returns an aggregated view of any errors seen by the visitor function.
+//
+// This can be used to check whether updates to the tile cache made by the visitor
+// were made correctly. Any errors returned here are most likely to be due to
+// the cache attempting to read an existing tile which is being updated.
+func (tc *tileWriteCache) Err() error {
+	return errors.Join(tc.err...)
+}
+
+// Visit should be called once for each newly set non-ephemeral node in the
```
mhutchinson, line 229: nit: This comment should start with `Visitor ...`. It appears to jump over what this function does and talks about how its return value should be used.

storage/integrate.go
```diff
@@ -0,0 +1,337 @@
...
+// tileWriteCache is a simple cache for storing the newly created tiles produced by
+// the integration of new leaves into the tree.
+//
+// Calls to Visit will cause the map of tiles to become filled with the set of
+// `dirty` tiles which need to be flushed back to storage to preserve the updated
+// tree state.
+//
+// Note that by itself, this cache does not update any persisted state.
+type tileWriteCache struct {
+	sync.Mutex
```
mhutchinson, line 202: I don't see where this is used in parallel. Can we remove this mutex to keep it simpler looking?

storage/integrate.go
```diff
@@ -0,0 +1,337 @@
...
+	// Calculate the new root hash - don't pass in the tileCache visitor here since
+	// this will construct any ephemeral nodes and we do not want to store those.
+	newRoot, err := baseRange.GetRootHash(nil)
+	if err != nil {
+		return 0, nil, nil, fmt.Errorf("failed to calculate new root hash: %w", err)
+	}
+
+	if err := tc.Err(); err != nil {
+		return 0, nil, nil, err
+	}
```
mhutchinson, line 149: Worth moving this up a few lines? Maybe even checking it before and after `baseRange.AppendRange(newRange, visitor)` ? Reasoning being that any errors in here are likely to cause failures in the other operations that depend on the writeTileCache succeeding. It's better to return the root error rather than a symptomatic error.

storage/integrate.go
```diff
@@ -0,0 +1,337 @@
...
+type TreeBuilder struct {
+	getTile getPopulatedTileFunc
+	rf      *compact.RangeFactory
+}
+
+// NewTreeBuilder creates a new instance of TreeBuilder.
+//
+// The getTile param must know how to fetch the specified tile from storage. It must return an instance of os.ErrNotExist
+// (either directly, or wrapped) if the requested tile was not found.
+func NewTreeBuilder(getTile func(ctx context.Context, tileID TileID, treeSize uint64) (*api.HashTile, error)) *TreeBuilder {
```
mhutchinson, line 52: As discussed in person, making this function take all the tiles to look up in batch allows for more flexibility in implementation. And as a side effect, makes it so that we don't end up fetching the same tile multiple times if there are several nodes required from the same tile.

storage/integrate.go
AlCutter: It should be pretty uncommon. Unfortunately, it'll be a little more work to move it all the way up since we return the root hash; we'd have to return some sort of sentinel error or similar to signal.

Not sure if it's worth it at the mo'?

storage/integrate.go
AlCutter: That's gone away now

storage/integrate.go
AlCutter: Yeah, it was there from an abundance of caution, but it should never be used in parallel since it's driven by the visitor.

Removed.

storage/integrate.go
AlCutter: Updated

storage/integrate.go
AlCutter: This is a great observation, thank you!
I've re-worked quite a bit of how this hangs together to support this, PTAL.

storage/integrate.go
AlCutter: done

storage/integrate.go
mhutchinson: I'd missed that it returned `r`. No worries.

storage/integrate.go
```diff
@@ -0,0 +1,346 @@
...
+	rf        *compact.RangeFactory
+}
+
+// NewTreeBuilder creates a new instance of TreeBuilder.
+//
+// The getTile param must know how to fetch the specified tile from storage. It must return an instance of os.ErrNotExist
+// (either directly, or wrapped) if any of the the requested tiles were not found.
+func NewTreeBuilder(getTiles func(ctx context.Context, tileIDs []TileID, treeSize uint64) ([]*api.HashTile, error)) *TreeBuilder {
+	readCache := newTileReadCache(getTiles)
+	r := &TreeBuilder{
```
mhutchinson, line 55: This is missing an assignment to `getTile` in the `TreeBuilder`. Which probably suggests missing tests as CI hasn't caught any problems, and this field is used in `Integrate`.

storage/integrate.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+// Get returns a previously set tile and true, or, if no such tile is in the cache, attempt to fetch it.
+func (r *tileReadCache) Get(ctx context.Context, tileID TileID, treeSize uint64) (*populatedTile, error) {
+	r.Lock()
+	defer r.Unlock()
+	k := layout.TilePath(uint64(tileID.Level), tileID.Index, treeSize)
+	e, ok := r.entries[k]
+	if !ok {
+		t, err := r.getTiles(ctx, []TileID{tileID}, treeSize)
```
mhutchinson, line 173: Could be worth logging something here because there shouldn't be any misses if prewarm was used, and anyone not doing prewarm is missing out on performance wins.

mhutchinson (approved):

storage/integrate.go
AlCutter: I'll pop a `V` log in there, I think there are legit times when this could happen during Integrate (e.g. when entirely new tiles are being created), but it'll be useful to peer in there and see what's happening.

storage/integrate.go
AlCutter: Whoops, that shouldn't be there - fixed.
Added a simple test for Integrate too.

PR #62, opened 2024-07-15
Add support for integration in GCP storage

AlCutter:
This PR pulls in the integrate mix-in from #61 and uses it in the GCP storage.

Towards #23

### DIFF
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -45,11 +45,14 @@ import (
 	"github.com/transparency-dev/trillian-tessera/api"
 	"github.com/transparency-dev/trillian-tessera/api/layout"
 	"github.com/transparency-dev/trillian-tessera/storage"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
 	"google.golang.org/grpc/codes"
 	"k8s.io/klog/v2"
 )
+
+const entryBundleSize = 256
 
 // Storage is a GCP based storage implementation for Tessera.
 type Storage struct {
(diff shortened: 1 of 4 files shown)


storage/gcp/gcp.go
```diff
@@ -158,22 +161,39 @@ func (s *Storage) setTile(ctx context.Context, level, index, logSize uint64, til
...
+			if err := t.UnmarshalText(data); err != nil {
+				return fmt.Errorf("unmarshal(%q): %v", objName, err)
+			}
+			r[i] = t
+			return nil
+		})
+	}
+	if err := errG.Wait(); err != nil {
 		return nil, err
 	}
```
mhutchinson, line 194: Does this cause a problem if the first error is `NotExist` but another one is a real failure?

storage/gcp/gcp.go
```diff
@@ -217,7 +237,102 @@ func (s *Storage) setEntryBundle(ctx context.Context, bundleIndex uint64, logSiz
...
+	seqErr := errgroup.Group{}
+
+	// goSetEntryBundle is a function which uses seqErr to spin off a go-routine to write out an entry bundle.
+	// It's used in the for loop below.
+	goSetEntryBundle := func(ctx context.Context, bundleIndex uint64, fromSeq uint64, bundle api.EntryBundle) {
+		seqErr.Go(func() error {
+			if err := s.setEntryBundle(ctx, bundleIndex, fromSeq, &bundle); err != nil {
+				if !errors.Is(os.ErrExist, err) {
+					return err
+				}
```
mhutchinson, line 306: It could be worth logging something here when a duplicate write has happened. What I'd actually prefer would be prometheus counters for successful write, error write, duplicate write, etc. But that can come later?

storage/gcp/gcp.go
```diff
@@ -217,7 +237,102 @@ func (s *Storage) setEntryBundle(ctx context.Context, bundleIndex uint64, logSiz
...
+		}
+		for k, v := range tiles {
+			func(ctx context.Context, k storage.TileID, v *api.HashTile) {
+				errG.Go(func() error {
+					return s.setTile(ctx, uint64(k.Level), k.Index, newSize, v)
+				})
+			}(ctx, k, v)
+		}
+		errG.Go(func() error {
+			//TODO: write out checkpoint
```
mhutchinson, line 270: This is quite a load bearing TODO :-)

mhutchinson (approved):

storage/gcp/gcp.go
AlCutter: Coming soon, to a PR near you! :)

storage/gcp/gcp.go
AlCutter: FWIW the GCP impl already logs a (verbose level) message when it encounters an idempotent dupe write, but agree that metrics are the right thing.

mhutchinson (approved):

Issue #22, opened 2024-06-17
POSIX FS storage implementation


patflynn:

mhutchinson:
Functionality added in #67. Will leave this open as @AlCutter has some more changes that build on that PR and can be tracked against this issue.

PR #67, opened 2024-07-17
Posix log implementation of tlog-tiles

mhutchinson:
This is largely a copy and paste job from both the experimental Betty repo and from serverless. The primary motivation to have this posix implementation right now is to generate test data, such as that used in #49.

There has been little in the way of optimization. The primary concern was minimal code changes in order to get a working posix log to generate test data.

Towards #22

### DIFF
--- a/cmd/example-posix/main.go
+++ b/cmd/example-posix/main.go
@@ -1,0 +1,134 @@
+// Copyright 2021 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


cmd/example-posix/main.go
```diff
@@ -47,19 +53,50 @@ func main() {
...
+		klog.Exitf("Failed to instantiate signer: %q", err)
+	}
+	if *initialise {
+		if err := os.MkdirAll(*storageDir, dirPerm); err != nil {
+			klog.Exitf("Failed to create log directory: %q", err)
+		}
+		cp := fmtlog.Checkpoint{
+			Hash: rfc6962.DefaultHasher.EmptyRoot(),
+		}
+		if err := signAndWrite(&cp, cpNote, s, *storageDir); err != nil {
```
AlCutter, line 94: Probably worth trying to remove `signAndWrite` and use `writeCP` (or similar) if possible - currently initialising the log will use `*origin` for the `CP.Origin` but updates to the log will use `s.Name()`.

AlCutter (approved):
Just a quick review since most of this code was already reviewed in `serverless`.

Just the one thing inline stuck out.

PR #68, opened 2024-07-17
Add Checkpoint option & batch options

AlCutter:
This PR adds two common storage options:
- one for setting a signer to be used when signing checkpoints
- one for configuring the batching queue.


storage/gcp/gcp.go
```diff
@@ -273,8 +277,16 @@ func (s *Storage) integrate(ctx context.Context, fromSeq uint64, entries []tesse
 			}(ctx, k, v)
 		}
 		errG.Go(func() error {
-			//TODO: write out checkpoint
 			klog.Infof("New CP: %d, %x", newSize, newRoot)
```
mhutchinson, line 280: This should probably be made `V(2)` or so.

mhutchinson (approved):

PR #70, opened 2024-07-17
GCP: temporary hack to proxy read requests to GCS

AlCutter:
This PR allows an easy way for the hammer to get access to the log static resources by proxying read requests into GCS reads.

### DIFF
--- a/cmd/example-gcp/main.go
+++ b/cmd/example-gcp/main.go
@@ -80,11 +80,10 @@ func main() {
 		if err != nil {
 			klog.V(1).Infof("Get: %v", err)
 			w.WriteHeader(http.StatusBadRequest)
-			w.Write([]byte(fmt.Sprintf("Get: %v", err)))
+			_, _ = w.Write([]byte(fmt.Sprintf("Get: %v", err)))
 			return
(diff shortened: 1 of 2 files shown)


cmd/example-gcp/main.go
```diff
@@ -72,6 +73,21 @@ func main() {
 		_, _ = w.Write([]byte(fmt.Sprintf("%d", idx)))
 	})
 
+	// TODO: remove this proxy
+	serveGCS := func(w http.ResponseWriter, r *http.Request) {
+		resource := strings.TrimLeft(r.URL.Path, "/")
+		b, err := storage.Get(r.Context(), resource)
+		if err != nil {
+			klog.V(1).Infof("Get: %v", err)
+			w.WriteHeader(http.StatusBadRequest)
```
mhutchinson, line 82: Classic denial of any responsibility here :-)

mhutchinson (approved):

cmd/example-gcp/main.go
AlCutter: It's probably your fault, to be fair. :)

PR #75, opened 2024-07-18
Add TODOs to handle partial tile scenarios in MySQL storage implementation

roger2hk:
#21

### DIFF
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -99,17 +99,44 @@ func (s *Storage) writeCheckpoint(ctx context.Context, rawCheckpoint []byte) err
 }
 
 // ReadTile returns a full tile or a partial tile at the given level and index.
-func (s *Storage) ReadTile(ctx context.Context, level, index uint64) ([]byte, error) {
+//
+// TODO: Handle the following scenarios:
+// 1. Full tile request with full tile output: Return full tile.
+// 2. Full tile request with partial tile output: Return error.
+// 3. Partial tile request with full/larger partial tile output: Return trimmed partial tile with correct tile width.
+// 4. Partial tile request with partial tile (same width) output: Return partial tile.
+// 5. Partial tile request with smaller partial tile output: Return error.
+func (s *Storage) ReadTile(ctx context.Context, level, index, width uint64) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


storage/mysql/mysql.go
```diff
@@ -99,17 +99,44 @@ func (s *Storage) writeCheckpoint(ctx context.Context, rawCheckpoint []byte) err
 }
 
 // ReadTile returns a full tile or a partial tile at the given level and index.
-func (s *Storage) ReadTile(ctx context.Context, level, index uint64) ([]byte, error) {
+//
+// TODO: Handle the following scenarios:
+// 1. Full tile request with full tile output: Return full tile.
+// 2. Full tile request with partial tile output: Return error.
+// 3. Partial tile request with full/larger partial tile output: Return trimmed partial tile with correct tile width.
```
AlCutter, line 106: Worth just thinking about returning the full tile/bundle in the cases where the request is for a smaller partial size...

Clients of the log should only be requesting tiles when they have a checkpoint "in hand" which implies their existence. They should never rely on anything beyond the bounds of the tree which the checkpoint commits to since anything there is by definition not committed to. So, there shouldn't really be any issue if returning a larger/full tile when a partial one was requested.

storage/mysql/mysql.go
```diff
@@ -99,17 +99,44 @@ func (s *Storage) writeCheckpoint(ctx context.Context, rawCheckpoint []byte) err
...
+// 4. Partial tile request with partial tile (same width) output: Return partial tile.
+// 5. Partial tile request with smaller partial tile output: Return error.
+func (s *Storage) ReadTile(ctx context.Context, level, index, width uint64) ([]byte, error) {
```
AlCutter, line 109: One thing to note is that most of the other code (e.g. the get tiles function passed into `TreeBuilder`) uses `treeSize` rather than width here, but feel free to adjust/adapt to that when you're ready to integrate (haha!) with that.

AlCutter (approved):

storage/mysql/mysql.go
roger2hk: Good to know. This method will probably be updated when the sequencing and integrate flows are added.

storage/mysql/mysql.go
roger2hk: Quite right. Let's have a team discussion this week.

PR #76, opened 2024-07-19
Reintroduce sync for flushing

AlCutter:
This PR reintroduces a single-threaded nature for queue flushing.

Given the strictly serialised nature of assigning sequence numbers without gaps it's unlikely that storage implementations would be able to increase performance by taking multiple flushes in parallel.

Toward #6

### DIFF
--- a/storage/queue.go
+++ b/storage/queue.go
@@ -82,7 +82,7 @@ func NewQueue(ctx context.Context, maxAge time.Duration, maxSize uint, f FlushFu
 		buffer.WithFlusher(buffer.FlusherFunc(toWork)),
 	)
 
-	// This will allow
+	// Spin off a worker thread to write the queue flushes to storage.
 	go func(ctx context.Context) {
 		for {
 			select {
--- a/storage/queue.go
+++ b/storage/queue.go
@@ -55,16 +55,44 @@ type FlushFunc func(ctx context.Context, entries []tessera.Entry) (index uint64,
 // The provided FlushFunc will be called with a slice containing the contents of the queue, in
 // the same order as they were added, when either the oldest entry in the queue has been there
 // for maxAge, or the size of the queue reaches maxSize.
-func NewQueue(maxAge time.Duration, maxSize uint, f FlushFunc) *Queue {
+func NewQueue(ctx context.Context, maxAge time.Duration, maxSize uint, f FlushFunc) *Queue {
 	q := &Queue{
 		flush:    f,
 		inFlight: make(map[string]*entry, maxSize),
 	}
+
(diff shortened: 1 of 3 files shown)


storage/queue.go
```diff
@@ -55,16 +55,44 @@ type FlushFunc func(ctx context.Context, entries []tessera.Entry) (index uint64,
...
-func NewQueue(maxAge time.Duration, maxSize uint, f FlushFunc) *Queue {
+func NewQueue(ctx context.Context, maxAge time.Duration, maxSize uint, f FlushFunc) *Queue {
 	q := &Queue{
 		flush:    f,
 		inFlight: make(map[string]*entry, maxSize),
 	}
+
+	// The underlying queue implementation blocks additions during a flush.
+	// This blocks the filling of the next batch unnecessarily, so we'll
+	// decouple the queue flush and storage write by handling the later in
```
roger2hk, line 66:
```suggestion
	// decouple the queue flush and storage write by handling the latter in
```

storage/queue.go
```diff
@@ -101,27 +128,23 @@ func (q *Queue) Add(ctx context.Context, e tessera.Entry) IndexFunc {
 //
 // To prevent blocking the queue longer than necessary, the notifications happen in a
 // separate goroutine.
```
roger2hk, line 130: Is this comment still valid as the separate goroutine is now moved to the upper level?

storage/queue.go
```diff
@@ -55,16 +55,44 @@ type FlushFunc func(ctx context.Context, entries []tessera.Entry) (index uint64,
...
 		inFlight: make(map[string]*entry, maxSize),
 	}
+
+	// The underlying queue implementation blocks additions during a flush.
+	// This blocks the filling of the next batch unnecessarily, so we'll
+	// decouple the queue flush and storage write by handling the later in
+	// a worker goroutine.
+	// This same worker thread will also handle the callbacks to f.
+	work := make(chan []*entry, 1)
+	toWork := func(items []interface{}) {
```
roger2hk, line 70: Why is the `interface{}` instead of `*entry` used here?

roger2hk (approved):

storage/queue.go
AlCutter: Great spot, thank you!
fixed.

storage/queue.go
AlCutter: That's an artifact of the underlying queue library, it's pre-generics (I guess) so is built as a queue of `interface{}`. This func just converts those back to the actual type we know it contains when the queue flushes.

PR #82, opened 2024-07-23
Improve `writeCheckpoint` in MySQL storage implementation

roger2hk:
#21

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,7 +56,7 @@ require (
 	go.opentelemetry.io/otel v1.24.0 // indirect
 	go.opentelemetry.io/otel/metric v1.24.0 // indirect
 	go.opentelemetry.io/otel/trace v1.24.0 // indirect
-	golang.org/x/crypto v0.25.0 // indirect
+	golang.org/x/crypto v0.25.0
 	golang.org/x/net v0.27.0 // indirect
 	golang.org/x/oauth2 v0.21.0 // indirect
 	golang.org/x/sync v0.7.0
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -20,6 +20,8 @@ import (
 	"database/sql"
 
 	_ "github.com/go-sql-driver/mysql"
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -56,7 +56,7 @@ require (
 	go.opentelemetry.io/otel v1.24.0 // indirect
 	go.opentelemetry.io/otel/metric v1.24.0 // indirect
 	go.opentelemetry.io/otel/trace v1.24.0 // indirect
-	golang.org/x/crypto v0.25.0 // indirect
+	golang.org/x/crypto v0.25.0
```
mhutchinson, line 59: I don't see how this change could have caused this. Did you run `go mod tidy`?

storage/mysql/mysql.go
```diff
@@ -55,10 +60,25 @@ func New(ctx context.Context, db *sql.DB) (*Storage, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		// Defer a rollback in case anything fails.
+		defer func() {
+			if err := tx.Rollback(); err != nil && err != sql.ErrTxDone {
+				klog.Errorf("Failed to rollback in write initial checkpoint: %v", err)
+			}
+		}()
+		if err := s.writeCheckpoint(ctx, tx, 0, []byte("")); err != nil {
```
mhutchinson, line 74: Is `[]byte("")` the right value? That's the empty byte string, where it should be a _hash_ of the empty bytes. See https://github.com/transparency-dev/trillian-tessera/blob/main/cmd/example-posix/main.go#L102C24-L102C57 where it uses `rfc6962.DefaultHasher.EmptyRoot()`

go.mod
roger2hk: Yes, I did.

storage/mysql/mysql.go
roger2hk: Updated to `rfc6962.DefaultHasher.EmptyRoot()`, thanks!

mhutchinson (approved):

PR #83, opened 2024-07-24
Add `WithCheckpointVerifier` to `tessera.StorageOptions` for parsing checkpoints

roger2hk:
#21

### DIFF
--- a/cmd/example-gcp/main.go
+++ b/cmd/example-gcp/main.go
@@ -50,7 +50,7 @@ func main() {
 		Bucket:    *bucket,
 		Spanner:   *spanner,
 	}
-	storage, err := gcp.New(ctx, gcpCfg, tessera.WithCheckpointSigner(signerFromFlags()))
+	storage, err := gcp.New(ctx, gcpCfg, tessera.WithCheckpointSignerVerifier(signerFromFlags(), nil))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


AlCutter (commented):
Just a thought, would it make sense to roll this into the existing `WithCheckpointSigner()` option?

The existing storage implementations don't need a verifier since they store state internally and just "render" checkpoint from that so never need to verify it, but if there will be other patterns where this isn't case (POSIX too, presumably?) then it'd probably make sense to pop signer & verifier in the same option; it'll avoid bugs where someone passes a signer but not a verifier, and then they only find out when their binary crashes when it tries to integrate.

roger2hk:
> Just a thought, would it make sense to roll this into the existing `WithCheckpointSigner()` option?
> 
> The existing storage implementations don't need a verifier since they store state internally and just "render" checkpoint from that so never need to verify it, but if there will be other patterns where this isn't case (POSIX too, presumably?) then it'd probably make sense to pop signer & verifier in the same option; it'll avoid bugs where someone passes a signer but not a verifier, and then they only find out when their binary crashes when it tries to integrate.

Good point. I have refactored the `WithCheckpointSigner` to `WithCheckpointSignerVerifier`.

mhutchinson (approved):

PR #86, opened 2024-07-24
Complete instructions to deploy Tessera GCP

phbnf:
This PR also changes the base name from "example-gcp" to "trillian-tessera" since I think it makes more sense as a default value. I'd rather fix this sooner rather than later, but happy to revert this change if you think it's too disruptive.

### DIFF
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -26,7 +26,7 @@ export GOOGLE_PROJECT={VALUE}
 ```
 
 Eventually, customize the region (defaults to "us-east1"), and bucket name prefix
-(defaults to "trillian-tessera"):
+(defaults to "tessera"):
 ```bash
 export GOOGLE_REGION={VALUE}
 export TESSERA_BASE_NAME={VALUE}
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -14,25 +14,5 @@ Deploying these examples requires installation of:
 
 ## Deploying
 
-First authenticate via `gcloud` as a principle with sufficient ACLs for
-the project:
-```bash
-gcloud auth application-default login
-```
-
-Then, specify your Google Cloud project ID:
-```bash
-export GOOGLE_PROJECT={VALUE}
-```
-
-Eventually, customize the region (defaults to "us-east1"), and bucket name prefix
-(defaults to "tessera"):
-```bash
-export GOOGLE_REGION={VALUE}
-export TESSERA_BASE_NAME={VALUE}
-```
-
-Terraforming the project can be done by:
- 1. `cd` to the relevant `live` directory for the environment to deploy/change
- 2. Run `terragrunt apply`
+See individual `live` subdirectories.
 
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -20,6 +20,18 @@ the project:
 gcloud auth application-default login
 ```
 
+Then, specify your Google Cloud project ID:
+```bash
+export GOOGLE_PROJECT={VALUE}
+```
+
+Eventually, customize the region (defaults to "us-east1"), and bucket name prefix
+(defaults to "trillian-tessera"):
+```bash
+export GOOGLE_REGION={VALUE}
+export TESSERA_BASE_NAME={VALUE}
+```
+
 Terraforming the project can be done by:
  1. `cd` to the relevant `live` directory for the environment to deploy/change
  2. Run `terragrunt apply`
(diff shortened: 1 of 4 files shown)


deployment/README.md
```diff
@@ -20,6 +20,18 @@ the project:
 gcloud auth application-default login
 ```
 
+Then, specify your Google Cloud project ID:
```
AlCutter, line 23: This GCP specific stuff should probably go into a README in the `live/example-gcp` directory (my fault - I accidentally put the `gcloud` line in above)

deployment/README.md
```diff
@@ -20,6 +20,18 @@ the project:
...
+```bash
+export GOOGLE_PROJECT={VALUE}
```
AlCutter, line 25: Could use this:
```bash
export GOOGLE_PROJECT=$(gcloud config get-value project 2> /dev/null)
```

deployment/README.md
```diff
@@ -20,6 +20,18 @@ the project:
...
+```
+
+Eventually, customize the region (defaults to "us-east1"), and bucket name prefix
+(defaults to "tessera"):
+```bash
+export GOOGLE_REGION={VALUE}
```
AlCutter, line 31: It feels a bit weird to me to have the behaviour of terragrunt be affected by environment vars (I think it's a niggle from "hermetic config as code & checked in" being violated), but:
a) I'm far from a TF/TG expert, and
b) I suppose this is meant to be an easily runnable example and having folks edit config files in a repo is a bit of an anti pattern.

deployment/live/example-gcp/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+First authenticate via `gcloud` as a principle with sufficient ACLs for
+the project:
+```bash
+gcloud auth application-default login
+```
+
+Then, specify your Google Cloud project ID:
+```bash
+export GOOGLE_PROJECT={VALUE}
```
mhutchinson, line 11: What's the advantage of doing this with environment variables? It moves away from the the deployment files being fully declarative, which means that two people can check out the same repo and then create different deployments because of inconsistencies in the env naming. It's also more work each time you set up a shell to get back to the same state. The obvious workaround is to create a shell script that sets these values, but then you've just put them in a file, which this change moves away from.

deployment/live/example-gcp/README.md
```diff
@@ -0,0 +1,24 @@
...
+```
+
+Eventually, customize the region (defaults to "us-east1"), and bucket name prefix
```
mhutchinson, line 14: It looks like the default is `us-central1` ?

deployment/live/example-gcp/README.md
mhutchinson: Huh, I just saw Al's comments which only appear at the overview screen and not on the "files changed" view. Looks like these comments are somewhat redundant!

deployment/README.md
phbnf: Maybe that works if you run this command after you've ran `gcloud auth` from a VM inside a project, but that might not always be the case. Otherwise, how can gcloud know which project it should select? When you authenticate via a web link, it doesn't give you the option to select the project.

AlCutter: Sadness :(

deployment/live/example-gcp/terragrunt.hcl
```diff
@@ -3,9 +3,9 @@ terraform {
 }
 
 locals {
-  project_id = "trillian-tessera"
-  location   = "us-central1"
-  base_name   = "example-gcp"
+  project_id = get_env("GOOGLE_PROJECT", "trillian-example")
```
mhutchinson, line 6: Is this supposed to be `trillian-tessera` to match the old version? I think this name is particularly confusing given we have a repo with (almost) this name which is for the old trillian!

phbnf: Duh. Done, thanks.

deployment/live/example-gcp/README.md
```diff
@@ -0,0 +1,24 @@
...
+```bash
+gcloud auth application-default login
+```
+
+Set your GCP project ID with:
+```bash
+export GOOGLE_PROJECT={VALUE}
+```
+
+Eventually, customize the GCP region (defaults to "us-central1"),
```
mhutchinson, line 14:
```suggestion
Optionally, customize the GCP region (defaults to "us-central1"),
```

mhutchinson (approved):

PR #87, opened 2024-07-24
Add sequence and integrate foundation to MySQL storage implementation

roger2hk:
#21

### DIFF
--- a/cmd/example-mysql/main.go
+++ b/cmd/example-mysql/main.go
@@ -17,7 +17,6 @@ package main
 
 import (
 	"context"
-	"crypto/sha256"
 	"database/sql"
 	"flag"
 	"fmt"
@@ -166,8 +165,7 @@ func main() {
 			}
 		}()
 
-		id := sha256.Sum256(b)
-		idx, err := storage.Add(r.Context(), tessera.NewEntry(b, tessera.WithIdentity(id[:])))
+		idx, err := storage.Add(r.Context(), tessera.NewEntry(b))
 		if err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
 			_, _ = w.Write([]byte(err.Error()))
--- a/cmd/example-mysql/main.go
+++ b/cmd/example-mysql/main.go
@@ -160,12 +160,23 @@ func main() {
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
-		defer r.Body.Close()
+		defer func() {
+			if err := r.Body.Close(); err != nil {
+				klog.Warningf("/add: %v", err)
+			}
+		}()
 
 		id := sha256.Sum256(b)
-		_ = tessera.NewEntry(b, tessera.WithIdentity(id[:]))
-
-		// TODO: Add entry to log and return assigned index.
+		idx, err := storage.Add(r.Context(), tessera.NewEntry(b, tessera.WithIdentity(id[:])))
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			_, _ = w.Write([]byte(err.Error()))
+			return
+		}
+		if _, err = w.Write([]byte(fmt.Sprintf("%d", idx))); err != nil {
+			klog.Errorf("/add: %v", err)
+			return
+		}
 	})
 
 	if err := http.ListenAndServe(*listen, http.DefaultServeMux); err != nil {
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -31,12 +31,13 @@ import (
 )
 
 const (
-	selectCheckpointByIDSQL         = "SELECT `note` FROM `Checkpoint` WHERE `id` = ?"
-	replaceCheckpointSQL            = "REPLACE INTO `Checkpoint` (`id`, `note`) VALUES (?, ?)"
-	selectSubtreeByLevelAndIndexSQL = "SELECT `nodes` FROM `Subtree` WHERE `level` = ? AND `index` = ?"
-	replaceSubtreeSQL               = "REPLACE INTO `Subtree` (`level`, `index`, `nodes`) VALUES (?, ?, ?)"
-	selectTiledLeavesSQL            = "SELECT `data` FROM `TiledLeaves` WHERE `tile_index` = ?"
-	replaceTiledLeavesSQL           = "REPLACE INTO `TiledLeaves` (`tile_index`, `data`) VALUES (?, ?)"
+	selectCheckpointByIDSQL          = "SELECT `note` FROM `Checkpoint` WHERE `id` = ?"
+	selectCheckpointByIDForUpdateSQL = selectCheckpointByIDSQL + " FOR UPDATE"
+	replaceCheckpointSQL             = "REPLACE INTO `Checkpoint` (`id`, `note`) VALUES (?, ?)"
+	selectSubtreeByLevelAndIndexSQL  = "SELECT `nodes` FROM `Subtree` WHERE `level` = ? AND `index` = ?"
+	replaceSubtreeSQL                = "REPLACE INTO `Subtree` (`level`, `index`, `nodes`) VALUES (?, ?, ?)"
+	selectTiledLeavesSQL             = "SELECT `data` FROM `TiledLeaves` WHERE `tile_index` = ?"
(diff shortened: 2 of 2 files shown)


cmd/example-mysql/main.go
```diff
@@ -160,12 +160,23 @@ func main() {
...
+			if err := r.Body.Close(); err != nil {
+				klog.Warningf("/add: %v", err)
+			}
+		}()
 
 		id := sha256.Sum256(b)
-		_ = tessera.NewEntry(b, tessera.WithIdentity(id[:]))
-
-		// TODO: Add entry to log and return assigned index.
+		idx, err := storage.Add(r.Context(), tessera.NewEntry(b, tessera.WithIdentity(id[:])))
```
mhutchinson, line 170: I think you can drop the `WithIdentity` spec, that should be the default?

storage/mysql/mysql.go
```diff
@@ -159,3 +182,126 @@ func (s *Storage) ReadEntryBundle(ctx context.Context, index uint64) ([]byte, er
 	var entryBundle []byte
 	return entryBundle, row.Scan(&entryBundle)
 }
+
+// Add is the entrypoint for adding entries to a sequencing log.
+func (s *Storage) Add(ctx context.Context, entry *tessera.Entry) (uint64, error) {
+	// TODO: Return index if the value is already stored.
```
mhutchinson, line 188: nit: TODOs should be either `TODO(username)` or `TODO(issueid)`. We should set up a presubmit validator for this...

storage/mysql/mysql.go
```diff
@@ -18,41 +18,54 @@ package mysql
...
+		db:              db,
+		newCheckpoint:   opt.NewCP,
+		parseCheckpoint: opt.ParseCP,
 	}
 	if err := s.db.Ping(); err != nil {
 		klog.Errorf("Failed to ping database: %v", err)
 		return nil, err
 	}
 
+	s.queue = storage.NewQueue(ctx, time.Second, entryBundleSize, s.sequenceBatch)
```
mhutchinson, line 67: Maybe pull out the `time.Second` to a constant with `entryBundleSize` so that these tuning parameters are co-located.

storage/mysql/mysql.go
```diff
@@ -159,3 +182,126 @@ func (s *Storage) ReadEntryBundle(ctx context.Context, index uint64) ([]byte, er
...
+}
+
+// sequenceBatch writes the entries from the provided batch into the entry bundle files of the log.
+//
+// This func starts filling entries bundles at the next available slot in the log, ensuring that the
+// sequenced entries are contiguous from the zeroth entry (i.e left-hand dense).
+// We try to minimise the number of partially complete entry bundles by writing entries in chunks rather
+// than one-by-one.
+//
+// TODO: Separate sequencing and integration for better performance.
```
mhutchinson, line 200: This one should be TODO(issueid) and raise an issue to do this, as we agreed we'll defer this for at least a week.

storage/mysql/mysql.go
```diff
@@ -159,3 +182,126 @@ func (s *Storage) ReadEntryBundle(ctx context.Context, index uint64) ([]byte, er
...
+	}
+	// Defer a rollback in case anything fails.
+	defer func() {
+		if err := tx.Rollback(); err != nil && err != sql.ErrTxDone {
+			klog.Errorf("Failed to rollback in sequenceBatch: %v", err)
+		}
+	}()
+
+	// Get tree size from checkpoint.
+	row := tx.QueryRowContext(ctx, selectCheckpointByIDSQL, checkpointID)
```
mhutchinson, line 220: I don't think this has an exclusive lock at this point? This may be repeating a race condition from the early experiment where a sufficiently high load of write traffic can cause multiple batches to be sequenced in parallel. Each of these performs this read, starts sequencing at the same checkpoint size, and then stomp on the other writes, with only the last winning.

The way this was fixed in the experiment was with a `SELECT ... FOR UPDATE`: https://github.com/AlCutter/betty/blob/mysql/storage/tsql/sql.go#L195

storage/mysql/mysql.go
```diff
@@ -159,3 +182,126 @@ func (s *Storage) ReadEntryBundle(ctx context.Context, index uint64) ([]byte, er
...
+		return err
+	}
+	var rawCheckpoint []byte
+	if err := row.Scan(&rawCheckpoint); err != nil {
+		return fmt.Errorf("failed to read checkpoint: %w", err)
+	}
+	checkpoint, err := s.parseCheckpoint(rawCheckpoint)
+	if err != nil {
+		return fmt.Errorf("failed to verify checkpoint: %w", err)
+	}
```
mhutchinson, line 231: It _might_ be worth considering an optimization to avoid parsing and verifying the sigs each time. Two options spring to mind:
 - storing the size in a column
 - a hackier version of parsing the checkpoint where we just pull out the second line and don't verify the sigs

We should leave this for now and approach this with benchmarking later.

storage/mysql/mysql.go
```diff
@@ -159,3 +182,126 @@ func (s *Storage) ReadEntryBundle(ctx context.Context, index uint64) ([]byte, er
...
+// integrate incorporates the provided entries into the log starting at fromSeq.
+func (s *Storage) integrate(ctx context.Context, tx *sql.Tx, fromSeq uint64, entries []*tessera.Entry) error {
+	tb := storage.NewTreeBuilder(func(ctx context.Context, tileIDs []storage.TileID, treeSize uint64) ([]*api.HashTile, error) {
+		r := make([]*api.HashTile, len(tileIDs))
+		errG := errgroup.Group{}
+		for i, id := range tileIDs {
+			i := i
+			id := id
+			errG.Go(func() error {
+				// TODO: Refactor the query into a function.
```
mhutchinson, line 251: A better TODO would be to not make parallel SQL queries, and to make a single query that fetches all the required tiles as a batch?

The original design that @AlCutter had for this tree builder called this function N times, once for each TileID. I convinced him to change it to pass in a whole batch so that implementations such as the SQL one could fetch the values as a batch rather than in parallel requests. Would be a shame to not use it here!

storage/mysql/mysql.go
```diff
@@ -159,3 +182,126 @@ func (s *Storage) ReadEntryBundle(ctx context.Context, index uint64) ([]byte, er
...
+				return nil
+			})
+		}
+		if err := errG.Wait(); err != nil {
+			return nil, err
+		}
+		return r, nil
+	})
+
+	// TODO: Add sequenced entries to entry bundles.
```
mhutchinson, line 275: This is quite the load bearing TODO 😆

storage/mysql/mysql.go
AlCutter: There's a storage opt for these BTW - `WithBatching()`
https://github.com/transparency-dev/trillian-tessera/blob/main/log.go#L82

cmd/example-mysql/main.go
roger2hk: Done.

storage/mysql/mysql.go
roger2hk: Linked it back to #21, and I've added a task list in #21.

storage/mysql/mysql.go
roger2hk: Added `defaultEntryBundleSize` and `defaultQueueMaxAge` as the default storage options. `WithBatching()` storage option is also respected.

storage/mysql/mysql.go
roger2hk: Linked it back to #21.

storage/mysql/mysql.go
roger2hk: Right, it is quite a load bearing TODO as I don't want to have a huge pull request. Linked this TODO with #21.

mhutchinson (approved):

storage/mysql/mysql.go
roger2hk: Good idea. Added a TODO and linked to #21 .

storage/mysql/mysql.go
roger2hk: 100% agree with this. It's a shame not to fully utilise the RDBMS here! Added a TODO and linked to #21.

PR #91, opened 2024-07-25
Support WithBatching in GCP

AlCutter:
This PR enables support of `WithBatching` in the GCP storage.

Toward #23

### DIFF
--- a/cmd/example-gcp/main.go
+++ b/cmd/example-gcp/main.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	tessera "github.com/transparency-dev/trillian-tessera"
 	"github.com/transparency-dev/trillian-tessera/storage/gcp"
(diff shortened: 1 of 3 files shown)


cmd/example-gcp/main.go
```diff
@@ -50,7 +51,10 @@ func main() {
 		Bucket:    *bucket,
 		Spanner:   *spanner,
 	}
-	storage, err := gcp.New(ctx, gcpCfg, tessera.WithCheckpointSignerVerifier(signerFromFlags(), nil))
+	storage, err := gcp.New(ctx, gcpCfg,
+		tessera.WithCheckpointSignerVerifier(signerFromFlags(), nil),
+		tessera.WithBatching(1024, time.Second),
```
roger2hk, line 56: If `WithBatching` is not defined, the queue will fail. Are we supposed to have default values in the storage implementation? Is this a mandatory option for personality to implement?

phbnf: You could set the default buffer size to 1 here, https://github.com/transparency-dev/trillian-tessera/blob/main/log.go#L43, and it should then flush for every new entry, i.e not batch.

AlCutter: Yeah, exactly - that's what that little `!= nil` hook is intended for, I just didn't get around to adding _default_ defaults yet :) 

This PR seems as good a place as any, I'll add that now.

Thanks both!

roger2hk: Ah, I've added some default values in MySQL storage implementation. I can remove them after this PR is merged.

https://github.com/transparency-dev/trillian-tessera/blob/8c5a7eef32de27b60645293dce813ca104ec70cf/storage/mysql/mysql.go#L58-L61

phbnf (approved):

PR #92, opened 2024-07-26
Add support for pushback on GCP

AlCutter:
This PR adds a simple pushback implementation for GCP.

This enables a configurable threshold for applying back-pressure when there are too many
as-yet unintegrated entries.

Toward #6 #23


roger2hk (approved):

cmd/example-gcp/main.go
```diff
@@ -70,6 +71,11 @@ func main() {
 		id := sha256.Sum256(b)
 		idx, err := storage.Add(r.Context(), tessera.NewEntry(b, tessera.WithIdentity(id[:])))
 		if err != nil {
+			if errors.Is(err, tessera.ErrPushback) {
+				w.Header().Add("Retry-After", "1")
+				w.WriteHeader(http.StatusTooManyRequests)
```
roger2hk, line 76: 429 means the client has sent too many requests but it may not be true in this scenario.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

AlCutter: Do you think a 503 would be better idea?

roger2hk: However, this is the best HTTP status code we can get.

roger2hk: 503 is probably better in this scenario (caused by server limit) while 429 can be used in the client quota limit scenario.

roger2hk (approved):

PR #95, opened 2024-07-29
Foundation for a load & conformance test

mhutchinson:
The hammer makes reads (and optionally writes) to a log in order to test
its performance envelope and to allow benchmarking. It also checks
inclusion and consistency proofs, and thus performs some level of
conformance checking too, though this is a secondary objective.


mhutchinson:
This draft PR is to allow people with log implementations to trial integrate. I haven't even looked through this code since it was pulled together, so I make no claim that this is approaching ready for review / merge. I'll take a look in the coming days when I'm finished with oncall interrupts.

client/client.go
```diff
@@ -286,27 +286,25 @@ func newTileFetcher(f Fetcher, logSize uint64) GetTileFunc {
...
-func GetLeaf(ctx context.Context, f Fetcher, i, logSize uint64) ([]byte, error) {
+// GetEntryBundle fetches the entry bundle at the given _tile index_.
+func GetEntryBundle(ctx context.Context, f Fetcher, i, logSize uint64) (api.EntryBundle, error) {
+	bundle := api.EntryBundle{}
 	p := layout.EntriesPath(i, logSize)
 	sRaw, err := f(ctx, p)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			return nil, fmt.Errorf("leaf index %d not found: %w", i, err)
+			return bundle, fmt.Errorf("leaf index %d not found: %w", i, err)
```
AlCutter, line 296: "bundle at index ..."

passim :)

client/client.go
```diff
@@ -286,27 +286,25 @@ func newTileFetcher(f Fetcher, logSize uint64) GetTileFunc {
...
 		}
-		return nil, fmt.Errorf("failed to fetch leaf index %d: %w", i, err)
+		return bundle, fmt.Errorf("failed to fetch leaf index %d: %w", i, err)
```
AlCutter, line 298: Are there other errors which are part of the API (like `os.ErrNotExist`)? If not you can probably demote the `%w` to a `%v`.

client/client.go
```diff
@@ -286,27 +286,25 @@ func newTileFetcher(f Fetcher, logSize uint64) GetTileFunc {
...
 	}
-	bundle := api.EntryBundle{}
 	if err := bundle.UnmarshalText(sRaw); err != nil {
-		return nil, fmt.Errorf("failed to parse EntryBundle for index %d: %v", i, err)
+		return bundle, fmt.Errorf("failed to parse EntryBundle for index %d: %v", i, err)
 	}
 	offset := int(i % 256)
```
AlCutter, line 303: /offset/expectedSize/? I think this section is a safety/correctness check against what the log returns right?

hammer/client.go
```diff
@@ -0,0 +1,190 @@
...
+		if err != nil {
+			return nil, err
+		}
+		return get(ctx, u)
+	}
+}
+
+var getByScheme = map[string]func(context.Context, *url.URL) ([]byte, error){
+	"http":  readHTTP,
+	"https": readHTTP,
```
AlCutter, line 80: Is it worth supporting the `file:///` scheme too for testing POSIX logs? Could "just" run a busybox HTTP server there, but if we want to use this in CI maybe fewer moving parts is better?

hammer/README.md
```diff
@@ -0,0 +1,27 @@
+# Hammer: A load testing tool for Tessera logs
+
+This hammer sets up read and (optionally) write traffic to a log to test correctness and performance under load.
+The read traffic is sent according to the [tlog-tiles](https://github.com/C2SP/C2SP/blob/main/tlog-tiles.md) spec, and thus could be used to load test any tiles-based log, not just Tessera logs.
+
+If write traffic is enabled, then the target log must support `POST` requests to a `/add` path.
```
AlCutter, line 6: Should we add that the returned body must be an ASCII decimal number representing the assigned index?

hammer/hammer.go
```diff
@@ -0,0 +1,362 @@
...
+	ctx := context.Background()
+
+	logSigV, err := note.NewVerifier(*logPubKey)
+	if err != nil {
+		klog.Exitf("failed to create verifier: %v", err)
+	}
+
+	f, w := newLogClientsFromFlags()
+
+	var cpRaw []byte
```
AlCutter, line 80: There's a catch-22 here at the mo' - Tessera logs don't create a checkpoint until they've added an entry, so the hammer bails here before it can add any entries.
Perhaps the right answer is to have the logs create a zero checkpoint, but it's also sorta special-case there too.

hammer/hammer.go
```diff
@@ -0,0 +1,362 @@
...
+
+func (t *Throttle) Run(ctx context.Context) {
+	ticker := time.NewTicker(1 * time.Second)
+	for {
+		select {
+		case <-ctx.Done(): //context cancelled
+			return
+		case <-ticker.C:
+			tokenCount := t.opsPerSecond
+			timeout := time.After(1 * time.Second)
```
AlCutter, line 260: I think this 1*Second would need to change if the `ticker` changed too - worth factoring the duration out to an `interval` var at the top of the func?

hammer/workerpool.go
```diff
@@ -0,0 +1,52 @@
...
+package main
+
+import "context"
+
+type worker interface {
+	Run(ctx context.Context)
+	Kill()
+}
+
+func newWorkerPool(factory func() worker) workerPool {
```
AlCutter, line 24: Came across this recently, dunno if it's useful here or overkill: https://github.com/alitto/pond
(Not asking you to use it though, just pointing at it in case you want to.)

hammer/workers.go
```diff
@@ -0,0 +1,203 @@
...
+	throttle <-chan bool
+	errchan  chan<- error
+	cancel   func()
+	c        leafBundleCache
+}
+
+// Run runs the log reader. This should be called in a goroutine.
+func (r *LeafReader) Run(ctx context.Context) {
+	if r.cancel != nil {
+		panic("LeafReader was ran multiple times")
```
AlCutter, line 61: /ran/run/

hammer/workers.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+// LeafReader reads leaves from the tree.
+// This class is not thread safe.
+type LeafReader struct {
+	tracker  *client.LogStateTracker
+	f        client.Fetcher
+	next     func(uint64) uint64
+	throttle <-chan bool
+	errchan  chan<- error
```
AlCutter, line 53: nit: `errChan` to be consistent with caps in hammer.go

hammer/workers.go
```diff
@@ -0,0 +1,203 @@
...
+		errchan:  errchan,
+	}
+}
+
+// LogWriter writes new leaves to the log that are generated by `gen`.
+type LogWriter struct {
+	writer   LeafWriter
+	gen      func() []byte
+	throttle <-chan bool
+	errchan  chan<- error
```
AlCutter, line 171: nit: `errChan`

hammer/hammer.go
mhutchinson: We discussed this some weeks ago and the conclusion is that logs ought to create a checkpoint for the 0 size. There was some notion being floated that this was out of bounds from the specs, but it's explicitly in the spec:

https://github.com/C2SP/C2SP/blob/main/tlog-checkpoint.md : `The second line is the tree size, the ASCII decimal representation of the number of leaves in the tree, with no leading zeroes (unless the tree is empty, in which case the tree size is 0).`

hammer/hammer.go
mhutchinson: Done.

hammer/workerpool.go
mhutchinson: Added a comment.

hammer/workers.go
mhutchinson: Done.

hammer/workers.go
mhutchinson: done.

hammer/workers.go
mhutchinson: already passim'd this one :-)

hammer/client.go
AlCutter: Agree from a load-test PoV, but this is probably also the only tool we have currently to test for correctness/conformance.

Happy for it to be added later, though.

AlCutter (approved):

PR #96, opened 2024-07-29
Hammer stats

AlCutter:
this PR adds a few (hopefully) useful stats lines to the top pane: 
- checkpoint size & growth-per-second averaged over the last 30s,
- a "sampled" min-avg-max of how long entries wait to have indices assigned (i.e. `Add()` latency)
- a "sampled" total-time-to-integrate (i.e latency from calling `Add()` to a checkpoint committing to the entry.

```
│Read: Current max: 10/s. Oversupply in last second: 0
│Write: Current max: 10000/s. Oversupply in last second: 1018
│TreeSize: 179489419 (Δ 957qps over 30s)
│Time-in-queue: 981ms/1863ms/3031ms (min/avg/max)
│Total-time-to-integrate: 1399ms/2432ms/4055ms (min/avg/max)
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.22.5
 require (
 	cloud.google.com/go/spanner v1.65.0
 	cloud.google.com/go/storage v1.43.0
+	github.com/RobinUS2/golang-moving-average v1.0.0
 	github.com/gdamore/tcell/v2 v2.7.4
 	github.com/globocom/go-buffer v1.2.2
 	github.com/google/go-cmp v0.6.0
@@ -17,8 +18,6 @@ require (
 	google.golang.org/grpc v1.65.0
 	k8s.io/klog/v2 v2.130.1
 )
-
-require github.com/RobinUS2/golang-moving-average v1.0.0
 
 require (
 	cel.dev/expr v0.15.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	k8s.io/klog/v2 v2.130.1
 )
 
-require github.com/RobinUS2/golang-moving-average v1.0.0 // indirect
+require github.com/RobinUS2/golang-moving-average v1.0.0
 
 require (
 	cel.dev/expr v0.15.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/gdamore/tcell/v2 v2.7.4
 	github.com/globocom/go-buffer v1.2.2
 	github.com/google/go-cmp v0.6.0
+	github.com/mxmCherry/movavg v1.1.0
 	github.com/rivo/tview v0.0.0-20240625185742-b0a7293b8130
 	github.com/transparency-dev/formats v0.0.0-20240715203801-9ff9b9e3905f
 	github.com/transparency-dev/merkle v0.0.2
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,6 @@ require (
 	github.com/gdamore/tcell/v2 v2.7.4
 	github.com/globocom/go-buffer v1.2.2
 	github.com/google/go-cmp v0.6.0
-	github.com/mxmCherry/movavg v1.1.0
 	github.com/rivo/tview v0.0.0-20240625185742-b0a7293b8130
 	github.com/transparency-dev/formats v0.0.0-20240715203801-9ff9b9e3905f
 	github.com/transparency-dev/merkle v0.0.2
@@ -18,6 +17,8 @@ require (
 	google.golang.org/grpc v1.65.0
 	k8s.io/klog/v2 v2.130.1
 )
+
+require github.com/RobinUS2/golang-moving-average v1.0.0 // indirect
 
 require (
 	cel.dev/expr v0.15.0 // indirect
--- a/hammer/hammer.go
+++ b/hammer/hammer.go
@@ -27,8 +27,8 @@ import (
 	"strings"
 	"time"
 
+	movingaverage "github.com/RobinUS2/golang-moving-average"
 	"github.com/gdamore/tcell/v2"
-	"github.com/mxmCherry/movavg"
 	"github.com/rivo/tview"
 	"github.com/transparency-dev/trillian-tessera/client"
 	"golang.org/x/mod/sumdb/note"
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -18,6 +18,8 @@ require (
 	k8s.io/klog/v2 v2.130.1
 )
 
+require github.com/RobinUS2/golang-moving-average v1.0.0 // indirect
```
mhutchinson, line 21: This could do with tidying; I don't see a reason it should be in its own block, and it isn't indirect ?

hammer/hammer.go
```diff
@@ -103,6 +104,7 @@ func NewHammer(tracker *client.LogStateTracker, f client.Fetcher, w LeafWriter)
 	readThrottle := NewThrottle(*maxReadOpsPerSecond)
 	writeThrottle := NewThrottle(*maxWriteOpsPerSecond)
 	errChan := make(chan error, 20)
+	leafSampleChan := make(chan leafTime, 100)
```
mhutchinson, line 107: Maybe `leafWriteStats` instead? It isn't sampled from what I can see, and it only comes from the writer workers.

hammer/hammer.go
```diff
@@ -182,6 +190,26 @@ func (h *Hammer) Run(ctx context.Context) {
...
+						// We've caught up with everything which happened in this "epoch"
+						break
+					}
+					queueLatency += l.assignedAt.Sub(l.queuedAt)
+					totalLatency += now.Sub(l.queuedAt)
+
+					numLeaves++
+				}
+				h.integrationTime.Add(float64(totalLatency/time.Millisecond) / float64(numLeaves))
+				h.queueTime.Add(float64(queueLatency/time.Millisecond) / float64(numLeaves))
```
mhutchinson: How about a separate goroutine for this? I don't think it needs to have a shared fate with the checkpoint updating, and separating the concerns is cleaner.

hammer/hammer.go
```diff
@@ -182,6 +190,26 @@ func (h *Hammer) Run(ctx context.Context) {
...
+				numLeaves := 0
+				for {
+					l, ok := <-h.leafSampleChan
+					if !ok {
+						break
+					}
+					if l.queuedAt.After(now) {
+						// We've caught up with everything which happened in this "epoch"
+						break
+					}
```
mhutchinson: This drops the write stat that was pulled into `l`. Are we happy with that? It doesn't seem like the end of the world, but it probably deserves a comment at least.

mhutchinson (commented):
Submitting comments I have to date because I can't even see 2 of them any more due to changes during the review cycle.

AlCutter:
> Submitting comments I have to date because I can't even see 2 of them any more due to changes during the review cycle.

Sorry -  I didn't realise you were trying to review it; I raised it in DRAFT as a signal that I'm still hacking on it, and only tagged you as an FYI in case you had other plans for how to do this sort of thing.

I can see a couple of comments at least, so I'll work those in.

hammer/hammer.go
AlCutter: well, it's sampled in a very primitive sense in that the channel only takes 100 entries, and if it's full we just drop the sample.

Happy to change that approach though.

hammer/hammer.go
AlCutter: Yeah, good call - done.

Maybe in the future there might be a case for some sort of channel/event based thing to happen on new checkpoints, but for now I've just added another ticker-based loop.

AlCutter:
Thanks for the early comments, ready for another pass now.

hammer/hammer.go
```diff
@@ -185,6 +193,48 @@ func (h *Hammer) Run(ctx context.Context) {
...
+				}
+				queueLatency += sample.assignedAt.Sub(sample.queuedAt)
+				totalLatency += now.Sub(sample.queuedAt)
+
+				numLeaves++
+				sample = nil
+			}
+			if numLeaves > 0 {
+				h.integrationTime.Add(float64(totalLatency/time.Millisecond) / float64(numLeaves))
+				h.queueTime.Add(float64(queueLatency/time.Millisecond) / float64(numLeaves))
```
mhutchinson, line 234: Is there a reason to do it this way instead of posting the row values? This is a bit blurry because it calculates a mean here, and then the moving average will calculate averages from them (if I understand the library right).

hammer/hammer.go
```diff
@@ -185,6 +193,48 @@ func (h *Hammer) Run(ctx context.Context) {
...
+			var sample *leafTime
+			for {
+				if sample == nil {
+					l, ok := <-h.leafSampleChan
+					if !ok {
+						break
+					}
+					sample = &l
+				}
+				if sample.idx >= newSize || sample.assignedAt.After(now) {
```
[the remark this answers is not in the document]
AlCutter: added.

hammer/workers.go
```diff
@@ -185,11 +198,18 @@ func (w *LogWriter) Run(ctx context.Context) {
...
 		index, err := w.writer(ctx, newLeaf)
 		if err != nil {
 			w.errChan <- fmt.Errorf("failed to create request: %v", err)
 			continue
 		}
+		lt.idx, lt.assignedAt = index, time.Now()
+		// See if we can send a leaf sample
+		select {
+		case w.leafChan <- lt:
+		default:
```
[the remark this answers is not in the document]
AlCutter: done

go.mod
```diff
@@ -18,6 +18,8 @@ require (
 	k8s.io/klog/v2 v2.130.1
 )
 
+require github.com/RobinUS2/golang-moving-average v1.0.0
```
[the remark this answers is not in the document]
AlCutter: I have no idea why `go mod tidy` decided to do this, but moving it into another block and re-running doesn't move it back out, so... done for now?

hammer/hammer.go
AlCutter: Mainly just trying to keep the size of the retained data down, and have some sort of loose idea of how the window relates to time even as throughput varies.

I believe that a mean of means is the same as mean of all the things _if_ the subsets are of equal size. Admittedly this is _probably_ not always the case here (certainly if throughput is varying and < 100qps then it's pretty unlikely to be), but tbh with the sampling and "observed" nature of this it's probably all only really good enough for a rough guide to what's going on, but that should be enough to tell you if your log is melting and falling further and further behind, for example.

I'm sure there are better/more precise ways of doing it :)

hammer/hammer.go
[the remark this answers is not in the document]
mhutchinson: Right, this helps. I think the second case can't happen as implemented right now though? It looks like we only feed it down the channel when it's got an index, right?

mhutchinson (approved):

PR #106, opened 2024-08-01
Hammer pushback support

AlCutter:
This PR adds support in the hammer for recognising and reacting to pushback from the target log.

### DIFF
--- a/hammer/client.go
+++ b/hammer/client.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/transparency-dev/trillian-tessera/client"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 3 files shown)


hammer/client.go
```diff
@@ -152,12 +156,20 @@ func (w httpLeafWriter) Write(ctx context.Context, newLeaf []byte) (uint64, erro
...
-	if resp.StatusCode != http.StatusOK {
+	switch resp.StatusCode {
+	case http.StatusOK:
+		if resp.Request.Method != http.MethodPost {
+			return 0, fmt.Errorf("write leaf was redirected to %s", resp.Request.URL)
+		}
+		// Continue below
+	case http.StatusServiceUnavailable, http.StatusBadGateway, http.StatusGatewayTimeout:
+		// These status codes may indicate a delay before retrying, so handle that here:
+		retryDelay(resp.Header.Get("RetryAfter"), time.Second)
```
mhutchinson, line 167: This just computes a value and then throws it away, right? Where does the retry happen?

AlCutter: d'oh, yes - it used to actually delay in the method.

hammer/workers.go
```diff
@@ -201,7 +201,9 @@ func (w *LogWriter) Run(ctx context.Context) {
 		lt := leafTime{queuedAt: time.Now()}
 		index, err := w.writer(ctx, newLeaf)
 		if err != nil {
-			w.errChan <- fmt.Errorf("failed to create request: %v", err)
+			if !errors.Is(ErrRetry, err) {
+				w.errChan <- fmt.Errorf("failed to create request: %w", err)
```
mhutchinson, line 205: I think it deserves a comment to say that as implemented this retry uses a token from the throttle for each attempt. So if we had a throttle of 100 writes per second, but one thread retried 99 times in a second, we'd only have 1 leaf written and no errors propagated down `w.errChan` to explain where all the writes were going.

mhutchinson (approved):

PR #109, opened 2024-08-01
Performance tweaks for GCP

AlCutter:
This PR adjusts behaviour of the GCP storage implementation slightly to improve performance, and avoid having to redo work due to GCS update limits.

#23

### DIFF
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -58,7 +58,7 @@ const (
 	entryBundleSize = 256
 
 	DefaultPushbackMaxOutstanding = 4096
-	DefaultIntegrationSizeLimit   = 2048
+	DefaultIntegrationSizeLimit   = 5 * 4096
 )
 
 // Storage is a GCP based storage implementation for Tessera.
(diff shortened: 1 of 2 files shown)


storage/gcp/gcp.go
```diff
@@ -145,16 +145,12 @@ func New(ctx context.Context, cfg Config, opts ...func(*tessera.StorageOptions))
...
-				defer cancel()
-				if more, err := r.sequencer.consumeEntries(cctx, DefaultIntegrationSizeLimit, r.integrate); err != nil {
-					klog.Errorf("integrate: %v", err)
-					break
-				} else if !more {
-					break
-				}
-				klog.V(1).Info("Quickloop")
+			// Don't quicklook for now, it causes issues updating checkpoint too frequently.
+			cctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
roger2hk, line 149: Is `10*time.Second` configurable? Is this part going to be reworked later to optimize the checkpoint update frquency?

roger2hk (approved):

storage/gcp/gcp.go
AlCutter: Not currently, this is really just a "fuse" to blow if some call to GCP gets wedged for whatever reason; things will retry and presumably continue as normal once whatever network/service issue has been resolved. 
I can't really imagine it needing to be greater than 10s, but we can always add a config option later on if it becomes necessary.

PR #110, opened 2024-08-01
Add Go tests in MySQL storage implementation

roger2hk:
#21 

Note that this pull request also makes `tessera.WithCheckpointSignerVerifier` optional in `mysql.New()` because the issue was found during the tests development.

### DIFF
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -56,6 +57,7 @@ type Storage struct {
 }
 
 // New creates a new instance of the MySQL-based Storage.
+// Note that `tessera.WithCheckpointSignerVerifier()` is mandatory in the `opts` argument.
 func New(ctx context.Context, db *sql.DB, opts ...func(*tessera.StorageOptions)) (*Storage, error) {
 	opt := tessera.ResolveStorageOptions(&tessera.StorageOptions{
 		BatchMaxAge:  defaultQueueMaxAge,
@@ -70,37 +72,38 @@ func New(ctx context.Context, db *sql.DB, opts ...func(*tessera.StorageOptions))
 		klog.Errorf("Failed to ping database: %v", err)
 		return nil, err
 	}
+	if s.newCheckpoint == nil || s.parseCheckpoint == nil {
+		return nil, errors.New("tessera.WithCheckpointSignerVerifier must be provided in New()")
+	}
 
 	s.queue = storage.NewQueue(ctx, opt.BatchMaxAge, opt.BatchMaxSize, s.sequenceBatch)
 
 	// Initialize checkpoint if there is no row in the Checkpoint table.
-	if s.newCheckpoint != nil && s.parseCheckpoint != nil {
-		if _, err := s.ReadCheckpoint(ctx); err != nil {
-			if err != sql.ErrNoRows {
-				klog.Errorf("Failed to read checkpoint: %v", err)
-				return nil, err
-			}
+	if _, err := s.ReadCheckpoint(ctx); err != nil {
+		if err != sql.ErrNoRows {
+			klog.Errorf("Failed to read checkpoint: %v", err)
+			return nil, err
+		}
 
-			klog.Infof("Initializing checkpoint")
-			// Get a Tx for making transaction requests.
-			tx, err := s.db.BeginTx(ctx, nil)
-			if err != nil {
-				return nil, err
+		klog.Infof("Initializing checkpoint")
+		// Get a Tx for making transaction requests.
+		tx, err := s.db.BeginTx(ctx, nil)
+		if err != nil {
+			return nil, err
+		}
+		// Defer a rollback in case anything fails.
+		defer func() {
+			if err := tx.Rollback(); err != nil && err != sql.ErrTxDone {
+				klog.Errorf("Failed to rollback in write initial checkpoint: %v", err)
 			}
-			// Defer a rollback in case anything fails.
-			defer func() {
-				if err := tx.Rollback(); err != nil && err != sql.ErrTxDone {
-					klog.Errorf("Failed to rollback in write initial checkpoint: %v", err)
-				}
-			}()
-			if err := s.writeCheckpoint(ctx, tx, 0, rfc6962.DefaultHasher.EmptyRoot()); err != nil {
-				klog.Errorf("Failed to write initial checkpoint: %v", err)
-				return nil, err
-			}
-			// Commit the transaction.
-			if err := tx.Commit(); err != nil {
-				return nil, err
-			}
+		}()
+		if err := s.writeCheckpoint(ctx, tx, 0, rfc6962.DefaultHasher.EmptyRoot()); err != nil {
+			klog.Errorf("Failed to write initial checkpoint: %v", err)
+			return nil, err
+		}
+		// Commit the transaction.
+		if err := tx.Commit(); err != nil {
+			return nil, err
 		}
 	}
 
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -57,6 +57,7 @@ type Storage struct {
 }
 
 // New creates a new instance of the MySQL-based Storage.
+// Note that `tessera.WithCheckpointSignerVerifier()` is mandatory in the `opts` argument.
 func New(ctx context.Context, db *sql.DB, opts ...func(*tessera.StorageOptions)) (*Storage, error) {
 	opt := tessera.ResolveStorageOptions(&tessera.StorageOptions{
 		BatchMaxAge:  defaultQueueMaxAge,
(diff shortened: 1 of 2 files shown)


roger2hk:
Would the `655536` be too large in the `TestParallelAdd` test?

storage/mysql/mysql_test.go
```diff
@@ -28,14 +63,253 @@ func TestMain(m *testing.M) {
...
+			wantErr: false,
+		},
+		{
+			name:    "empty byte",
+			entry:   []byte{},
+			wantErr: false,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			for i := 0; i < 655536; i++ {
```
AlCutter, line 293: Typo - 65536? That's still quite a number of goroutines to run...
What's the goal of this test/which functionality of the MySQL implementation are you wanting to exercise here?

storage/mysql/mysql_test.go
```diff
@@ -28,14 +63,253 @@ func TestMain(m *testing.M) {
...
+
+func TestParallelAdd(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	s := newTestMySQLStorage(t, ctx)
+
+	for _, test := range []struct {
+		name    string
+		entry   []byte
+		wantErr bool
```
AlCutter, line 274: This is never `true`, could it be removed?

storage/mysql/mysql_test.go
```diff
@@ -28,14 +63,253 @@ func TestMain(m *testing.M) {
...
+
+func TestAdd(t *testing.T) {
+	ctx := context.Background()
+	s := newTestMySQLStorage(t, ctx)
+
+	for _, test := range []struct {
+		name      string
+		entry     []byte
+		wantIndex uint64
+		wantErr   bool
```
AlCutter, line 232: This is always `false`, could it be removed?

storage/mysql/mysql_test.go
```diff
@@ -28,14 +63,253 @@ func TestMain(m *testing.M) {
...
+				t.Errorf("got index %d want %d", index, test.wantIndex)
+			}
+		})
+	}
+}
+
+func TestParallelAdd(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	s := newTestMySQLStorage(t, ctx)
```
AlCutter, line 269: Currently, this func will give you a storage configured with batching which flushes after 1 microsecond, is that intended?

storage/mysql/mysql_test.go
```diff
@@ -28,14 +63,253 @@ func TestMain(m *testing.M) {
...
+			name:    "no tessera.StorageOption",
+			opts:    nil,
+			wantErr: true,
+		},
+		{
+			name: "standard tessera.WithCheckpointSignerVerifier",
+			opts: []func(*tessera.StorageOptions){
+				tessera.WithCheckpointSignerVerifier(noteSigner, noteVerifier),
+			},
+			wantErr: false,
```
AlCutter, line 142: nit: can just leave these out and rely on the default value of bool being `false`.

storage/mysql/mysql.go
```diff
@@ -70,6 +71,9 @@ func New(ctx context.Context, db *sql.DB, opts ...func(*tessera.StorageOptions))
 		klog.Errorf("Failed to ping database: %v", err)
 		return nil, err
 	}
+	if s.newCheckpoint == nil || s.parseCheckpoint == nil {
```
AlCutter, line 74: Is this a required option here because this storage implementation doesn't track the tree size separately from the stored `checkpoint`?

Probably should say that this is a requirement for _this_ storage impl on the c'tor comment.

storage/mysql/mysql_test.go
```diff
@@ -28,14 +63,253 @@ func TestMain(m *testing.M) {
...
+
+	db, err := sql.Open("mysql", *mysqlURI+"?multiStatements=true")
+	if err != nil {
+		klog.Fatalf("Failed to connect to DB: %v", err)
+	}
+	defer func() {
+		if err := db.Close(); err != nil {
+			klog.Warningf("Failed to close db: %v", err)
+		}
+	}()
```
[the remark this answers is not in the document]
roger2hk: This is intended since `multiStatements=true` is only required during the database schema import.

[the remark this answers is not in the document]
AlCutter: I don't really understand what that means, but it sounds important :D 
Worth popping a comment in so that future folks don't try to simplify the code?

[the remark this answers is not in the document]
roger2hk: Good idea. Added some comment.

storage/mysql/mysql_test.go
roger2hk: Done.

storage/mysql/mysql_test.go
roger2hk: That was intended to shorten the development time of the tests. Removed this option.

storage/mysql/mysql.go
roger2hk: In the current MySQL implementation, the `tessera.WithCheckpointSignerVerifier` option is mandatory. I've added comments to the `mysql.New()` method.

Is there any scenario that doesn't require the signer and verifier? Let's say we mirror the log. Is it necessary to sign the checkpoint with another pairs of keys?

storage/mysql/mysql_test.go
roger2hk: Done.

storage/mysql/mysql_test.go
roger2hk: Done.

storage/mysql/mysql.go
AlCutter: Yeah, that's one scenario - for mirrors we don't have access to the source's private key, but I can't imagine we'd want to sign with a separate key.

roger2hk: Gotcha. I've updated the logic to make `tessera.WithCheckpointSignerVerifier` opt optional in `mysql.New()`.

storage/mysql/mysql_test.go
```diff
@@ -28,14 +63,253 @@ func TestMain(m *testing.M) {
...
+			_, err := mysql.New(ctx, testDB, test.opts...)
+			gotErr := err != nil
+			if gotErr != test.wantErr {
+				t.Errorf("got err %v want %v", gotErr, test.wantErr)
+			}
+		})
+	}
+}
+
+func TestReadTile(t *testing.T) {
```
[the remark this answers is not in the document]
roger2hk: Renamed to `TestReadMissingTile` and `TestReadMissingEntryBundle`. Added `nil` check to these two tests.

Added `TestTileRoundTrip` and `TestEntryBundleRoundTrip` tests.

storage/mysql/mysql_test.go
```diff
@@ -28,14 +67,335 @@ func TestMain(m *testing.M) {
...
+			entry:     []byte("123"),
+			wantIndex: 1,
+		},
+		{
+			name:      "empty byte",
+			entry:     []byte{},
+			wantIndex: 2,
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
```
mhutchinson, line 248: These test case functionally depend on each other for the `wantIndex` value, so I suggest that we remove the pretense that they are separate test cases? It is idiomatic in my experience with go that table-driven tests each start with a clean environment.

storage/mysql/mysql.go
```diff
@@ -74,31 +74,33 @@ func New(ctx context.Context, db *sql.DB, opts ...func(*tessera.StorageOptions))
 	s.queue = storage.NewQueue(ctx, opt.BatchMaxAge, opt.BatchMaxSize, s.sequenceBatch)
 
 	// Initialize checkpoint if there is no row in the Checkpoint table.
-	if _, err := s.ReadCheckpoint(ctx); err != nil {
-		if err != sql.ErrNoRows {
-			klog.Errorf("Failed to read checkpoint: %v", err)
-			return nil, err
-		}
+	if s.newCheckpoint != nil && s.parseCheckpoint != nil {
```
mhutchinson, line 77: This makes me quite uneasy. I think we should require these fields until we have a concrete use-case (that is tested) for doing it. At the moment `New` works if these are not present, but any attempt to integrate leaves looks like it will fail.

storage/mysql/mysql_test.go
roger2hk: Gotcha, I've removed `TestAdd`. `TestParallelAdd` has the test coverage for `mysql.Stoage.Add`.

storage/mysql/mysql.go
roger2hk: From the [discussion](https://github.com/transparency-dev/trillian-tessera/pull/110#discussion_r1701644012) with @AlCutter, there is a scenario for mirrors we don't have access to the source's private key. That is why I updated the logic when I work on the test cases.

I've updated the logic to require these fields in `New`.

mhutchinson: Yeah I saw that conversation. My philosophy is that this precondition should be weakened at the point where we're going to add (and test!) the feature to support mirrors. Weakening the guarantee of the state now but not using it and exercising it leaves it in a more brittle state.

mhutchinson (approved):

PR #117, opened 2024-08-07
Refactor TUI to separate setup and runtime controls

mhutchinson:
This makes it a bit more readable now, but really sets this up to grow in a way that doesn't turn into one enormous method with all of the UI logic in it.

### DIFF
--- a/hammer/tui.go
+++ b/hammer/tui.go
@@ -26,97 +26,120 @@ import (
 	"k8s.io/klog/v2"
 )
 
-func hostUI(ctx context.Context, hammer *Hammer) {
+type tuiController struct {
+	hammer     *Hammer
+	app        *tview.Application
+	statusView *tview.TextView
+	logView    *tview.TextView
+	helpView   *tview.TextView
+}
+
+func newController(h *Hammer) *tuiController {
+	c := tuiController{
+		hammer: h,
+		app:    tview.NewApplication(),
+	}
 	grid := tview.NewGrid()
 	grid.SetRows(5, 0, 10).SetColumns(0).SetBorders(true)
-	// Status box
+
+	// Top: status box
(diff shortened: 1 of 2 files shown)


hammer/tui.go
```diff
@@ -26,97 +26,120 @@ import (
...
 	logView.ScrollToEnd()
 	logView.SetMaxLines(10000)
 	grid.AddItem(logView, 1, 0, 1, 1, 0, 0, false)
+	c.logView = logView
+
+	// Bottom: help text
+	helpView := tview.NewTextView()
+	helpView.SetText("+/- to increase/decrease read load\n>/< to increase/decrease write load\nw/W to increase/decrease workers")
+	grid.AddItem(helpView, 2, 0, 1, 1, 0, 0, false)
+	c.helpView = helpView
```
roger2hk, line 61: Is it necessary to create the local variables `statusView`, `logView` and `helpView`?

mhutchinson: I agree it's not necessary but it's also not harmful. Personally I like the separation of having the local variables be used for construction, and then assign when it's finished, but it doesn't matter. I'll leave it this way.

mhutchinson:
> I'm always tempted to use the uppercase (`W`) to increase the number of workers. Is it reasonable to swap `w` and `W`?

I originally had it the other way around because that was more logical to me, but after using it the first few times I found that the most common use-case is to increase workers and having fewer keypresses (avoiding shift modifier) made that more comfortable. Feel free to raise an issue and we can get a vote for each system :-) Or we can just have two different keys for this?

> It would be great if we can see the number of workers in the UI.

I'll see if we can get the number of workers displayed in the UI later today. I want this PR to stay as refactoring only. Keeping refactoring and behaviour changes separate is a really good practice to get into.

PR #122, opened 2024-08-07
move c-t-go/util/timesource.go to sctfe package

phbnf:
I don't see any reason why it would be in its own package. Packages named util are discouraged, https://go.dev/blog/package-names#bad-package-names, let's get rid of it.

### DIFF
--- a/personalities/sctfe/timesource.go
+++ b/personalities/sctfe/timesource.go
@@ -1,0 +1,48 @@
+// Copyright 2016 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


personalities/sctfe/timesource.go
```diff
@@ -0,0 +1,48 @@
...
+
+// NewFixedTimeSource creates a FixedTimeSource instance
+func NewFixedTimeSource(t time.Time) *FixedTimeSource {
+	return &FixedTimeSource{fakeTime: t}
+}
+
+// Now returns the time value this instance contains
+func (f *FixedTimeSource) Now() time.Time {
+	return f.fakeTime
+}
```
roger2hk, line 48: Move these to the test file (e.g. timesource_test.go)?

roger2hk (approved):

personalities/sctfe/timesource.go
phbnf: Good idea. Since it wouldn't be a testfile to test timesource.go, I'll move them to a testfile called `export_test.go`

PR #126, opened 2024-08-08
Pass a signer to Tessera

phbnf:
This PR makes it possible to sign https://c2sp.org/static-ct-api checkpoints, and passes a signer to Tessera.

Before we can pass a Checkpoint signer to Tessera, we need to initiate the time source and crypto.signer, so I had to re-architecture the code a bit before that.

### DIFF
--- a/personalities/sctfe/serialize.go
+++ b/personalities/sctfe/serialize.go
@@ -133,8 +133,7 @@ func (cts *CpSigner) Sign(msg []byte) ([]byte, error) {
 		return nil, fmt.Errorf("checkpoint contains trailing data: %s", string(rest))
 	} else if err != nil {
 		return nil, fmt.Errorf("ckpt.Unmarshal: %v", err)
-	}
-	if ckpt.Origin != cts.origin {
+	} else if ckpt.Origin != cts.origin {
 		return nil, fmt.Errorf("checkpoint's origin %s doesn't match signer's origin %s", ckpt.Origin, cts.origin)
 	}
 
--- a/personalities/sctfe/serialize.go
+++ b/personalities/sctfe/serialize.go
@@ -18,9 +18,12 @@ import (
 	"crypto"
 	"crypto/rand"
 	"crypto/sha256"
+	"encoding/binary"
 	"fmt"
 
 	"github.com/google/certificate-transparency-go/tls"
+	"github.com/transparency-dev/formats/log"
+	"golang.org/x/mod/sumdb/note"
 
 	ct "github.com/google/certificate-transparency-go"
 )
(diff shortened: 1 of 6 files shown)


personalities/sctfe/serialize.go
```diff
@@ -64,3 +67,110 @@ func buildV1SCT(signer crypto.Signer, leaf *ct.MerkleTreeLeaf) (*ct.SignedCertif
...
+	rest, err := ckpt.Unmarshal(msg)
+
+	if len(rest) != 0 {
+		return nil, fmt.Errorf("checkpoint contains trailing data: %s", string(rest))
+	} else if err != nil {
+		return nil, fmt.Errorf("ckpt.Unmarshal: %v", err)
+	}
+	if ckpt.Origin != cts.origin {
+		return nil, fmt.Errorf("checkpoint's origin %s doesn't match signer's origin %s", ckpt.Origin, cts.origin)
+	}
```
roger2hk, line 139: nit: The `if` and `else if` blocks are not consistent here although the logic is the same.

personalities/sctfe/serialize.go
```diff
@@ -64,3 +67,110 @@ func buildV1SCT(signer crypto.Signer, leaf *ct.MerkleTreeLeaf) (*ct.SignedCertif
...
+	// TODO(phboneff): make sure that it's ok to generate the timestamp here
+	t := uint64(cts.timeSource.Now().UnixMilli())
+	sig, err := buildCTCheckpoint(cts.sthSigner, ckpt.Size, t, ckpt.Hash[:])
+	if err != nil {
+		return nil, fmt.Errorf("coudn't sign CT checkpoint: %v", err)
+	}
+	return sig, nil
+}
+
+func (cts *CTSigner) Name() string {
```
roger2hk, line 150: How about calling this `Origin()` instead of `Name()`?

personalities/sctfe/serialize.go
```diff
@@ -64,3 +67,110 @@ func buildV1SCT(signer crypto.Signer, leaf *ct.MerkleTreeLeaf) (*ct.SignedCertif
...
+
+	sig, err := tls.Marshal(rfc6962Note)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't encode RFC6962NoteSignature: %w", err)
+	}
+
+	return sig, nil
+}
+
+type CTSigner struct {
```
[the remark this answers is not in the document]
phbnf: Yes, there's also the signer used for SCTs, it's the signer that is passed as an argument to NewCTSigner. I can rename things to CSigner

personalities/sctfe/serialize.go
phbnf: Ah, I see where you're coming from, but Signer must implement the note signer interface with a Name() method: https://cs.opensource.google/go/x/mod/+/refs/tags/v0.20.0:sumdb/note/note.go;l=204

roger2hk: Gotcha, thanks!

personalities/sctfe/serialize.go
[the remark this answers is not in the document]
roger2hk: Oh okay, `CTSigner` sounds good to me.

personalities/sctfe/serialize.go
```diff
@@ -64,3 +67,110 @@ func buildV1SCT(signer crypto.Signer, leaf *ct.MerkleTreeLeaf) (*ct.SignedCertif
...
 }
+
+type RFC6962NoteSignature struct {
+	timestamp uint64
+	signature ct.DigitallySigned
+}
+
+// buildCTCheckpoints builds a https://c2sp.org/static-ct-api checkpoint.
+// TODO(phboneff): add tests
+func buildCTCheckpoint(signer crypto.Signer, size uint64, timeMilli uint64, hash []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
phbnf: Good point, thanks.

personalities/sctfe/serialize.go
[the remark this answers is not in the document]
phbnf: I've gone for CpSigner

personalities/sctfe/serialize.go
phbnf: I've added a doc to the struct to make this more readable.

roger2hk (approved):

PR #131, opened 2024-08-10
Fix incorrect http status not found behaviour in MySQL storage

roger2hk:
#21

### DIFF
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -111,6 +111,7 @@ func New(ctx context.Context, db *sql.DB, opts ...func(*tessera.StorageOptions))
 }
 
 // ReadCheckpoint returns the latest stored checkpoint.
+// If the checkpoint is not found, nil is returned with no error.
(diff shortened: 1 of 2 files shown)


storage/mysql/mysql.go
```diff
@@ -118,7 +118,13 @@ func (s *Storage) ReadCheckpoint(ctx context.Context) ([]byte, error) {
 	}
 
 	var checkpoint []byte
-	return checkpoint, row.Scan(&checkpoint)
+	if err := row.Scan(&checkpoint); err != nil {
+		if err == sql.ErrNoRows {
+			return nil, nil
```
mhutchinson, line 123: Please document this behaviour in the method comment if this is part of the contract (i.e. document that no checkpoint will not return an error, but will return an empty byte slice).

mhutchinson (approved):

storage/mysql/mysql.go
roger2hk: Added method comments to `ReadCheckpoint`, `ReadTile` and `ReadEntryBundle`.

Issue #88, opened 2024-07-25
CT static API personality


phbnf:
Implements https://c2sp.org/static-ct-api

Pieces of work:
- [ ] port certificate-transparency-go CTFE to this repo, removing all the things we don't need anymore, and connect it to GCP Tessera to write tiles (no checkpoint at this stage)
  - [x] break ground with an initial port
  - [x] move more things over to drop dependency on non X509/TLS stuff from `github.com/google/certificate-transparency-go` (for instance `github.com/google/certificate-transparency-go/trillian/util`, `github.com/google/certificate-transparency-go/trillian/testdata`)
  - [ ] double check that tiles and leaf are compliant with https://c2sp.org/static-ct-api
    - [x] Fix Data Type
    - [x] Store fingerprints chain in leaves 
  - [ ] have a look through all the old TODOs I ported over
  - [x] move over handlers_test.go
  - [x] add README and/or docker config to generate proto and mocks
- [x] add a signer option
- [x] design+build a chain storage service (see what we can re-use from Roger's recent work on certificate-transparengy-go CTFE): see #132
- [ ] design+build deduplication: ongoing
- [x] design+modify configuration pipeline if need be: we've decided to stick to flags
- [x] quota system: we've decided not rely on Tessera's pushbacks to start with. We'll add a quota system later if needed. See #186 
- [ ] decide how we handle errors, and which one we pass to clients
- [x] migrate codebase to a different repo
- [ ] load test with real world CT data
- [ ] deployment instructions

mhutchinson:
What's the difference between this issue and #41 ?

AlCutter:
I think #41 is Tessera APIs which _enable_ a CT Static API personality to be built, and this one is the task of actually building such a personality.

phbnf:
I'm going to close this issue to migrate it under https://github.com/transparency-dev/static-ct, where the SCTFE now lives.

phbnf:
https://github.com/transparency-dev/static-ct/issues/10

PR #132, opened 2024-08-12
Store issuer chains

phbnf:
#88

### DIFF
--- a/personalities/sctfe/storage.go
+++ b/personalities/sctfe/storage.go
@@ -28,9 +28,9 @@ import (
 
 // Storage provides all the storage primitives necessary to write to a ct-static-api log.
 type Storage interface {
-	// Add assign an index to the provided Entry, stages the entry for integration, and return it the assigned index.
+	// Add assigns an index to the provided Entry, stages the entry for integration, and return it the assigned index.
 	Add(context.Context, *ctonly.Entry) (uint64, error)
-	// AddIssuerChain stores all certificates in the chain in a content-addressable store under their sha256 hash.
+	// AddIssuerChain stores every the chain certificate in a content-addressable store under their sha256 hash.
 	AddIssuerChain(context.Context, []*x509.Certificate) error
 }
 
@@ -60,7 +60,7 @@ func (cts *CTStorage) Add(ctx context.Context, entry *ctonly.Entry) (uint64, err
 	return cts.storeData(ctx, entry)
 }
 
-// AddIssuerChain stores every certificate in the chain under its sha256.
+// AddIssuerChain stores every chain certificate under its sha256.
 // If an object is already stored under this hash, continues.
 func (cts *CTStorage) AddIssuerChain(ctx context.Context, chain []*x509.Certificate) error {
 	errG := errgroup.Group{}
--- a/personalities/sctfe/storage.go
+++ b/personalities/sctfe/storage.go
@@ -16,34 +16,75 @@ package sctfe
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
 
+	"github.com/google/certificate-transparency-go/x509"
 	tessera "github.com/transparency-dev/trillian-tessera"
 	"github.com/transparency-dev/trillian-tessera/ctonly"
+	"golang.org/x/sync/errgroup"
 )
 
 // Storage provides all the storage primitives necessary to write to a ct-static-api log.
 type Storage interface {
-	// 	Add assign an index to the provided Entry, stages the entry for integration, and return it the assigned index.
+	// Add assign an index to the provided Entry, stages the entry for integration, and return it the assigned index.
 	Add(context.Context, *ctonly.Entry) (uint64, error)
+	// AddIssuerChain stores all certificates in the chain in a content-addressable store under their sha256 hash.
+	AddIssuerChain(context.Context, []*x509.Certificate) error
+}
+
+type IssuerStorage interface {
+	Exists(ctx context.Context, key [32]byte) (bool, error)
+	Add(ctx context.Context, key [32]byte, data []byte) error
 }
 
 // CTStorage implements Storage.
 type CTStorage struct {
 	storeData func(context.Context, *ctonly.Entry) (uint64, error)
-	// TODO(phboneff): add storeExtraData
-	// TODO(phboneff): add dedupe
+	issuers   IssuerStorage
 }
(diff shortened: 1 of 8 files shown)


personalities/sctfe/storage.go
```diff
@@ -16,34 +16,77 @@ package sctfe
...
 	// TODO(phboneff): add deduplication and chain storage
 	return cts.storeData(ctx, entry)
 }
+
+// AddIssuerChain stores every chain certificate under its sha256.
+// If an object is already stored under this hash, continues.
+func (cts *CTStorage) AddIssuerChain(ctx context.Context, chain []*x509.Certificate) error {
+	errG := errgroup.Group{}
+	for _, c := range chain {
+		errG.Go(func() error {
```
AlCutter, line 68: Did you consider pushing this down into the storage implementation to decide how best to store these entries? (e.g. some storage infra may do better with some sort of non-transactional batch insert as opposed to several concurrent write transactions).

May be more of an issue for fresh logs and become less so once the chain storage is close to being maximally populated.

personalities/sctfe/storage/gcp/map.go
```diff
@@ -0,0 +1,116 @@
+// Copyright 2016 Google LLC. All Rights Reserved.
```
AlCutter, line 1: Your time-machine is showing :)

personalities/sctfe/storage/gcp/map.go
```diff
@@ -0,0 +1,116 @@
...
+	"net/http"
+	"path"
+
+	gcs "cloud.google.com/go/storage"
+	"google.golang.org/api/googleapi"
+	"google.golang.org/api/iterator"
+	"k8s.io/klog/v2"
+)
+
+// GCSStorage is a map backed by GCS on GCP.
```
AlCutter, line 32: map -> "key value store"?

personalities/sctfe/storage.go
phbnf: Storage implementations can, if they want to, stage all the requests and then batch write them if they want to. Each go routine will hang until then. That would be batched across all AddChain requests though, which means that two issuers.Add for the same chain might not be in the same batch.

I could change the API to accept multiple k,v instead, but since keys can't be of type []byte, I'd need to pass a map[string][]byte, or two [][]byte that would be meant to track one another, or a map[string][]byte. Can you think of a better way of doing this?

The PR that will come after this adds a local cache. That means that issuers.Add won't necessarily be called on all the issuers. So if we modify the api to get multiple k,v I'd have to batch caching as well, and then call issuers.Add with a potential "skipped chain".

It feels to me that it would add a non trivial amount of complexity to me. What do you think?

AlCutter: > Storage implementations can, if they want to, stage all the requests and then batch write them if they want to. Each go routine will hang until then. That would be batched across all AddChain requests though, which means that two issuers.Add for the same chain might not be in the same batch.

I agree that storage implementations should be free to batch up writes if that's what best for them, but your proposal here is a multiplier for the goroutines which are serving the HTTP requests these certs came from - i.e. if the queue is configured to flush at 1000 entries, there will be 1000 goroutines parked on the futures, if each of those chains has 3 entries then with this change we'll have 1000+3*1000 parked goroutines.

Obvs that's fine if that really is the best way for the storage impl to handle the calls to Set, but if we're saying that storage impls which would prefer to use a native "batch" write have to then block and "fan-in" all those extra 3000 goroutines in order to get back to just having a slice of entries - that seems like we've let an implementation detail of "lots of goroutines" type storage types leak out into this higher level.

Feels like the "fan-out" should be done by the storage implementation iff that's what's needed?

> I could change the API to accept multiple k,v instead, but since keys can't be of type []byte, I'd need to pass a map[string][]byte, or two [][]byte that would be meant to track one another, or a map[string][]byte. Can you think of a better way of doing this?

Hmm, could you pass a `[]struct{K []byte, V []byte}`?

As you say, these would need to be independently added rather than transactional (i.e. I think that means the semantics of the API becomes a bit more `AddTheseIntermediates()` rather than `AddIssuerChain()`), both for local caching reasons, but also because distributed systems in the case where there's a central shared intermediates cache.

phbnf: Yeah - I wanted to avoid defining a new struct because it leads to more dependencies. Go lacks tuples :/
I had a go at it, (still need to add comments and co), but have a look at let me know what you think. I could go even further an, leave it to the storage implementation to do the "Exists" calls.

While I'm starting with the SCTFE here, I'm thinking of making deduplication a feature that other personalities can use. That probably means that we'll need to have a base "personalitystorage" package, where KV will be defined?

phbnf: I think it's either this, or a map[string]string / map[string][]byte, and we do some type conversion.

AlCutter: I think it's probably ok for the storage impls to take a dep on a struct here - they have to meet the implicit interface requirement specific here anyway.

I think having the storage do the `Exists` inline with the `Add` would be good, they might be able to optimize that away - e.g. GCS might do a single WriteIfNotExists and save a round-trip, etc.

For now, I'd keep the chain storage and dedup storages separate and have one do `[]byte -> []byte` and the other do `[]byte -> uint64` natively, rather than prematurely push them together just yet.

phbnf: > I think having the storage do the Exists inline with the Add would be good, they might be able to optimize that away - e.g. GCS might do a single WriteIfNotExists and save a round-trip, etc.

Maybe, yes. Not 100% sure since write are more expensive than reads, and eventually the list issuers will stop growing, so it will be cheaper to check for existence first. But also, caching should make all of this even cheaper. There's already a TODO somewhere to evaluate this. I've also left a TODO for parallel write, let's keep them out for now. I've pushed it down to the storage layer.

> For now, I'd keep the chain storage and dedup storages separate and have one do []byte -> []byte and the other do []byte -> uint64 natively, rather than prematurely push them together just yet.
In practice, I had written a wrapper around map.go to handle the conversion: https://github.com/phbnf/trillian-tessera/blob/dedup/personalities/sctfe/storage/gcp/dedup.go. Honestly, I'm not even super attached to this dedup code on GCS, I don't think any CT log will use it, it doesn't scale price-wise. It might be useful for a test personality though or other applications once deduplication becomes a "standard" personality plugin.

AlCutter (approved):

PR #136, opened 2024-08-14
Explicitly Set content types on tiles and leaves

phbnf:
If not specified, GCS automatically detects the content type, so let's be explicit and avoid unfortunate divergence outside of our control.

I haven't added tests, but since there was some divergence before https://github.com/C2SP/C2SP/pull/92, it proves that it can happen, and doesn't sound crazy to add some. I can do this in a later PR but we'd also need to modify the getObject signature, which sounds a bit cumbersome.

### DIFF
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -56,6 +56,8 @@ import (
 
 const (
 	entryBundleSize = 256
+	appDataContType = "application/data"
+	txtContType     = "text/plain; charset=utf-8"
 
 	DefaultPushbackMaxOutstanding = 4096
 	DefaultIntegrationSizeLimit   = 5 * 4096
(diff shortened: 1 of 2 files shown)


AlCutter:
As an aside, I'm not sure what `application/data` is - it's not one of the IANA registered types: https://www.iana.org/assignments/media-types/media-types.xhtml#application

I'm checking with Filippo what he intended there, but I've speculatively raised C2SP/C2SP#92 against the specs to change to `application/octet-stream` which [_is_ a registered type](https://www.iana.org/assignments/media-types/application/octet-stream).

I agree that being explicit here is >> than relying on auto-detection which may change at any time, though!

Do you want to leave this PR open for the moment until we've clarified the type?

phbnf:
Ah you beat me to it I was about to speak with Martin and raise and issue after lunch if need be. My plan was to submit this to align with specs, and to change this later if the specs change.

phbnf:
I've updated the PR title, but I'll leave the description as it is to keep history and context.

AlCutter:
I think it'd be better to update the description to match what the PR does now - this is going to be what's associated with the squashed commit when it's merged so better to have it describe the "atomic whole" rather than a point-in-time history.  The evolution is available for archeology via this PR if anyone needs the context.

AlCutter (approved):

AlCutter (approved):

PR #141, opened 2024-08-15
Hammer: exit criteria

mhutchinson:
The hammer can be configured to stop successfully if a desired number of leaves are written, and to fail after a max runtime is reached. These two can be used in conjunction to run a headless test for write performance.

### DIFF
--- a/hammer/hammer.go
+++ b/hammer/hammer.go
@@ -113,7 +113,7 @@ func main() {
 					return
 				case <-tick.C:
 					if tracker.LatestConsistent.Size >= goal {
-						elapsed := time.Now().Sub(startTime)
+						elapsed := time.Since(startTime)
 						klog.Infof("Reached tree size goal of %d after %s; exiting", goal, elapsed)
 						cancel()
 						return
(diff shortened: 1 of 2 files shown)


hammer/hammer.go
```diff
@@ -97,12 +100,52 @@ func main() {
 	hammer := NewHammer(&tracker, f.Fetch, w.Write, gen, ha.seqLeafChan, ha.errChan)
 	hammer.Run(ctx)
 
+	exitCode := 0
+	if *leafWriteGoal > 0 {
+		go func() {
+			startTime := time.Now()
+			goal := tracker.LatestConsistent.Size + uint64(*leafWriteGoal)
```
AlCutter, line 107: Is there a race here since `hammer.Run()` has already been called?
(Probably not _all that important_ if there is, but might lead to a "huh, why 2504?!" moment if the leaf generators were ever augmented to stop after creating `*leafWriteGoal` leaves.)

AlCutter (approved):

hammer/hammer.go
mhutchinson: Fixed by moving the hammer run until after this block.

PR #142, opened 2024-08-15
Run hammer against MySQL in CI

mhutchinson:
This ensures a basic level of write performance. It isn't a perfect performance benchmark because the job could be run on different specs of machine, but it's a first foray into getting the hammer running in CI.

### DIFF
--- a/.github/workflows/benchmark_mysql.yml
+++ b/.github/workflows/benchmark_mysql.yml
@@ -1,0 +1,21 @@
+name: Benchmark MySQL
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  benchmark-mysql:
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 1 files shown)


AlCutter (approved):

.github/workflows/benchmark_mysql.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  benchmark-mysql:
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: Start Tessera
+        run: docker compose -f ./cmd/example-mysql/docker/compose.yaml up --build --detach
+      - name: Run integration test
```
roger2hk, line 17: Call this `Run benchmark test`?

mhutchinson: #150

PR #144, opened 2024-08-15
Terraform for GCP CI base

AlCutter:
This PR adds support for spinning up an instance of the example personality on GCP.

This is intended to be used for CI tests.

Towards #7

### DIFF
--- a/deployment/live/gcp/example-gcp/ci/terragrunt.hcl
+++ b/deployment/live/gcp/example-gcp/ci/terragrunt.hcl
@@ -10,6 +10,6 @@ include "root" {
 inputs = merge(
   include.root.locals,
   {
-    example_gcp_docker_image = "todo"
+    example_gcp_docker_image = "us-central1-docker.pkg.dev/trillian-tessera/docker-prod/example-gcp:latest"
   }
 )
--- a/deployment/live/gcp/example-gcp/ci/terragrunt.hcl
+++ b/deployment/live/gcp/example-gcp/ci/terragrunt.hcl
@@ -11,5 +11,6 @@ inputs = merge(
   include.root.locals,
   {
     example_gcp_docker_image = "us-central1-docker.pkg.dev/trillian-tessera/docker-prod/example-gcp:latest"
+    log_origin = "example-gcp"
   }
 )
--- a/deployment/live/gcp/example-gcp/ci/terragrunt.hcl
+++ b/deployment/live/gcp/example-gcp/ci/terragrunt.hcl
@@ -1,0 +1,15 @@
+terraform {
+  source = "${get_repo_root()}/deployment/modules/gcp//example-gcp"
+}
+
+include "root" {
+  path   = find_in_parent_folders()
+  expose = true
+}
+
+inputs = merge(
+  include.root.locals,
+  {
+    example_gcp_docker_image = "todo"
+  }
+)
--- a/deployment/live/gcp/example-gcp/terragrunt.hcl
+++ b/deployment/live/gcp/example-gcp/terragrunt.hcl
@@ -6,7 +6,7 @@ locals {
   env         = path_relative_to_include()
   project_id = get_env("GOOGLE_PROJECT", "trillian-tessera")
   location   = get_env("GOOGLE_REGION", "us-central1")
-  base_name   = get_env("TESSERA_BASE_NAME", "${path_relative_to_include()}-example-gcp")
+  base_name   = get_env("TESSERA_BASE_NAME", "${local.env}-example-gcp")
 }
 
 remote_state {
@@ -16,7 +16,7 @@ remote_state {
     project  = local.project_id
     location = local.location
     bucket   = "${local.project_id}-${local.base_name}-terraform-state"
-    prefix   = "${path_relative_to_include()}/terraform.tfstate"
+    prefix   = "${local.env}/terraform.tfstate"
 
     gcs_bucket_labels = {
       name  = "terraform_state_storage"
--- a/deployment/live/gcp/example-gcp/terragrunt.hcl
+++ b/deployment/live/gcp/example-gcp/terragrunt.hcl
@@ -4,8 +4,8 @@ terraform {
 
 locals {
   env         = path_relative_to_include()
-  project_id = get_env("GOOGLE_PROJECT", "trillian-tessera")
-  location   = get_env("GOOGLE_REGION", "us-central1")
+  project_id  = get_env("GOOGLE_PROJECT", "trillian-tessera")
+  location    = get_env("GOOGLE_REGION", "us-central1")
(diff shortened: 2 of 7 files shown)


deployment/live/gcp/example-gcp/ci/terragrunt.hcl
```diff
@@ -0,0 +1,16 @@
+terraform {
+  source = "${get_repo_root()}/deployment/modules/gcp//example-gcp"
```
mhutchinson, line 2: This double slash is a bit weird?

deployment/live/gcp/example-gcp/terragrunt.hcl
```diff
@@ -1,25 +1,22 @@
...
-  source = "${get_repo_root()}/deployment/modules/gcp/gcs"
+  source = "${get_repo_root()}/deployment/modules/gcp//example-gcp"
 }
 
 locals {
+  env         = path_relative_to_include()
   project_id = get_env("GOOGLE_PROJECT", "trillian-tessera")
   location   = get_env("GOOGLE_REGION", "us-central1")
-  base_name   = get_env("TESSERA_BASE_NAME", "example-gcp")
+  base_name   = get_env("TESSERA_BASE_NAME", "${path_relative_to_include()}-example-gcp")
```
mhutchinson, line 9: Can you use $env here instead of the path calculation? It'd be clearer for a human reading it, even if the result is the same.

deployment/live/gcp/example-gcp/terragrunt.hcl
```diff
@@ -1,25 +1,22 @@
 terraform {
-  source = "${get_repo_root()}/deployment/modules/gcp/gcs"
+  source = "${get_repo_root()}/deployment/modules/gcp//example-gcp"
 }
 
 locals {
+  env         = path_relative_to_include()
   project_id = get_env("GOOGLE_PROJECT", "trillian-tessera")
```
mhutchinson, line 7: nit: alignment of `=` has got wonky.

deployment/live/gcp/example-gcp/terragrunt.hcl
```diff
@@ -1,25 +1,22 @@
...
-)
-
 remote_state {
   backend = "gcs"
 
   config = {
     project  = local.project_id
     location = local.location
     bucket   = "${local.project_id}-${local.base_name}-terraform-state"
+    prefix   = "${path_relative_to_include()}/terraform.tfstate"
```
mhutchinson, line 19: Again, using `${local.env}` would be clearer, IMO.

deployment/modules/gcp/example-gcp/main.tf
```diff
@@ -0,0 +1,134 @@
+terraform {
+  backend "gcs" {}
+}
+
+## Call the Tessera GCP module
+##
+## This will configure all the storage infrastructure required to run a Tessera log on GCP.
+module "gcs" {
+  source = "..//gcs"
```
mhutchinson, line 9: Is the double slash needed here? I think this is one of those cases where there is something needed about the 2 slashes but best to check.

deployment/modules/gcp/example-gcp/main.tf
```diff
@@ -0,0 +1,134 @@
...
+        tcp_socket {
+          port = 8080
+        }
+      }
+    }
+    containers {
+      image      = "us-docker.pkg.dev/cloud-ops-agents-artifacts/cloud-run-gmp-sidecar/cloud-run-gmp-sidecar:1.0.0"
+      name       = "collector"
+      depends_on = ["example-gcp"]
+    }
```
mhutchinson, line 122: This does the prometheus integration into GCP monitoring. Not sure if you need it in the example?

It could do with some comments put through here for things like this though, or people will just steal all of this whether it's needed or not. The curse of writing examples that try to strike a balance between "hello world" and showing actual real-life deployment.

deployment/modules/gcp/example-gcp/main.tf
```diff
@@ -0,0 +1,134 @@
...
+  template {
+    service_account = google_service_account.cloudrun_service_account.email
+    containers {
+      image = var.example_gcp_docker_image
+      name  = "example-gcp"
+      args = [
+        "--logtostderr",
+        "--v=1",
+        "--bucket=${module.gcs.log_bucket.id}",
+        "--spanner=projects/${var.project_id}/instances/${module.gcs.log_spanner_instance.name}/databases/${module.gcs.log_spanner_db.name}",
```
mhutchinson, line 98: nit: You could consider evaluation this monster as a local, and even exporting it as an output variable. It would make this line shorter and thus easier to eyeball the flags.

deployment/modules/gcp/example-gcp/main.tf
```diff
@@ -0,0 +1,134 @@
...
+  crypto_key = google_kms_crypto_key.log_signer.id
+}
+
+###
+### Set up Cloud Run service
+###
+resource "google_service_account" "cloudrun_service_account" {
+  account_id   = "cloudrun-${var.env}-sa"
+  display_name = "Service Account for Cloud Run (${var.env})"
+}
```
mhutchinson, line 56: This isn't strictly necessary as there's a default service account you could use. It's good practice though. We should probably write in some docs our approach to the example on how to balance best practice vs minimalism.

mhutchinson (approved):

deployment/live/gcp/example-gcp/ci/terragrunt.hcl
AlCutter: This is how you tell terraform about module "boundaries", with the double slash it brings in anything under there, without them it fails when you try to reference the sibling `gcs` from within the `example-gc` module.

This is a bit of a dusty corner in terraform, but it's why e.g. you had to put double slashes in for your distributor config which referenced other modules:https://github.com/transparency-dev/distributor/blob/main/deployment/modules/distributor/main.tf#L110

deployment/live/gcp/example-gcp/terragrunt.hcl
AlCutter: I assumed I couldn't, but rather unexpectedly it seems like you can (or at least `${local.env}` you can)...

deployment/live/gcp/example-gcp/terragrunt.hcl
AlCutter: done

deployment/modules/gcp/example-gcp/main.tf
AlCutter: yup, same as above.

deployment/modules/gcp/example-gcp/main.tf
AlCutter: Yeah, guidance is a good idea. I'll merge this as-is while we chew on that.

PR #146, opened 2024-08-16
First cut at CloudBuild config

AlCutter:
This PR adds a first cut of a Cloud Build config for the Tessera repo.

It's mostly a trimmed-down version of the same thing from the `distributor` repo.

This is primarily intended to provide support for doing CI/CD for the Tessera GCP backend via the `example-gcp` binary.

Towards #7.


deployment/live/cloudbuild/README.md
mhutchinson, line 1: WDYT about a subdirectory like `live/gcp/cloudbuild` ? At the moment this is disambiguated by different cloud providers having different naming, which seems like it could get hard to understand if any author/user doesn't have a good mental map of all these names?

This would also affect the `modules` directory. I imagine both `modules` and `live` would contain {gcp, aws, local} or something like that? The other option is to refactor later, but it's a bit messier/riskier because there are non statically-built links like docs and the terraform files that are only checked at runtime.

deployment/live/cloudbuild/README.md
```diff
@@ -0,0 +1,14 @@
+# Cloudbuild Triggers and Steps
+
+This directory contains a terragrunt file that can be deployed using `terragrunt apply` to define the necessary triggers and steps in GCB.
```
mhutchinson, line 3: Link to `terragrunt` installation docs?

deployment/modules/cloudbuild/main.tf
```diff
@@ -0,0 +1,97 @@
+terraform {
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = "5.14.0"
```
mhutchinson, line 5: Lastest version is 5.41. Would recommend using the latest version (if not now, in a follow-up PR). Usually there are no changes needed, but with this many releases, there could be...

https://registry.terraform.io/providers/hashicorp/google/latest

Relatedly, but orthogonally, we should look at dependabot for these config bumps. https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#terraform suggests the feature is there, but the TL;DR of https://github.com/dependabot/dependabot-core/issues/8725 is "use renovate".

AlCutter:
.

PR #151, opened 2024-08-19
Set content type appropriately

phbnf:
Need to initialize the writer before setting the content type

### DIFF
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -669,13 +669,13 @@ func (s *gcsStorage) setObject(ctx context.Context, objName string, data []byte,
 	obj := bkt.Object(objName)
 
 	var w *gcs.Writer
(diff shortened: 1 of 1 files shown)


storage/gcp/gcp.go
```diff
@@ -669,13 +669,13 @@ func (s *gcsStorage) setObject(ctx context.Context, objName string, data []byte,
...
 
 	var w *gcs.Writer
-	w.ObjectAttrs.ContentType = contType
 	if cond == nil {
 		w = obj.NewWriter(ctx)
 
 	} else {
 		w = obj.If(*cond).NewWriter(ctx)
 	}
+	w.ObjectAttrs.ContentType = contType
```
AlCutter, line 678: Ah, yes - that seems like a better location :)

AlCutter (approved):

PR #153, opened 2024-08-20
Add support for using GCP KMS in example-gcp

AlCutter:
This PR enables `example-gcp` to use GCP's KMS support for Ed25519 signatures.


mhutchinson (approved):
I guess you'll be updating the terraform to match. Docs (or ideally automated docs like terraform) for setting up this more complex example would be good.

AlCutter:
Terraform for this has been added to #144, just trying to make sure it all works more or less as intended before I send that over to you.

PR #160, opened 2024-08-21
README: direct users to the examples

mhutchinson:
We can come back and add example usage in this main README if that's useful, but for now pointing to the examples that are fleshed out and are known to compile is a pragmatic step forward. It also gives newcomers links to binaries they can run to understand what Tessera can do, before they start writing any code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Take a look at the example personalities in the `/cmd/` directory:
   - [example-gcp](./cmd/example-gcp/): example of operating a log running in GCP
     - This example can be deployed via terraform (see the [deployment](./deployment/) directory)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,17 +48,16 @@ migration to Tessera and adopting the patterns it encourages.
...
-```go
-
-import (
-    tessera "github.com/transparency-dev/trillian-tessera"
-)
-
-// TODO...
-
-```
+The `main.go` files for each of these example personalities tries to strike a balance when demonstrating features of Tessera between simplicity, and demonstrating best practices.
```
AlCutter, line 59: /tries/try/

AlCutter (approved):

PR #165, opened 2024-08-21
return Add error messages

phbnf:
#88

### DIFF
--- a/personalities/sctfe/handlers.go
+++ b/personalities/sctfe/handlers.go
@@ -324,7 +324,7 @@ func addChainInternal(ctx context.Context, li *logInfo, w http.ResponseWriter, r
 	klog.V(2).Infof("%s: %s => storage.Add", li.LogOrigin, method)
 	idx, err := li.storage.Add(ctx, entry)
 	if err != nil {
-		return http.StatusInternalServerError, fmt.Errorf("couldn't store the leaf")
+		return http.StatusInternalServerError, fmt.Errorf("couldn't store the leaf: %v", err)
 	}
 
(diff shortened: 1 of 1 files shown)


personalities/sctfe/handlers.go
```diff
@@ -324,7 +324,7 @@ func addChainInternal(ctx context.Context, li *logInfo, w http.ResponseWriter, r
 	klog.V(2).Infof("%s: %s => storage.Add", li.LogOrigin, method)
 	idx, err := li.storage.Add(ctx, entry)
 	if err != nil {
-		return http.StatusInternalServerError, fmt.Errorf("couldn't store the leaf")
+		return http.StatusInternalServerError, fmt.Errorf("couldn't store the leaf: %v", err)
```
AlCutter, line 327: Will this error get written out as a response body?
May want to think about how much internal info we want to print out there if so...

Perhaps worth an issue to come back and audit?

AlCutter (approved):

personalities/sctfe/handlers.go
phbnf: Yes - I added an AI on #88.

phbnf: Coming back to this one - I think this error will never been sent to clients, thanks to [this](https://github.com/phbnf/trillian-tessera/blob/d2a7cc7db04d9d8de3d42f9fb23de81942d8c1b0/personalities/sctfe/handlers.go#L262-L267). I still think it's worth doing an error overhall, but thought it would be nice to update this thread.

PR #167, opened 2024-08-22
MySQL: extract tlog http API config to function

mhutchinson:
This small refactoring clearly demonstrates the handlers that are configured for the core tlog spec. The method could be moved in the future, but I suggest leaving it here at first to make sure that any other implementation is exactly the same.

Towards #21.

### DIFF
--- a/cmd/example-mysql/main.go
+++ b/cmd/example-mysql/main.go
@@ -80,7 +80,7 @@ func main() {
 		klog.Exitf("Failed to create new MySQL storage: %v", err)
 	}
 
-	configureTilesReadApi(http.DefaultServeMux, storage)
+	configureTilesReadAPI(http.DefaultServeMux, storage)
 	http.HandleFunc("POST /add", func(w http.ResponseWriter, r *http.Request) {
 		b, err := io.ReadAll(r.Body)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/example-mysql/main.go
```diff
@@ -80,7 +80,41 @@ func main() {
...
+	if err := http.ListenAndServe(*listen, http.DefaultServeMux); err != nil {
+		klog.Exitf("ListenAndServe: %v", err)
+	}
+}
+
+// configureTilesReadApi adds the API methods from https://c2sp.org/tlog-tiles to the mux,
+// routing the requests to the mysql storage.
+// This method could be moved into the storage API as it's likely this will be
+// the same for any implementation of a personality based on MySQL.
+func configureTilesReadApi(mux *http.ServeMux, storage *mysql.Storage) {
```
roger2hk, line 116:
```suggestion
func configureTilesReadAPI(mux *http.ServeMux, storage *mysql.Storage) {
```

roger2hk (approved):

cmd/example-mysql/main.go
mhutchinson: Done.

PR #168, opened 2024-08-22
GCP CI Test

AlCutter:


roger2hk:
The magic modules seems to be a good fit here.

https://github.com/GoogleCloudPlatform/magic-modules

roger2hk:
> The magic modules seems to be a good fit here.
> 
> https://github.com/GoogleCloudPlatform/magic-modules

Ah, maybe not. It is a tool for developing the Terraform provider for GCP.

PR #169, opened 2024-08-22
Cache locally whether issuers have been persisted or not already.

phbnf:
This saves a few round trips to the IssuerStorage.

Towards #88


personalities/sctfe/storage.go
```diff
@@ -78,3 +87,65 @@ func (cts *CTStorage) AddIssuerChain(ctx context.Context, chain []*x509.Certific
...
 	return nil
 }
+
+// cachedIssuerStorage wraps an IssuerStorage, and keeps a copy the sha256 of certs it contains.
+//
+// This is intended to make querying faster. It does not keep a copy of the certs, only sha256.
+// Only up to N keys will be stored locally.
+// TODO(phboneff): add monitoring for the number of keys
+type cachedIssuerStorage struct {
+	m map[string]bool
```
roger2hk, line 97: `map` is not safe for concurrent use.

AlCutter: Would it make sense to use something like https://github.com/hashicorp/golang-lru ?
Might be overkill, OTOH it should degrade more gracefully if we do bump into the limits.

personalities/sctfe/storage.go
```diff
@@ -78,3 +87,65 @@ func (cts *CTStorage) AddIssuerChain(ctx context.Context, chain []*x509.Certific
...
+	N int // maximum number of entries allowed in m
+	s IssuerStorage
+}
+
+// Exists checks if the key exists in the local cache, if not checks in the underlying storage.
+// If it finds it there, caches the key locally.
+func (c cachedIssuerStorage) Exists(ctx context.Context, key []byte) (bool, error) {
+	_, ok := c.m[string(key)]
+	if ok {
```
AlCutter, line 106: the comment at the top suggests that map keys are 64byte - is `key` here already hex encoded?

If not, then might be worth updating the comment, if it is, then L107 below will re-hexencode it.

personalities/sctfe/storage.go
AlCutter, line 97: If you choose to continue using `map`, you can save a few bytes and a bit of ambiguity here/below by defining as `m map[string]struct{}` (the value takes no space like that), then below you mark existence with `m[key] = struct{}{}`

personalities/sctfe/storage.go
```diff
@@ -78,3 +87,65 @@ func (cts *CTStorage) AddIssuerChain(ctx context.Context, chain []*x509.Certific
...
+			return nil
+		}
+		c.m[string(kv.K)] = true
+	}
+	return nil
+}
+
+func NewCachedIssuerStorage(s IssuerStorage) cachedIssuerStorage {
+	c := cachedIssuerStorage{s: s, N: maxCachedIssuerKeys}
+	c.m = make(map[string]bool)
```
AlCutter, line 149: nit: can inline this into the instantiation of `cachedIssuerStorage{}` above, and even inline all _that_ into the `return` :)

personalities/sctfe/storage.go
phbnf: Hm - I did not page it would actually *fail* on concurrent writes, thanks! I had made the assumption that we'd only write the same value so concurrent writes wouldn't matter. I could use sync.Map, or a regular map with just a mutex on writes.

LRU: I did consider it and concluded it was not necessary / misleading.  Given the math above, it's extremely unlikely that we'll ever run into the limit, so I thought it was better to be explicit about it. Also, I don't necessarily think that a lru would improve things: what matters here is the read frequency, not the write order. I think that  https://github.com/hashicorp/golang-lru only updates the frequency when you Get an item, not when you check for its existence. Last, it's reasonable to assume that operators are meant to restart their binary every week. The goal of caching is to speed things up and not make pricey calls to GCS. If Add requests fill 1 << 20 entries of caching within a week.... I think we should just rethink it from the ground up, and probably store intermediates in the same way that we do deduplication.

personalities/sctfe/storage.go
```diff
@@ -94,33 +95,38 @@ func (cts *CTStorage) AddIssuerChain(ctx context.Context, chain []*x509.Certific
...
+	sync.RWMutex
+	m map[string]struct{}
 	N int // maximum number of entries allowed in m
 	s IssuerStorage
 }
 
 // Exists checks if the key exists in the local cache, if not checks in the underlying storage.
 // If it finds it there, caches the key locally.
-func (c cachedIssuerStorage) Exists(ctx context.Context, key []byte) (bool, error) {
+func (c *cachedIssuerStorage) Exists(ctx context.Context, key []byte) (bool, error) {
```
AlCutter, line 106: Is `Exists` meant to be used by higher level code?

The implementation of `Exists` and `AddIssuers` are not really all that different; I wonder if you could simply higher level code just call `AddIssuers` (`AddIssuersIfNotExist`?) for every submission, and modify that func to also check for the entry in `m` at the top of the loop, and only call the underlying storage Add() iff there is anything to store.

As it is, if a new issuer is seen it looks like you'll end up taking the lock and (re-)writing the same entry to `m` 2 or possibly 3 times:
- once in a higher level direct call to `Exists` (L119-121 if the issuer is new and is not in `m` or in the storage) (assuming that's supposed to happen)
- once via a call to `Exists` from `AddIssuers` once the higher level code sees that the issuer doesn't exist and wants to add it
- once inside `AddIssuers` at L119-121

AlCutter: Oke doke - can you pop a TODO in there?

personalities/sctfe/storage.go
```diff
@@ -104,9 +104,7 @@ type cachedIssuerStorage struct {
 // Exists checks if the key exists in the local cache, if not checks in the underlying storage.
 // If it finds it there, caches the key locally.
 func (c *cachedIssuerStorage) Exists(ctx context.Context, key []byte) (bool, error) {
-	c.RLock()
```
AlCutter: Aren't you back in `panic` territory now, though?
You'll need to `c.Lock()` here still, I think?

phbnf: Huh - I thought only two concurrent writes would fail.

personalities/sctfe/storage.go
phbnf: Sure thing, PR here: https://github.com/phbnf/trillian-tessera/compare/localcache...phbnf:trillian-tessera:inline?expand=1

personalities/sctfe/storage.go
AlCutter: No, unprotected `map` are not safe for anything other than concurrent reads; a concurrent read & write would be bad too.

personalities/sctfe/storage.go
```diff
@@ -104,7 +105,9 @@ type cachedIssuerStorage struct {
 // Exists checks if the key exists in the local cache, if not checks in the underlying storage.
 // If it finds it there, caches the key locally.
 func (c *cachedIssuerStorage) Exists(ctx context.Context, key []byte) (bool, error) {
+	c.RLock()
 	_, ok := c.m[string(key)]
+	c.RLock()
```
AlCutter, line 110: `RUnlock()`? :)

AlCutter (approved):

PR #178, opened 2024-08-28
Add support for Static CT API entry bundle paths

### LINKED ISSUE #171: Tessera needs to support CT tile/data path
`static-ct-api` entry bundles are stored under `tile/data`: https://github.com/C2SP/C2SP/blob/main/static-ct-api.md#log-entries. This is different from the tiles specs, where entry bundles are stored under `tile/entries`. Josie had a go at trying to align both https://github.com/C2SP/C2SP/issues/94, and it seems unlikely that they will converge.

I'm thinking of adding an option to tessera setting this path. Having said that, I can't think of a use case different than CT, so we might as not make this option accessible outside of CT. I can be explicit with names that this should on be used for CT, but maybe you can think of a better way of doing it?

AlCutter:
This PR adds support to Tessera for storing CT entry bundles in under `data`.

The [Static CT API](https://c2sp.org/static-ct-api) defines leaf entry bundles as being stored under a `tiles/data` path. This is incompatible with the [tlog-tiles spec](https://c2sp.org/tlog-tiles) which defines that path as `tiles/entries`.

This PR adds a CT-only storage option, `WithCTLayout`, which instructs storage implementations to use the Static CT API path.

A strong goal of this PR was to make it impossible for users of Tessera to implement arbitrary layout schemes, or accidentally use the CT layout without a clear signal that this is what they are doing. 

Fixes #171

### DIFF
--- a/api/layout/paths.go
+++ b/api/layout/paths.go
@@ -49,12 +49,12 @@ func NWithSuffix(l, n, logSize uint64) string {
 // EntriesPath returns the local path for the nth entry bundle. p denotes the partial
 // tile size, or 0 if the tile is complete.
 func EntriesPath(n, logSize uint64) string {
-	return fmt.Sprintf("tile/entries%s", NWithSuffix(0, n, logSize))
+	return fmt.Sprintf("tile/entries/%s", NWithSuffix(0, n, logSize))
 }
 
 // TilePath builds the path to the subtree tile with the given level and index in tile space.
 func TilePath(tileLevel, tileIndex, logSize uint64) string {
-	return fmt.Sprintf("tile/%d%s", tileLevel, NWithSuffix(tileLevel, tileIndex, logSize))
+	return fmt.Sprintf("tile/%d/%s", tileLevel, NWithSuffix(tileLevel, tileIndex, logSize))
 }
 
 // fmtN returns the "N" part of a Tiles-spec path.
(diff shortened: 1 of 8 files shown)


ct_only.go
```diff
@@ -56,3 +58,12 @@ func convertCTEntry(e *ctonly.Entry) *Entry {
 
 	return r
 }
+
+// WithCTLayout instructs the underlying storage to use a Static CT API compatible scheme for layout.
+func WithCTLayout() func(*StorageOptions) {
+	return func(opts *StorageOptions) {
+		opts.EntriesPath = func(n, logSize uint64) string {
+			return fmt.Sprintf("tile/entries%s", layout.NWithSuffix(0, n, logSize))
```
mhutchinson, line 66:
```suggestion
			return fmt.Sprintf("tile/data%s", layout.NWithSuffix(0, n, logSize))
```

??

api/layout/paths.go
```diff
@@ -37,25 +37,24 @@ func EntriesPathForLogIndex(seq, logSize uint64) string {
 	return EntriesPath(seq/256, logSize)
 }
 
-// EntriesPath returns the local path for the nth entry bundle. p denotes the partial
-// tile size, or 0 if the tile is complete.
-func EntriesPath(n, logSize uint64) string {
+// NWithSuffix returns a tiles-spec "N" path, with a partial suffix if applicable.
+func NWithSuffix(l, n, logSize uint64) string {
```
phbnf, line 41: Is there any logic to what function returns with a leading slash, and which one don't?
For instance, "EntriesPath" doesn't start with a leading one, but "fmtN" does. It's slightly orthogonal to this change, but it would be good to make things consistent / document them / use path.Join if it makes sense

phbnf (approved):

mhutchinson (approved):

PR #180, opened 2024-08-28
Remove default options from ResolveStorageOptions

AlCutter:
This PR removes the default options param from `ResolveStorageOptions`.

This is an unnecessary source of complexity at the moment.

### DIFF
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -22,7 +22,6 @@ import (
 	"errors"
 	"fmt"
 	"strings"
-	"time"
 
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/transparency-dev/merkle/rfc6962"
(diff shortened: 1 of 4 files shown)


storage/mysql/mysql.go
```diff
@@ -59,10 +59,7 @@ type Storage struct {
 // New creates a new instance of the MySQL-based Storage.
 // Note that `tessera.WithCheckpointSignerVerifier()` is mandatory in the `opts` argument.
 func New(ctx context.Context, db *sql.DB, opts ...func(*tessera.StorageOptions)) (*Storage, error) {
-	opt := tessera.ResolveStorageOptions(&tessera.StorageOptions{
-		BatchMaxAge:  defaultQueueMaxAge,
-		BatchMaxSize: defaultBatchMaxSize,
-	}, opts...)
+	opt := tessera.ResolveStorageOptions(opts...)
```
mhutchinson, line 62: This changes the behaviour of mysql to switch to a batch size of 1 instead of 256?

AlCutter: It does, but that brings it into line with the others, and I've just changed the "line" to `256` within `250ms`.

We can revisit in the future.

roger2hk: The `defaultBatchMaxSize` and `defaultQueueMaxAge` consts in this file can be removed.

mhutchinson (approved):

PR #181, opened 2024-08-28
example-posix brings up an HTTP POST /add handler

mhutchinson:
This makes this example consistent in API with the other examples, and allows the implementation to be load tested via the hammer

PR #182, opened 2024-08-28
Surface futures from Add()

AlCutter:
This PR returns the futures already present within calls to `Add()`, rather than blocking the caller directly. This allows the user of Tessera to decide what is best for them.

For HTTP-based services, a blocking call to `Add` isn't really an issue - there's generally a goroutine per request anyway, but for other use-cases having to artificially spin out a bunch of goroutines which call `Add` only to then need to join them again immediately afterwards is a bit weird. This PR solves that by allowing the caller of the `Add` method to choose _when_ to resolve the future.

The POSIX example/one-shot is a good example of the complexity this pushes onto users. Exposing the future removes a lot of that, although adds some (but arguably less) complexity of its own.

### DIFF
--- a/log.go
+++ b/log.go
@@ -45,6 +45,9 @@ type ParseCPFunc func(raw []byte) (*f_log.Checkpoint, error)
 // EntriesPathFunc is the signature of a function which knows how to format entry bundle paths.
 type EntriesPathFunc func(n, logSize uint64) string
 
+// IndexFuture is the signature of a function which can return an assigned index or error.
+type IndexFuture func() (uint64, error)
+
 // StorageOptions holds optional settings for all storage implementations.
 type StorageOptions struct {
 	NewCP   NewCPFunc
--- a/log.go
+++ b/log.go
@@ -46,6 +46,9 @@ type ParseCPFunc func(raw []byte) (*f_log.Checkpoint, error)
 type EntriesPathFunc func(n, logSize uint64) string
 
 // IndexFuture is the signature of a function which can return an assigned index or error.
+//
+// Implementations of this func are likely to be "futures", or a promise to return this data at
+// some point in the future, and as such will block when called if the data isn't yet available.
 type IndexFuture func() (uint64, error)
 
 // StorageOptions holds optional settings for all storage implementations.
--- a/cmd/example-gcp/main.go
+++ b/cmd/example-gcp/main.go
@@ -83,7 +83,7 @@ func main() {
 		defer r.Body.Close()
 
 		id := sha256.Sum256(b)
-		idx, err := storage.Add(r.Context(), tessera.NewEntry(b, tessera.WithIdentity(id[:])))
+		idx, err := storage.Add(r.Context(), tessera.NewEntry(b, tessera.WithIdentity(id[:])))()
 		if err != nil {
 			if errors.Is(err, tessera.ErrPushback) {
 				w.Header().Add("Retry-After", "1")
--- a/cmd/example-mysql/main.go
+++ b/cmd/example-mysql/main.go
@@ -92,7 +92,7 @@ func main() {
 				klog.Warningf("/add: %v", err)
 			}
 		}()
-		idx, err := storage.Add(r.Context(), tessera.NewEntry(b))
+		idx, err := storage.Add(r.Context(), tessera.NewEntry(b))()
 		if err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
 			_, _ = w.Write([]byte(err.Error()))
--- a/cmd/example-posix/main.go
+++ b/cmd/example-posix/main.go
@@ -142,7 +142,7 @@ func main() {
 				klog.Warningf("/add: %v", err)
 			}
 		}()
-		idx, err := storage.Add(r.Context(), tessera.NewEntry(b))
+		idx, err := storage.Add(r.Context(), tessera.NewEntry(b))()
 		if err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
 			_, _ = w.Write([]byte(err.Error()))
--- a/cmd/posix-oneshot/main.go
+++ b/cmd/posix-oneshot/main.go
@@ -25,7 +25,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"sync"
 	"time"
 
 	"golang.org/x/mod/sumdb/note"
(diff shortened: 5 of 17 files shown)


log.go
```diff
@@ -41,6 +41,9 @@ type NewCPFunc func(size uint64, hash []byte) ([]byte, error)
 // ParseCPFunc is the signature of a function which knows how to verify and parse checkpoints.
 type ParseCPFunc func(raw []byte) (*f_log.Checkpoint, error)
 
+// IndexFuture is the signature of a function which can return an assigned index or error.
```
mhutchinson, line 44: Obvious to people that know about futures, but worth adding a comment to mention that this will block?

mhutchinson (commented):
LGTM, though I'm keen we get #181 in first or I'll have to rewrite most of that PR again due to the merge. You'll have to rebase on that, but it _should_ be as simple as reverting your changes to `example-posix` and then simply adding `()` after the call to `Add`.

log.go
AlCutter: Yeah, good call - done.

mhutchinson (approved):

phbnf:
I worry that this might push personalities towards returning to clients, without returning the index, which is a step away from my understanding of Tessera's philosophy to steer folks in the right direction, and a step closer to SCTs.
It seems to me that the problem this is solving for POSIX, is that it's requesting to sequence multiple leaves at a time. Should Tessera simply support this? What's the tradeoff here? Do you have other use cases in mind?

AlCutter:
> I worry that this might push personalities towards returning to clients, without returning the index, which is a step away from my understanding of Tessera's philosophy to steer folks in the right direction, and a step closer to SCTs. It seems to me that the problem this is solving for POSIX, is that it's requesting to sequence multiple leaves at a time. Should Tessera simply support this? What's the tradeoff here? Do you have other use cases in mind?

I think that particular concern can be addressed with docs/comments and examples, but do bear in mind that that the _only_ way to check whether there's been an error is also to call the future so I'd be a little surprised if folks just entirely ignored it.

As an aside, supporting adding multiple entries has a bunch of interesting implications for API design and storage contracts, e.g. should the user/storage try to add all entries or none if there's a failure? If it should, then we've just made life very difficult for non-transactional storage implementations and undermined the queuing we currently have. If not, then we also need to figure out some mechanism for returning individual indices and errs for all the entries.

phbnf:
I think that all your points are valid - and they will persist with futures, but will be carried over to personalities, hence my hesitations. How should a personality react if it receives a `tessera.ErrPushback` error? What if request A and B gets in, the future for A returns a `tessera.ErrPushback`, but the future for B goes through?

If futures add risk and complexity to all personalities, except POSIX then I'd say that we need a very good POSIX use case to push the balance towards it, and this is what I'm trying to understand. What's the problem with POSIX reading all entries with a glob statement, and calling Add() sequentially for them, after the previous one has returned? Decreased likelihood that we sequence them simultaneously in the same bundle? Are we aware of any high throughput use case for POSIX today?

mhutchinson:
This is still pre-alpha so I think we should adopt and experiment with this API now. If we learned anything from Trillian v1 it's that hiding the internals from the API was a mistake. In v1 the thing we hid was tiles. Here, we're in danger of hiding an important state transition:

1. The entry exists only in the personality
2. The entry is acknowledged by Tessera, but not durably assigned
3. The entry is durably assigned
4. The entry is integrated

State 2 here is an important state transition, similar to "positive exchange of flight controls" used by pilots. Exposing a future gives access to state 2, where only returning the index hides it and jumps straight to state 3.

roger2hk:
Two thoughts on my mind:

1. What if we provide both sync and async add methods?

2. Currently there is no way for personality to know whether the entry is integrated or not. Can we add a callback function?

mhutchinson:
> 1. What if we provide both sync and async add methods?

What would the sync version do? Wait until state 3 or state 4 in the description in my last comment?

> 2. Currently there is no way for personality to know whether the entry is integrated or not. Can we add a callback function?

This is #193

AlCutter:
> Two thoughts on my mind:
> 
> 1. What if we provide both sync and async add methods?

If it were complex to convert between them that might be a good idea, but the sync implementation would be a single line so I don't think it's worthwhile:

```
func (s Storage) SyncAdd(ctx context.Context, e *Entry) (uint64, error) {
  return s.AsyncAdd(ctx, e)()
 }
```

> 2. Currently there is no way for personality to know whether the entry is integrated or not. Can we add a callback function?

This is relatively straightforward: wait for the checkpoint to have a size larger than the assigned index, but #193 tracks creating a "mixin" for doing that in the personality support packages.

phbnf:
I think it would make sense for both these APIs to look the same unless there is a specific reason that they don't: either they both return a future, or they both return an index.

mhutchinson:
> I think it would make sense for both these APIs to look the same unless there is a specific reason that they don't: either they both return a future, or they both return an index.

Both of which APIs?

phbnf:
Tessera's Add() API, and the API offered by the mixin Al is referring to.

roger2hk:
> > 1. What if we provide both sync and async add methods?
> 
> What would the sync version do? Wait until state 3 or state 4 in the description in my last comment?
> 
> > 2. Currently there is no way for personality to know whether the entry is integrated or not. Can we add a callback function?
> 
> This is #193

The sync version is exactly the one mentioned by @AlCutter. It's indeed a one line wrapper. If we don't provide this method, can we demostrate that in the method doc comment or the unit test?

roger2hk:
> > Two thoughts on my mind:
> > 
> > 1. What if we provide both sync and async add methods?
> 
> If it were complex to convert between them that might be a good idea, but the sync implementation would be a single line so I don't think it's worthwhile:
> 
> ```
> func (s Storage) SyncAdd(ctx context.Context, e *Entry) (uint64, error) {
>   return s.AsyncAdd(ctx, e)()
>  }
> ```
> 
> > 2. Currently there is no way for personality to know whether the entry is integrated or not. Can we add a callback function?
> 
> This is relatively straightforward: wait for the checkpoint to have a size larger than the assigned index, but #193 tracks creating a "mixin" for doing that in the personality support packages.

Waiting for the checkpoint to have a size larger than the assigned index is a pull mechanism. With the callback function, that would be a push mechanism. However, there is no personality that would require sending callback URL to the entry submitter (although this is very common in payment system).

roger2hk (approved):

PR #192, opened 2024-09-02
Flesh out design readme a bit

AlCutter:
This PR adds a bit more detail to `docs/design/README.md`.

Towards #4 & #9

### DIFF
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -2,15 +2,124 @@
 
 This directory contains design documentation for Tessera.
 
-It's probably wise to start with the [philosophy](philosophy.md]) doc first in order to provide
+It's probably wise to start with the [philosophy](philosophy.md) doc first in order to establish
 the context around the approach and design trade-offs made herein.
+
+## Objective
+
+Tessera is a library for managing Merkle trees & building transparency applications.
+It is designed to embody and guide users towards the current best practices developed over time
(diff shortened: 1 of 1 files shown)


docs/design/README.md
```diff
@@ -2,15 +2,124 @@
...
+1. Take a simpler MVP approach compared to Trillian v1.
+1. Correctly, reliably, and cheaply manage Merkle trees.
+1. Enable users of the library to build safe transparency systems, conforming to the [tlog-tiles][] spec
+   (or the proposed new version of CT described by the [static-ct-api][] spec).
+   + Tessera means _"An individual tile, used in creating a mosaic"_!)
+1. Reduce TCO of operating logs compared to Trillian v1.
+
+### Non-goals:
+
+1. Multi-tenancy.
```
patflynn, line 27: I'm kind of wondering whether this call out makes sense, since multi-tenancy is not impossible at the personality level.  Perhaps the key is that with a move from service -> library the tenancy question goes out of scope?

patflynn (approved):

docs/design/README.md
AlCutter: Yeah, I was trying to say "Tessera (library) won't handle multi-tenancy (for you)", mainly just since it's a difference from Trillian v1.

Can take it out if it's not helpful though?

PR #199, opened 2024-09-03
GCP implementation of dedup storage

AlCutter:
This PR adds a first cut of a GCP Spanner based dedup storage for SCTFE.

TODO: needs #188


AlCutter:
(Rebased onto `main` now that #188 is in.)

personalities/sctfe/storage/gcp/dedup.go
```diff
@@ -0,0 +1,93 @@
...
+		return idx, true, nil
+	}
+}
+
+// Add stores associations between the passed-in identities and their indices.
+func (d *DedupStorage) Add(ctx context.Context, entries []dedup.LeafIdx) error {
+	m := make([]*spanner.MutationGroup, 0, len(entries))
+	for _, e := range entries {
+		m = append(m, &spanner.MutationGroup{
+			Mutations: []*spanner.Mutation{spanner.Insert("IDSeq", []string{"id", "h", "idx"}, []interface{}{0, e.LeafID, int64(e.Idx)})},
```
phbnf, line 81: Should this check whether it's overriding a smaller index?
I didn't specify this on the interface, but I did this for bbolt. I don't think it's a strong requirement, but if it's not too expensive (it might be?), might be worth having it? How about we put it for consistency (between implementation, and knowing that deduplication will always converge on the smaller index), and remove it later if it hurts performance?

phbnf: There could be two frontends that receive the same entry. Frontend A would get idx 1, the request on frontend B would get idx 2. But frontend B could put it in the deduplication database before frontend A, so the deduplication index will remain 2 forever.
It's not the end of the world, and if it hurts performance, I don't think it would be worth it, but I got from our recent conversation that it would be nice for the log to converge on the smallest possible index for a given leaf.

AlCutter: Hmm, yeah - ok, how about if we merge this as is for now, and I'll send another PR with the "update if lower" algo? That way we've got a baseline + patch for ease of diff/rollback/etc.

phbnf (approved):

PR #202, opened 2024-09-04
POSIX: inline log initialization

mhutchinson:
Both personalities had duplicate code for creating directories and writing out an empty checkpoint. This logic is going to be the same for any POSIX log usage, so it should be part of the log implementation.

For now, this is behind an extra parameter that is passed to New. There may be better ways to set this up, but I tried a few and this was the best. This is a pragmatic step forward and better than what we had, so let's get this in and refactor further from here if needed.

Also cleaned up the `New` method by moving an inline function that was assigned to a field is instead defined as a method.

### DIFF
--- a/cmd/posix-oneshot/main.go
+++ b/cmd/posix-oneshot/main.go
@@ -29,16 +29,9 @@ import (
 
 	"golang.org/x/mod/sumdb/note"
 
-	"github.com/transparency-dev/merkle/rfc6962"
 	tessera "github.com/transparency-dev/trillian-tessera"
 	"github.com/transparency-dev/trillian-tessera/storage/posix"
 	"k8s.io/klog/v2"
-
-	fmtlog "github.com/transparency-dev/formats/log"
-)
-
-const (
-	dirPerm = 0o755
 )
 
 var (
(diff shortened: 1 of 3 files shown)


cmd/posix-oneshot/main.go
```diff
@@ -57,33 +50,23 @@ func main() {
...
-			klog.Exitf("Failed to sign empty checkpoint: %s", err)
-		}
-		if err := posix.WriteCheckpoint(*storageDir, n); err != nil {
-			klog.Exitf("Failed to write empty checkpoint: %s", err)
+	if len(*entries) == 0 {
+		if *initialise {
+			_, err := posix.New(ctx, *storageDir, *initialise, tessera.WithCheckpointSignerVerifier(s, v))
+			if err != nil {
+				klog.Exitf("Failed to initialise storage: %v", err)
+			}
```
AlCutter, line 59: Could this be simplified by instantiating the storage once (using the `*initialise` so it does the right thing), but then exiting immediately afterwards if `len(*entries) == 0`, or am I missing something?

I notice that the instantiations differ in using/not using the `WithBatching`, is there a side-effect there?

storage/posix/files.go
```diff
@@ -362,6 +364,33 @@ func (s *Storage) StoreTile(_ context.Context, level, index, logSize uint64, til
...
 
+// initialise ensures that the storage location is valid by loading the checkpoint from this location.
+// If `create` is set to true, then this will first ensure that the directory path is created, and
+// an empty checkpoint is created in this directory.
+func (s *Storage) initialise(create bool) error {
+	if create {
+		// Create the directory structure and write out an empty checkpoint
+		klog.Infof("Initializing directory for POSIX log at %q (this should only happen ONCE per log!)", s.path)
+		if err := os.MkdirAll(s.path, dirPerm); err != nil {
+			return fmt.Errorf("Failed to create log directory: %q", err)
```
AlCutter, line 375: nit: /Failed/failed

cmd/posix-oneshot/main.go
mhutchinson: The `WithBatching` uses the length of the `filesToAdd`, which requires evaluating `*entries`. If this isn't provided then `readEntriesOrDie` dies. I could make it so that function instead returns an error, and then we'd have to set the batch size to 0, and then remember to exit after init.... urgh. It's fewer lines of code that other way, but I think harder to follow.

storage/posix/files.go
mhutchinson: Done.

cmd/posix-oneshot/main.go
AlCutter: Ahh, right, yeah, that makes sense.

Separately, is it worth breaking the init lines out into a func? (coming from making the main func easy to grok for the common path)

AlCutter (approved):

cmd/posix-oneshot/main.go
mhutchinson: Merging this as is. We can revisit this if you feel strongly about it.

PR #204, opened 2024-09-04
Documentation for POSIX and one-shot command

mhutchinson:
This documentation makes it clear when integration is performed in this implementation, and gives a clear motivation in the one-shot tooling for iterating over all of the futures to resolve them.

### DIFF
--- a/cmd/posix-oneshot/main.go
+++ b/cmd/posix-oneshot/main.go
@@ -72,7 +72,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	// Evaluate the glob provided by the flag in to determine the files containing leaves
+	// Evaluate the glob provided by the --entries flag to determine the files containing leaves
 	filesToAdd := readEntriesOrDie()
 
 	// Construct a new Tessera POSIX log storage, anchored at the correct directory, and initialising it if requested.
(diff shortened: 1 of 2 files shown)


cmd/posix-oneshot/main.go
```diff
@@ -62,41 +72,41 @@ func main() {
 		os.Exit(0)
 	}
 
+	// Evaluate the glob provided by the flag in to determine the files containing leaves
```
AlCutter, line 75:
```suggestion
	// Evaluate the glob provided by the --entries flag to determine the files containing leaves
```

storage/posix/files.go
```diff
@@ -128,8 +128,19 @@ func (s *Storage) unlockCP() error {
 	return nil
 }
 
-// Add commits to sequence numbers for an entry
-// Returns the sequence number assigned to the first entry in the batch, or an error.
+// Add takes an entry and queues it for inclusion in the log.
+// Upon storing the entry in a queue to be sequenced, it returns a future that will
+// evaluate to either the sequence number assigned to this entry, or an error.
+// This future is made available when the entry is queued, but not sequenced.
```
AlCutter, line 134: The last sentence of this para is correct, but might be easy to misinterpret. At the risk of a little repeating, how about "The future is returned/made available when the entry is queued, but not fulfilled until the entry is durably sequenced"?

cmd/posix-oneshot/main.go
mhutchinson: Done.

storage/posix/files.go
mhutchinson: Reworked this. PTAL?

AlCutter: Bonza

PR #206, opened 2024-09-04
Propagate fatal errors in integration tests

mhutchinson:
Previously this was limping on because calling t.Fatal in a goroutine doesn't kill the whole test. This change propagates the error so that it poisons the error group, causing the test to exit properly. This means that nil checkpoints can't live on inside the slice of collected checkpoints.


roger2hk (approved):
Good catch!

PR #209, opened 2024-09-05
HTTP handlers: skip closing the body

mhutchinson:
From Request.Body docs: The Server will close the request body. The ServeHTTP Handler does not need to.

### DIFF
--- a/cmd/conformance/gcp/main.go
+++ b/cmd/conformance/gcp/main.go
@@ -79,7 +79,6 @@ func main() {
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
-		defer r.Body.Close()
 
 		id := sha256.Sum256(b)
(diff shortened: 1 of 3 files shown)


roger2hk:
This is true when the body is fully read. Does the body need to be closed when the `io.ReadAll` returns an error?

mhutchinson:
The contract doesn't say anything about that: https://pkg.go.dev/net/http#Request (you need to scroll down a bit to get to Body, and there's no way to link to the field members for some reason)

mhutchinson:
Note that the contract is different on the http client request. That body does need to be closed. Carefully.

AlCutter (approved):

roger2hk:
> Note that the contract is different on the http client request. That body does need to be closed. Carefully.

Found it, thanks!

image

PR #215, opened 2024-09-05
POSIX conformance: run an HTTP file server

mhutchinson:
This simplifies a lot of other code that was spinning up busybox as an http server. This has failed a number of times in CI. The simpler deployment of this approach has a lot going for it. The one downside is that it shows non-canonical usage of the POSIX log. But as there's already an HTTP server running, it seems wasteful to not take advantage in this way.


AlCutter (approved):

roger2hk:
Ah, I was about to make the integration test to support file scheme.

mhutchinson:
> Ah, I was about to make the integration test to support file scheme.

I don't think that's needed now, but if you wanted to do that work then reading directly from the FS should still work.

PR #221, opened 2024-09-05
[POSIX] Refactor and use createExclusive

AlCutter:
This PR fixes an issue where files created using `createExclusive` have `0o600` perms.

This was caused by the use of `os.CreateTemp`, so we now just create a temporary file manually and `Rename` it into the target location. 

#22

### DIFF
--- a/storage/posix/files.go
+++ b/storage/posix/files.go
@@ -416,7 +416,7 @@ func readCheckpoint(path string) ([]byte, error) {
 func createExclusive(f string, d []byte) error {
 	tmpName := f + ".temp"
 	if err := os.WriteFile(tmpName, d, filePerm); err != nil {
-		return fmt.Errorf("unable to write leafdata to temporary file: %w", err)
+		return fmt.Errorf("unable to write data to temporary file: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


storage/posix/files.go
```diff
@@ -422,21 +414,10 @@ func readCheckpoint(path string) ([]byte, error) {
...
-	tmpFile, err := os.CreateTemp(filepath.Dir(f), "")
-	if err != nil {
-		return fmt.Errorf("unable to create temporary file: %w", err)
-	}
-	tmpName := tmpFile.Name()
-	n, err := tmpFile.Write(d)
-	if err != nil {
+	tmpName := f + ".temp"
+	if err := os.WriteFile(tmpName, d, filePerm); err != nil {
 		return fmt.Errorf("unable to write leafdata to temporary file: %w", err)
```
roger2hk, line 419:
```suggestion
		return fmt.Errorf("unable to write data to temporary file: %w", err)
```

roger2hk: `createExclusive` is used in both checkpoint and tile.

roger2hk (approved):

roger2hk (approved):

PR #223, opened 2024-09-05
Enable log data removal during POSIX docker compose down

roger2hk:
#8


roger2hk:
Closing this pull request as there is no straightforward solution to create, share and remove volume during docker compose up and down.

PR #224, opened 2024-09-05
Add storage performance doc

roger2hk:


docs/performance.md
```diff
@@ -0,0 +1,254 @@
+# Tessera Storage Performance
+
+This document describes the performance of each Trillian Tessera storage implementation under different conditions.
+
+## MySQL
+
+### GCP Free Tier VM Instance
+
+**e2-micro**
```
AlCutter, line 9: I think we should put caveats in here since the `e2-micro/small/medium` machines are all  "shared core", and don't have local storage for MySQL etc. - I suspect these numbers will vary wildly depending on local "weather".

It's great that Tessera _can_ run in the free tier on these machines, and it's worth calling that out, but we should also be clear about what folks would be getting if they do so.

docs/performance.md
```diff
@@ -0,0 +1,254 @@
...
+┌──────────────────────────────────────────────────────────────────────┐
+│Read (8 workers): Current max: 0/s. Oversupply in last second: 0      │
+│Write (256 workers): Current max: 409/s. Oversupply in last second: 0 │
+│TreeSize: 240921 (Δ 307qps over 30s)                                  │
+│Time-in-queue: 86ms/566ms/2172ms (min/avg/max)                        │
+│Observed-time-to-integrate: 516ms/1056ms/2531ms (min/avg/max)         │
+└──────────────────────────────────────────────────────────────────────┘
+```
+
+The bottleneck is at the dockerized MySQL instance, which consumes around 50% of the memory.
```
AlCutter, line 27: Could you connect the dots for me as to why 50% memory usage is the bottleneck? I suspect you know something I don't :)

docs/performance.md
```diff
@@ -0,0 +1,254 @@
...
+┌───────────────────────────────────────────────────────────────────────┐
+│Read (8 workers): Current max: 0/s. Oversupply in last second: 0       │
+│Write (512 workers): Current max: 2571/s. Oversupply in last second: 0 │
+│TreeSize: 2530480 (Δ 2047qps over 30s)                                 │
+│Time-in-queue: 41ms/120ms/288ms (min/avg/max)                          │
+│Observed-time-to-integrate: 568ms/636ms/782ms (min/avg/max)            │
+└───────────────────────────────────────────────────────────────────────┘
+```
+
+The bottleneck comes from the `cmd/conformance/mysql` binary on the free tier VM instance. The MySQL CPU usage is lower than 10%.
```
AlCutter, line 115: Isn't this one using CloudSQL?

docs/performance.md
roger2hk: If you scroll down a bit, you will see `mysqld` consumed 49.7% of the memory with 74.5 MiB free in total. The CPU usage is low. The swap memory is low as well, so the machine can barely work...

roger2hk: That's why I did another round of test with the same VM instance and Cloud SQL. The numbers improved a lot.

docs/performance.md
roger2hk: Correct, this is the Cloud SQL CPU usage.

roger2hk: Oops.. I forgot to capture the local machine CPU usage.

docs/performance.md
AlCutter: Could the bottleneck be very restricted IO limiting MySQL's performance?

roger2hk: It could be as IO increases when memory is running low. The default MySQL docker image is used without any performance tuning.

docs/performance.md
roger2hk: Added a note about the potential performance impact using vCPUs shared core.

AlCutter (approved):

docs/performance.md
roger2hk: Mentiond Tessera can run on GCP free tier.

docs/performance.md
roger2hk: I will add the missing data in the upcoming pull request.

Issue #200, opened 2024-09-04
Replace SCTFE multi-config with flags for a single log


phbnf:
Towards #88 

Today, the SCTFE is configured via a config proto file, and supports multiple logs. We've decided to drop the support for multiple logs, and move to flags.

I'll do this in multiple steps along these lines:
 1. drop multi log support
 2. move from config to flags where we can, keep some things in there if need be, especially for keys that will undergo further changes
 3. on GCP, use KMS and/or Tink
 4. simplify the internals of the SCTFE to remove the virtual configs structs, if we can

patflynn:
makes a ton of sense to minimize the surface for the MVP.

phbnf:
The config is now gone. A bit more work is required here to simplify the internals, and pass keys in a better way.

phbnf:
Closing this issue as we're migrating SCTFE code to transparency-dev/static-ct.

Here are issues for the two outstanding items:
 - https://github.com/transparency-dev/static-ct/issues/5
 - https://github.com/transparency-dev/static-ct/issues/6

PR #225, opened 2024-09-06
remove root_pem_file from cfg, and only allow a single file

phbnf:
Towards #200 and #88.

This PR does two things:
  - Requires for all roots to be defined in a single file, rather than multiple
  - Moves roots_pem_file from the config, to a flag

Leaving a TODO to move the roots_pem_file check rather than deleting in this PR because I've already done it further down in the chain of commits, but rebasing these commits is a bit of a pain with all the tests.


personalities/sctfe/config.go
```diff
@@ -48,9 +48,9 @@ type LogConfig struct {
 	// origin identifies the log. It will be used in its checkpoint, and
 	// is also its submission prefix, as per https://c2sp.org/static-ct-api
 	Origin string
-	// Paths to the files containing root certificates that are acceptable to the
+	// Paths to the file containing root certificates that are acceptable to the
```
AlCutter, line 51:
```suggestion
	// Path to the file containing root certificates that are acceptable to the
```

personalities/sctfe/ct_server_gcp/main.go
```diff
@@ -61,12 +61,12 @@ var (
...
-	origin    = flag.String("origin", "", "origin of the log, for checkpoints and the monitoring prefix")
-	projectID = flag.String("project_id", "", "origin of the log, for checkpoints and the monitoring prefix")
-	bucket    = flag.String("bucket", "", "name of the bucket to store the log in")
-	spannerDB = flag.String("spanner_db_path", "", "projects/{projectId}/instances/{instanceId}/databases/{databaseId}")
+	dedupPath          = flag.String("dedup_path", "", "Path to the deduplication database")
+	origin             = flag.String("origin", "", "origin of the log, for checkpoints and the monitoring prefix")
+	projectID          = flag.String("project_id", "", "origin of the log, for checkpoints and the monitoring prefix")
+	bucket             = flag.String("bucket", "", "name of the bucket to store the log in")
+	spannerDB          = flag.String("spanner_db_path", "", "projects/{projectId}/instances/{instanceId}/databases/{databaseId}")
+	rootsPemFile       = flag.String("roots_pem_file", "", "Paths to the file containing root certificates that are acceptable to the log. The certs are served through get-roots endpoint.")
```
AlCutter, line 69:
```suggestion
	rootsPemFile       = flag.String("roots_pem_file", "", "Path to the file containing root certificates that are acceptable to the log. The certs are served through get-roots endpoint.")
```

AlCutter (approved):

PR #226, opened 2024-09-06
Move more settings from the config to flags

phbnf:
Towards #200 and #88

This PR moves rejectExpired, rejectUnexpired, extKeyUsages  and rejectExtensions  from the config to flags.

I considered parsing the extKeyUsages and rejectExtensions with custom flag types in main.go, or even with multiflags... but one step at a time, let's err on simplicity for now.


personalities/sctfe/config.go
```diff
@@ -122,7 +122,7 @@ func LogConfigFromFile(filename string) (*configpb.LogConfig, error) {
 //   - Merge delays (if present) are correct.
 //
 // Returns the validated structures (useful to avoid double validation).
-func ValidateLogConfig(cfg *configpb.LogConfig, origin string, projectID string, bucket string, spannerDB string, rootsPemFile string) (*ValidatedLogConfig, error) {
+func ValidateLogConfig(cfg *configpb.LogConfig, origin string, projectID string, bucket string, spannerDB string, rootsPemFile string, rejectExpired bool, rejectUnexpired bool, extKeyUsages string, rejectExtensions string) (*ValidatedLogConfig, error) {
```
AlCutter, line 125: Is the plan eventually to have a non-proto config struct which gets passed in here with all the required info inside?

personalities/sctfe/config.go
```diff
@@ -140,20 +140,28 @@ func ValidateLogConfig(cfg *configpb.LogConfig, origin string, projectID string,
...
 
+	lExtKeyUsages := []string{}
+	lRejectExtensions := []string{}
+	if extKeyUsages != "" {
+		lExtKeyUsages = strings.Split(extKeyUsages, ",")
+	}
+	if rejectExtensions != "" {
+		lRejectExtensions = strings.Split(rejectExtensions, ",")
+	}
+
```
AlCutter, line 151: I think this is equivalent to:

```suggestion
		lExtKeyUsages := strings.Split(extKeyUsages, ",")
		lRejectExtensions := strings.Split(rejectExtensions, ",")
```

In which case they could then also be inlined into the instantiation of the `ValidatedLogConfig` struct below.

personalities/sctfe/config.go
```diff
@@ -174,13 +182,13 @@ func ValidateLogConfig(cfg *configpb.LogConfig, origin string, projectID string,
...
 
-	if cfg.RejectExpired && cfg.RejectUnexpired {
+	if rejectExpired && rejectUnexpired {
 		return nil, errors.New("rejecting all certificates")
 	}
 
 	// Validate the extended key usages list.
-	if len(cfg.ExtKeyUsages) > 0 {
-		for _, kuStr := range cfg.ExtKeyUsages {
+	if len(lExtKeyUsages) > 0 {
```
AlCutter, line 190: Could the `if` be removed? (I think the `for`  will just not do anything if `lExtKeyUsages` is `[]`)

personalities/sctfe/ct_server_gcp/main.go
```diff
@@ -67,6 +67,10 @@ var (
 	bucket             = flag.String("bucket", "", "name of the bucket to store the log in")
 	spannerDB          = flag.String("spanner_db_path", "", "projects/{projectId}/instances/{instanceId}/databases/{databaseId}")
 	rootsPemFile       = flag.String("roots_pem_file", "", "Path to the file containing root certificates that are acceptable to the log. The certs are served through get-roots endpoint.")
+	rejectExpired      = flag.Bool("reject_expired", false, "if true then the certificate validity period will be checked against the current time during the validation of submissions. This will cause expired certificates to be rejected.")
```
AlCutter, line 70: nit: there's a mix of initial capital and initial non-cap on the flag descriptions here, would be cool to make them all be the same :)

personalities/sctfe/ct_server_gcp/main.go
```diff
@@ -67,6 +67,10 @@ var (
...
+	rejectUnexpired    = flag.Bool("reject_unexpired", false, "If reject_unexpired is true then CTFE rejects certificates that are either currently valid or not yet valid.")
+	extKeyUsages       = flag.String("ext_key_usages", "", "If set, ext_key_usages will restrict the set of such usages that the server will accept. By default all are accepted. The values specified must be ones known to the x509 package.")
+	rejectExtensions   = flag.String("reject_extension", "", "A list of X.509 extension OIDs, in dotted string form (e.g. '2.3.4.5') which should cause submissions to be rejected.")
```
AlCutter, line 73: "which, if present, should ..."?

AlCutter (approved):

personalities/sctfe/config.go
phbnf: I thought as well, but unfortunately that's not the case: see https://go.dev/play/p/aN9KQ0ljiu6 and https://pkg.go.dev/strings#Split.

AlCutter: :(

personalities/sctfe/config.go
phbnf: Eventually yes. I would like:
 - `main.go` to communicate with sctfe libraries with a struct, rather than passing all these arguments.
 - `main.go` to be the one instantiating storages
 - to simplify the layers of structs in the sctfe libarary

There's a bunch of layers, and I'm trying to remove layers one after the other. The first thing I'm trying to do, is to get rid of `cfg *configpb.LogConfig` totally, and pass everything as an argument to `ValidateConfig`.  This is not  very pretty, not the end state, and will certainly change after this, I'm still trying a few things out to find a good shape. 

For now, as an intermediary step, I'm thinking of `ValidateConfig` being more of a `ParseFlag` function / sanitize flags function, to make sure that fields are not empty and that dates don't step on each other for instance. Right now, it also parses the ExtendedKeyUsage, I might move it as well. I don't really know how much logic I want in there.

Then there's `SetUpInstance` which actually applies some logic on stuff and open the root_pem_file, or parses the oids. Logically, I think that parsing the oids is similar to parsing the ExtendedKeyUsage, so I might group them together.

Also, right now, the Tessera storage cannot be created anywhere else than in `SetupInstance` because, it needs a signer, and the signer is created in there... because we need a note.Signer for the checkpoint, but not for the SCTs.

Then theres's `newLogInfo`, and this one is a bit annoying as well because it's the one that as the TimeSource object, which is also used in tests.

personalities/sctfe/config.go
phbnf: Yes indeed!

personalities/sctfe/ct_server_gcp/main.go
phbnf: I fixed them all.

PR #232, opened 2024-09-09
Migrate timestamps from config to flags

phbnf:
Towards #88, and #200.


personalities/sctfe/config.go
```diff
@@ -185,22 +178,7 @@ func ValidateLogConfig(cfg *configpb.LogConfig, origin string, projectID string,
...
-	}
-	if limit != nil {
-		vCfg.NotAfterLimit = &time.Time{}
-		if err := limit.CheckValid(); err != nil {
-			return nil, fmt.Errorf("invalid limit timestamp: %v", err)
-		}
-		*vCfg.NotAfterLimit = limit.AsTime()
-	}
-	if start != nil && limit != nil && (*vCfg.NotAfterLimit).Before(*vCfg.NotAfterStart) {
+	if notAfterStart != nil && notAfterLimit != nil && (notAfterLimit).Before(*notAfterStart) {
```
AlCutter, line 181: This looks like a change in behaviour; previously it looks like you could configure the log to enforce either `start`, `limit`, or both, but now it will silently ignore one of these values unless both are non-`nil`). Was that intended?

personalities/sctfe/ct_server_gcp/main.go
```diff
@@ -276,3 +284,23 @@ func newGCPStorage(ctx context.Context, signer note.Signer) (*sctfe.CTStorage, e
...
+}
+
+func (t *timestampFlag) String() string {
+	if t.t != nil {
+		return t.t.Format(time.RFC3339)
+	}
+	return ""
+}
+
+func (t *timestampFlag) Set(w string) error {
```
AlCutter, line 299: Would it be a good idea to require that these timestamps are specified in UTC?

phbnf: How would you do this in practice? Manually parse the string to check that no zone other than "Z" is specified?

personalities/sctfe/config.go
phbnf: The values are passed on directly in the ValidatedConfig directly above. The old code would stage them in the LogConfig, then set them in ValidatedConfig manually, after it did a prototimestamp to time.Time conversion. The nil checks here are only to make sure that we can call .`Before()` safely.

personalities/sctfe/ct_server_gcp/main.go
AlCutter: Good question :)

Some ideas/thoughts: you could make the flag names be `..._utc` to make it clear what they should be, and then either:
- Check for strings.HasSuffix("Z"), or
- Specify "2006-01-02T15:04:05" as the layout for the call to `Parse`
?

Probably not a good idea to try to use `Time.Location()` to figure anything out after the fact, though.

personalities/sctfe/config.go
AlCutter: Ah gotcha, thanks!

AlCutter (approved):

personalities/sctfe/ct_server_gcp/main.go
phbnf: Add a check to require UTC, i.e "Z" suffix. Temporal logs are not formally defined anywhere, so nothing mandates that they all end on an UTC date. In practice, all of them do I think.  This check should incentive to stick to this, but we can't really force folks to do so.  I'm a bit reluctant to specific a layout without a "Z", time formatting is already enough of a mess :)

PR #235, opened 2024-09-09
Add bearer token support to the hammer

AlCutter:


roger2hk (changes requested):
We need to support bearer token in writer as well!

roger2hk (approved):

roger2hk (approved):

PR #236, opened 2024-09-09
Add tool to create note verifier from KMS public key

AlCutter:


go.mod
roger2hk, line 1: It's weird to see dependencies being removed when you're not removing any code.

cmd/conformance/gcp/kmsnote/main.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"cloud.google.com/go/kms/apiv1/kmspb"
+
+	"k8s.io/klog/v2"
+)
+
+const (
+	// KeyVersionNameFormat is the GCP resource identifier for a key version.
+	// google.cloud.kms.v1.CryptoKeyVersion.name
+	// https://cloud.google.com/php/docs/reference/cloud-kms/latest/V1.CryptoKeyVersion
```
roger2hk, line 40: I find it weird when the docs reference is for PHP Cloud Client Libraries.

cmd/conformance/gcp/kmsnote/main.go
```diff
@@ -0,0 +1,109 @@
...
+func main() {
+	klog.InitFlags(nil)
+	flag.Parse()
+
+	ctx := context.Background()
+	kmClient, err := kms.NewKeyManagementClient(ctx)
+	if err != nil {
+		log.Fatalf("failed to create KeyManagementClient: %v", err)
+	}
+	defer kmClient.Close()
```
roger2hk, line 61: defer func to log the error from `kmClient.Close()`?

cmd/conformance/gcp/kmsnote/main.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+func main() {
+	klog.InitFlags(nil)
+	flag.Parse()
+
+	ctx := context.Background()
+	kmClient, err := kms.NewKeyManagementClient(ctx)
+	if err != nil {
+		log.Fatalf("failed to create KeyManagementClient: %v", err)
```
roger2hk, line 59: Use `klog`?

cmd/conformance/gcp/kmsnote/main.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	defer kmClient.Close()
+
+	v, err := verifierKeyString(ctx, kmClient, *keyID, *name)
+	if err != nil {
+		klog.Exitf("Failed to generate verifier string: %v", err)
+	}
+
+	fmt.Println(v)
```
roger2hk, line 64: Can this line be removed? Is this for debug use?

roger2hk (approved):

cmd/conformance/gcp/kmsnote/main.go
AlCutter: Unless this is a `golangci-lint` forcing us to, I think it's probably not really necessary here - we'll either already have printed the note or failed to do so.

cmd/conformance/gcp/kmsnote/main.go
AlCutter: This is the most important line in the tool - this prints out the note verifier string we've just created :)

roger2hk: Ah right, should we use `klog` instead of `fmt.Println()`?

AlCutter: No - I wanted it to go to `stdout` without any decoration so that we can e.g. set ENV vars with it (`export VERIFIER=$(go run ./cmd/conformance/gcp/kmsnote ...)`), etc.

roger2hk: Gotcha, it makes sense to me. It would be great if you can put this use case into the package doc comment.

cmd/conformance/gcp/kmsnote/main.go
```diff
@@ -45,13 +45,6 @@ import (
 	"k8s.io/klog/v2"
 )
 
-const (
-	// KeyVersionNameFormat is the GCP resource identifier for a key version.
-	// google.cloud.kms.v1.CryptoKeyVersion.name
-	// https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys.cryptoKeyVersions
```
roger2hk: This link looks useful. Maybe move this to the `key_id` flag comment or the package doc?

roger2hk (approved):

roger2hk (approved):

PR #253, opened 2024-09-13
[GCP] Allow GCP storage to self-initialise

AlCutter:
This PR allows the GCP storage implementation to initialise an empty log without requiring an extra tool.

This makes log turn ups much easier, and avoids a catch-22 with the hammer which requires a `checkpoint` to be present before it will submit any entries.

We can decide later whether we prefer to have all storage implementations have this as a separate tool/explicit step.


log.go
```diff
@@ -84,6 +85,12 @@ func ResolveStorageOptions(opts ...func(*StorageOptions)) *StorageOptions {
 func WithCheckpointSignerVerifier(s note.Signer, v note.Verifier) func(*StorageOptions) {
 	return func(o *StorageOptions) {
 		o.NewCP = func(size uint64, hash []byte) ([]byte, error) {
+			// If we're signing a zero-sized tree, the tlog-checkpoint spec says (via RFFC6962) that
```
roger2hk, line 88:
```suggestion
			// If we're signing a zero-sized tree, the tlog-checkpoint spec says (via RFC6962) that
```

storage/gcp/gcp.go
```diff
@@ -178,6 +186,40 @@ func (s *Storage) Get(ctx context.Context, path string) ([]byte, error) {
...
+func (s *Storage) init(ctx context.Context) error {
+	cpRaw, err := s.Get(ctx, layout.CheckpointPath)
+	if err != nil {
+		if errors.Is(err, gcs.ErrObjectNotExist) {
+			// No checkpoint exists, do a forced (possibly empty) integration to create one
+			cctx, c := context.WithTimeout(ctx, 10*time.Second)
+			defer c()
+			if _, err := s.sequencer.consumeEntries(cctx, DefaultIntegrationSizeLimit, s.integrate, true); err != nil {
+				return fmt.Errorf("forced integrate: %v", err)
+			}
```
roger2hk, line 199: Why don't we call `updateCP` here?

deployment/live/gcp/cloudbuild/README.md
```diff
@@ -1,13 +1,16 @@
 # Cloudbuild Triggers and Steps
 
-This directory contains a terragrunt file that can be deployed using `terragrunt apply`
-(see links in the [deployment dir](/deployment/README.md) to install necessary tools) to
-define the necessary triggers and steps in GCB.
+This directory contains a terragrunt files to configure our CloudBuild pipeline(s).
```
roger2hk, line 3:
```suggestion
This directory contains terragrunt files to configure our CloudBuild pipeline(s).
```

deployment/live/gcp/conformance/README.md
```diff
@@ -2,15 +2,23 @@
...
-You'll need to have already configured/created a KMS key which can safely be used by the
-conformance log.
+You'll need to have already configured/created whatever service accounts + IAM permissions
+you require, and update the terragrunt files here to match.
 
-> [Warning]
-> This key should not be used elsewhere or be in any way valuable!
+## Overview
+
+This config uses the [gcp/conformance](/deployment/gcp/conformance) module to define a
```
roger2hk, line 10: Would the link be `/deployment/module/gcp/conformance`?

storage/gcp/gcp.go
AlCutter: That would not be safe - imagine that two frontends start up, one happens to get a request to `/add` which flushes and then it integrates the entry into the tree and writes an `N=1` checkpoint. Meanwhile, the other frontend starts up, sees that there's no checkpoint, and calls `updateCP` overwriting the `N=1` checkpoint with an `N=0` one.

Someone then has a bad day :)

I've added a comment about safety here.
By going through the full integration flow via the sequencer we're protected against that because it involves a transaction on the `IntCoord` table in Spanner.

deployment/live/gcp/conformance/README.md
AlCutter: good spot, ta!

storage/gcp/gcp.go
roger2hk: Gotcha, thanks for adding the comments!

roger2hk (approved):

PR #266, opened 2024-09-25
Improve codelab and record asciinema demo

mhutchinson:
Also update test keys in mysql to have key names that better represent how they should look (with schemeless URLs).


AlCutter (approved):

mhutchinson:
I'm terrible at remembering to tag issues. This is towards #9.

Issue #278, opened 2024-10-17
Make `Fetcher` client function more flexible


cmurphy:
In integrating Tessera into the Rekor log personality, we would like to have the Rekor server be able to return an inclusion proof to a client. In Tessera's model, this effectively makes Rekor its own client, since it needs to query its internal storage to fetch the tiles needed for the proof.

Tessera's `client.NewProofBuilder` constructor requires a function of type `Fetcher` as input, which takes a path string as a parameter and returns a tile. This makes sense when the tile is stored in a posix-like file storage organized by directory paths. It makes less sense when using MySQL as the storage backend, and is awkward to use when the server has direct access to the MySQL storage provider instance.

Consider the tile fetcher constructor:

https://github.com/transparency-dev/trillian-tessera/blob/0c3daf45ad69eecc6e9cd548897df65a6b08a7a5/client/client.go#L272-L286

It takes the log size, tile level, and entry index to construct the tile path and then calls the fetcher function on that path to retrieve the tile from storage.

When using the MySQL storage instance directly, the path has no meaning. So to construct a proof builder, we have to deconstruct the path, reversing the work of the `newTileFetcher`:

```
tileOnlyFetcher := func(ctx context.Context, path string) ([]byte, error) {
        pathParts := strings.SplitN(path, "/", 3)
        level, index, width, err := layout.ParseTileLevelIndexWidth(pathParts[1], pathParts[2])
        if err != nil {
                return nil, err
        }
        return api.tesseraStorage.ReadTile(ctx, level, index, width)
}
proofBuilder, err := client.NewProofBuilder(ctx, checkpoint, tileOnlyFetcher)
```

This works, it's just a bit strange and roundabout. It would be nice to be able to construct a proof builder that can work natively with any of the storage providers, or at least doesn't assume that the tile is indexed by its path.

AlCutter:
Hi @cmurphy,

Thanks for taking the time to write this up!

I think what we'll do near term is have each storage impl also provide an implementation of `client.Fetcher` for use-cases like this, where the application layer needs to read log data directly. It can do that in whatever is the most idiomatic fashion for the infra.

I understand your point about using the string containing the resource path being used here, and another way to think about it is that this string _is_ essentially the Tessera public read API. The MySQL storage is unusual/unique in that it is the one storage infra which doesn't directly expose a filesystem-like view of that API based on resource path. 

[...]

PR #281, opened 2024-10-25
Storage impls have a common read API

AlCutter:
This PR adds the MySQL "read" API onto the other storage implementations.

While we wouldn't generally expect applications to serve the read API directly unless they're using they MySQL storage implementation, we do expect that some subset of applications will want to be able to inspect the state of the tree locally (e.g. in order to expose an API which appears to implement synchronous integration). Such applications are generally going to want to have access to the read-only resources via the same "fast" path as Tessera has rather than the public path which may plausibly involve going via a CDN, waiting for replication, etc. 

#278

### DIFF
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -178,17 +178,29 @@ func (s *Storage) Add(ctx context.Context, e *tessera.Entry) tessera.IndexFuture
 	return s.queue.Add(ctx, e)
 }
 
-// Get returns the requested object.
+func (s *Storage) ReadCheckpoint(ctx context.Context) ([]byte, error) {
+	return s.get(ctx, layout.CheckpointPath)
+}
+
+func (s *Storage) ReadTile(ctx context.Context, l, i, sz uint64) ([]byte, error) {
+	return s.get(ctx, layout.TilePath(l, i, sz))
+}
+
+func (s *Storage) ReadEntryBundle(ctx context.Context, i, sz uint64) ([]byte, error) {
+	return s.get(ctx, layout.EntriesPath(i, sz))
+}
+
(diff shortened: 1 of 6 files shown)


storage/gcp/gcp.go
```diff
@@ -178,17 +178,29 @@ func (s *Storage) Add(ctx context.Context, e *tessera.Entry) tessera.IndexFuture
...
+
+func (s *Storage) ReadTile(ctx context.Context, l, i, sz uint64) ([]byte, error) {
+	return s.get(ctx, layout.TilePath(l, i, sz))
+}
+
+func (s *Storage) ReadEntryBundle(ctx context.Context, i, sz uint64) ([]byte, error) {
+	return s.get(ctx, layout.EntriesPath(i, sz))
+}
+
+// get returns the requested object.
```
mhutchinson, line 193:
```suggestion
// Get returns the requested object.
```

mhutchinson (commented):
Maybe update https://github.com/transparency-dev/trillian-tessera/blob/main/integration/storage_uniformity_test.go too?

storage/gcp/gcp.go
AlCutter: I did that to match the newly-unexported method name (same style as on `init` below).

AlCutter:
> Maybe update https://github.com/transparency-dev/trillian-tessera/blob/main/integration/storage_uniformity_test.go too?

Ah, thanks - I didn't know about that!
Done.

storage/gcp/gcp.go
mhutchinson: I must have read the diff backwards or something...

mhutchinson (approved):

PR #282, opened 2024-10-25
MySQL storage: Separate sequencing and integration

roger2hk:
#190


roger2hk:
The tests failed due to the SequencingMetadata row-level locking in sequencing process. The integration process cannot read the next sequence index.

AlCutter:
This is nearly a year old, so I'll close for now.

PR #292, opened 2024-11-06
Add tighter-scoped function types in /client

AlCutter:
This PR adapts the `client` package to use tighter scoped functions for fetching different types of log artifact, as opposed to the more generic "fetchURL" approach it previously took.

These tighter function signatures match the ones in #281, allowing personality authors to directly use storage instances when e.g. building off-line proof bundles etc.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -37,6 +37,7 @@ import (
 	"github.com/transparency-dev/trillian-tessera/api"
 	"github.com/transparency-dev/trillian-tessera/api/layout"
 	"golang.org/x/mod/sumdb/note"
+	"k8s.io/klog/v2"
 )
 
 var (
@@ -72,20 +73,16 @@ type EntryBundleFetcherFunc func(ctx context.Context, bundleIndex, logSize uint6
 //
 // rootURL should end in a trailing slash.
 // c may be nil, in which case http.DefaultClient will be used.
-func NewHTTPFetcher(rootURL string, c *http.Client) (*HTTPFetcher, error) {
-	if !strings.HasSuffix(rootURL, "/") {
-		rootURL += "/"
-	}
-	u, err := url.Parse(rootURL)
-	if err != nil {
-		return nil, fmt.Errorf("URL invalid: %v", err)
+func NewHTTPFetcher(rootURL *url.URL, c *http.Client) (*HTTPFetcher, error) {
+	if !strings.HasSuffix(rootURL.String(), "/") {
+		rootURL.Path += "/"
 	}
 	if c == nil {
 		c = http.DefaultClient
 	}
 	return &HTTPFetcher{
 		c:       c,
-		rootURL: u,
+		rootURL: rootURL,
 	}, nil
 }
 
@@ -121,6 +118,7 @@ func (h HTTPFetcher) fetch(ctx context.Context, p string) ([]byte, error) {
 	switch r.StatusCode {
 	case http.StatusOK:
 		// All good, continue below
+		break
 	case http.StatusNotFound:
 		// Need to return ErrNotExist here, by contract.
 		return nil, fmt.Errorf("get(%q): %v", u.String(), os.ErrNotExist)
@@ -128,7 +126,11 @@ func (h HTTPFetcher) fetch(ctx context.Context, p string) ([]byte, error) {
 		return nil, fmt.Errorf("get(%q): %v", u.String(), r.StatusCode)
 	}
 
-	defer r.Body.Close()
+	defer func() {
+		if err := r.Body.Close(); err != nil {
+			klog.Errorf("resp.Body.Close(): %v", err)
+		}
+	}()
 	return io.ReadAll(r.Body)
 }
 
--- a/client/client.go
+++ b/client/client.go
@@ -43,7 +43,7 @@ var (
 // checkpoint from a log's data storage.
 //
 // Note that the implementation of this MUST return (either directly or wrapped)
-// an os.ErrIsNotExit when the file referenced by path does not exist, e.g. a HTTP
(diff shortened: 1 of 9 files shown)


client/client.go
```diff
@@ -48,23 +48,53 @@ var (
 // based implementation MUST return this error when it receives a 404 StatusCode.
 type Fetcher func(ctx context.Context, path string) ([]byte, error)
 
+// CheckpointFetcher is the signature of a function which can retrieve the latest
```
mhutchinson, line 51:
```suggestion
// CheckpointFetcherFunc is the signature of a function which can retrieve the latest
```

But maybe `GetCheckpointFn` or `GetCheckpointFunc` is better? That leaves the `Fetcher` terminology still mostly pointing at path-based approaches.

client/client.go
```diff
@@ -48,23 +48,53 @@ var (
...
+// checkpoint from a log's data storage.
+type CheckpointFetcherFunc func(ctx context.Context) ([]byte, error)
+
+func CheckpointFetcher(f Fetcher) CheckpointFetcherFunc {
```
mhutchinson, line 55: Maybe it's a bit spicy, but I'd prefer the 3 functions to be something like:

```go
func FetcherToGetFuncs(f Fetcher) (GetCheckpointFunc, GetTileFunc, GetEntryBundleFunc) {
  return ....
}
```

At the point of use this would be one line instead of 3.

mhutchinson (commented):
Initial thoughts.

client/client.go
```diff
@@ -33,20 +38,131 @@ import (
...
 )
 
-// Fetcher is the signature of a function which can retrieve arbitrary files from
-// a log's data storage, via whatever appropriate mechanism.
-// The path parameter is relative to the root of the log storage.
+// CheckpointFetcherFunc is the signature of a function which can retrieve the latest
+// checkpoint from a log's data storage.
 //
 // Note that the implementation of this MUST return (either directly or wrapped)
 // an os.ErrIsNotExit when the file referenced by path does not exist, e.g. a HTTP
```
mhutchinson, line 52:
```suggestion
// an os.ErrIsNotExist when the file referenced by path does not exist, e.g. a HTTP
```

Existing typo, but I just spotted it now.

passim.

client/client.go
```diff
@@ -33,20 +38,131 @@ import (
...
+// an os.ErrIsNotExit when the file referenced by path does not exist, e.g. a HTTP
+// based implementation MUST return this error when it receives a 404 StatusCode.
+type EntryBundleFetcherFunc func(ctx context.Context, bundleIndex, logSize uint64) ([]byte, error)
+
+// NewHTTPFetcher creates a new HTTPFetcher for the log rooted at the given URL, using
+// the provided HTTP client.
+//
+// rootURL should end in a trailing slash.
+// c may be nil, in which case http.DefaultClient will be used.
+func NewHTTPFetcher(rootURL *url.URL, c *http.Client) (*HTTPFetcher, error) {
```
mhutchinson, line 77: What do you think about putting all this fetcher code in a fetcher.go file in this same package?

client/client.go
```diff
@@ -163,10 +278,14 @@ func (pb *ProofBuilder) fetchNodes(ctx context.Context, nodes proof.Nodes) ([][]
 	return hashes, nil
 }
 
+// GetTileFunc is the signature of a function which knows how to fetch a
+// specific tile.
+type GetTileFunc func(ctx context.Context, level, index uint64) (*api.HashTile, error)
+
```
mhutchinson, line 284: I think this is unused now?

internal/hammer/hammer.go
```diff
@@ -364,7 +364,7 @@ func (a *HammerAnalyser) errorLoop(ctx context.Context) {
 // dupChance provides the probability that a new leaf will be a duplicate of a previous entry.
 // Leaves will be unique if dupChance is 0, and if set to 1 then all values will be duplicates.
 // startSize should be set to the initial size of the log so that repeated runs of the
-// hammer can start seeding leaves to avoid duplicates with previous runs.
+// hammer can start seeding leaves to avoid duplicates with p:e revious runs.
```
mhutchinson, line 367: Do you use vim, BTW?

integration/storage_uniformity_test.go
```diff
@@ -25,6 +25,9 @@ import (
 
 type StorageContract interface {
 	Add(ctx context.Context, entry *tessera.Entry) tessera.IndexFuture
+	ReadCheckpoint(ctx context.Context) ([]byte, error)
+	ReadTile(ctx context.Context, level, index, treeSize uint64) ([]byte, error)
+	ReadEntryBundle(ctx context.Context, index, treeSize uint64) ([]byte, error)
 }
 
 var (
```
mhutchinson, line 33: I have a local change to add posix to this list below. Probably makes more sense in this PR than in a standalone commit though?

mhutchinson (commented):
Lots of re-wiring, but I think it's worth it. It'll be nice to not need to copy and paste that http/https/file handler another time.

client/client.go
AlCutter: done

client/client.go
AlCutter: I think that's a good idea!
done.

client/client.go
AlCutter: It's been _un-existed_ now, too :)

internal/hammer/hammer.go
AlCutter: This is referring to duplicate leaves showing a profit-to-earnings ratio resulting in _envious revenue_ (or "p:e revious", if you will).

*cough*

fixed.

mhutchinson (approved):

PR #294, opened 2024-11-07
Wrapper for Add to block until leaf is integrated

### LINKED ISSUE #193: [PersonalityKit] Provide a simple sync-add wrapper
Some personalities want to delay returning from calls to their `Add` API until the provided entry is integrated into the tree (e.g. Sigstore does this currently).

Tessera's `Add` function returns as soon as an index is durably assigned to the entry, but it does not guarantee that the entry has been integrated at that point (although it _should_ be integrated within seconds).

A simple wrapper which can be used to "decorate" Tessera's `Add` API could monitor the growth of the log (e.g. by watching `checkpoint`), and block until the observed log size is larger than the durably assigned index.

mhutchinson:
The core of this new functionality is a loop that polls for new
checkpoints peridiodically. Client code that calls Await will block
until this polling loop finds a checkpoint that is bigger than the index
being awaited, at which point it releases the client and provides it
with the checkpoint and the index.

This fixes #193.


await.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+		// Parsing a checkpoint like this is only acceptable because we're in the same binary as the
+		// log implementation that generated it and thus we can safely assume it's a well formed and
+		// validly signed checkpoint. Anyone copying similar logic into client code will get hurt.
+		parts := bytes.SplitN(rawCp, []byte{'\n'}, 3)
+		if want, got := 3, len(parts); want != got {
+			a.releaseClientsErr(fmt.Errorf("invalid checkpoint: %q", rawCp))
+			continue
+		}
+		sizeStr := string(bytes.SplitN(rawCp, []byte{'\n'}, 4)[1])
```
AlCutter, line 126: can this not use `parts[1]`?

await.go
```diff
@@ -0,0 +1,201 @@
...
+	cod := checkpointOrDeath{
+		err: err,
+	}
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	for e := a.waiters.Front(); e != nil; e = e.Next() {
+		w := e.Value.(waiter)
+		w.result <- cod
+		close(w.result)
+		a.waiters.Remove(e)
```
AlCutter, line 168: Since you have the lock and you're removing all the entries on the list, could you use `a.waiters = a.waiters.Init()` just below this loop instead?

await.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func (a *IntegrationAwaiter) releaseClients(size uint64, cp []byte) {
+	cod := checkpointOrDeath{
+		cp: cp,
+	}
+	a.mu.Lock()
+	a.size = size
+	a.checkpoint = cp
+	defer a.mu.Unlock()
```
AlCutter, line 179: nit: move this up to immediately after `Lock`?

AlCutter (approved):

await.go
```diff
@@ -0,0 +1,201 @@
...
+		waiters: list.New(),
+	}
+	go a.pollLoop(ctx, readCheckpoint, pollPeriod)
+	return a
+}
+
+// IntegrationAwaiter allows client threads to block until a leaf is both
+// sequenced and integrated. An IntegrationAwaiter should be reused for all
+// requests in the client code as there is some overhead to each one; the
+// core of an IntegrationAwaiter is a poll loop that will fetch checkpoints
```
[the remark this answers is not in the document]
mhutchinson: Updated

await.go
mhutchinson: Good spot, I extracted the variable but didn't reuse it.

await.go
mhutchinson: Yeah that's nicer.

await.go
mhutchinson: Done

AlCutter (approved):

PR #295, opened 2024-11-07
Multiball signatures

AlCutter:
This PR allows Tessera to be configured with multiple checkpoint signers.

This is useful when rotating the key of a log whilst avoiding having to coordinate a global "flag day" for everyone to switch over, or for situations where different audiences of a log require different signature schemes.

### DIFF
--- a/log.go
+++ b/log.go
@@ -23,6 +23,7 @@ import (
 	f_log "github.com/transparency-dev/formats/log"
 	"github.com/transparency-dev/trillian-tessera/api/layout"
 	"golang.org/x/mod/sumdb/note"
+	"k8s.io/klog/v2"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


log.go
```diff
@@ -80,9 +81,28 @@ func ResolveStorageOptions(opts ...func(*StorageOptions)) *StorageOptions {
...
+//
+// Zero or more dditional signers may also be provided.
+// This enables cases like:
+//   - a rolling key rotation, where checkpoints are signed by both the old and new keys for some period of time,
+//   - using different signature schemes for different audiences, etc.
+//
+// When providing additional signers, their names MUST be identical to the primary signer name, and this name will be used
+// as the checkpoint Origin line.
+//
+// Checkpoints signed by these signer(s) and verified by the provied verifier will be standard checkpoints as defined by https://c2sp.org/tlog-checkpoint.
```
mhutchinson, line 97:
```suggestion
// Checkpoints signed by these signer(s) and verified by the provided verifier will be standard checkpoints as defined by https://c2sp.org/tlog-checkpoint.
```

Issue #280, opened 2024-10-22
Use `internal` directory to hide all non-public code


mhutchinson:
This repo doesn't have any `internal` directories. While we still don't have any releases that's probably OK.

Before we tag any releases, we should perform an audit to ensure that all code that is not `internal` is intended to be public. Public code should be bound by semver, and is hard to change. Thus, I would expect that a large portion of the codebase is moved into `internal` directories.

mhutchinson:
Current list of packages and corresponding public files:

```shell
❯ go list -f '{{ .ImportPath }} {{ .GoFiles }}' ./... | grep -v internal | grep -v "\[\]"
github.com/transparency-dev/trillian-tessera [ct_only.go entry.go log.go]
github.com/transparency-dev/trillian-tessera/api [state.go]
github.com/transparency-dev/trillian-tessera/api/layout [paths.go tile.go]
github.com/transparency-dev/trillian-tessera/client [client.go]
github.com/transparency-dev/trillian-tessera/cmd/conformance/gcp [main.go]
github.com/transparency-dev/trillian-tessera/cmd/conformance/mysql [main.go]
github.com/transparency-dev/trillian-tessera/cmd/conformance/posix [main.go]
github.com/transparency-dev/trillian-tessera/cmd/examples/posix-oneshot [main.go]
github.com/transparency-dev/trillian-tessera/ctonly [ct.go]
github.com/transparency-dev/trillian-tessera/storage/gcp [gcp.go]
github.com/transparency-dev/trillian-tessera/storage/mysql [mysql.go]
github.com/transparency-dev/trillian-tessera/storage/posix [files.go]
```

This LGTM. @AlCutter at a file level, is there anything I've missed that should be hidden?

PR #299, opened 2024-11-07
Move stuff to internal and improve comments

mhutchinson:
I'm looking over https://pkg.go.dev/github.com/transparency-dev/trillian-tessera and making it clearer to newcomers.

This moves fields and types into internal, which is more refined approach to the file-based solution taken thus far to #280.


AlCutter (approved):

mhutchinson:
@AlCutter PTAL as my new commit spices things up a bit

AlCutter (approved):

Issue #24, opened 2024-06-17
AWS storage implementation


patflynn:
- [x] copy GCP code
- [x] migrate from GCS to S3
- [x] migrate Spanner to AuroraDB
- [x] migrate conformance test binary to AWS
- [x] add tests

PR #314, opened 2024-11-15
AWS storage implementation on S3 + MySQL (via RDS)

phbnf:
Towards #24 

This PR migrates from GCS to S3, and from Spanner to MySQL, thus enabling the code to run on AWS with S3 + RDS.

This PR introduces two method signature changes:
 - It removes the generation number from `getObject`, because it's not actually used, both in the GCP implementation and in this one
 - It replaces preconditions with a boolean in `setObject`, because a. there's only one type precondition used in this code, b. the ways preconditions are passed is different anyways so this looks nicer


storage/aws/aws.go
```diff
@@ -64,7 +66,7 @@ const (
 	DefaultIntegrationSizeLimit   = 5 * 4096
 )
 
-// Storage is a GCP based storage implementation for Tessera.
+// Storage is a AWS based storage implementation for Tessera.
```
AlCutter, line 69: nit: "an AWS"

storage/aws/aws.go
```diff
@@ -183,15 +190,16 @@ func (s *Storage) ReadEntryBundle(ctx context.Context, i, sz uint64) ([]byte, er
...
 	return d, err
 }
 
 // init ensures that the storage represents a log in a valid state.
 func (s *Storage) init(ctx context.Context) error {
 	_, err := s.get(ctx, layout.CheckpointPath)
 	if err != nil {
-		if errors.Is(err, gcs.ErrObjectNotExist) {
+		var nske *types.NoSuchKey
+		if errors.As(err, &nske) {
```
AlCutter, line 202: Since `nske` isn't used below, could you use `errors.Is` here?

storage/aws/aws.go
```diff
@@ -213,7 +221,7 @@ func (s *Storage) updateCP(ctx context.Context, newSize uint64, newRoot []byte)
 	if err != nil {
 		return fmt.Errorf("newCP: %v", err)
 	}
-	if err := s.objStore.setObject(ctx, layout.CheckpointPath, cpRaw, nil, ckptContType); err != nil {
+	if err := s.objStore.setObject(ctx, layout.CheckpointPath, cpRaw, false, ckptContType); err != nil {
```
AlCutter, line 224: For readability, do you think it's worth adding named consts for this new bool param?

storage/aws/aws.go
```diff
@@ -245,9 +253,10 @@ func (s *Storage) getTiles(ctx context.Context, tileIDs []storage.TileID, logSiz
 		id := id
 		errG.Go(func() error {
 			objName := layout.TilePath(id.Level, id.Index, logSize)
-			data, _, err := s.objStore.getObject(ctx, objName)
+			data, err := s.objStore.getObject(ctx, objName)
 			if err != nil {
-				if errors.Is(err, gcs.ErrObjectNotExist) {
+				var nske *types.NoSuchKey
+				if errors.As(err, &nske) {
```
AlCutter, line 259: (same observation as above about `nske` - one more below too but I won't drop a comment there)

storage/aws/aws.go
```diff
@@ -410,24 +420,37 @@ func (s *Storage) updateEntryBundles(ctx context.Context, fromSeq uint64, entrie
 	return seqErr.Wait()
 }
 
-// spannerSequencer uses Cloud Spanner to provide
+// mysqlSequencer uses MySQL to provide
 // a durable and thread/multi-process safe sequencer.
-type spannerSequencer struct {
-	dbPool         *spanner.Client
+type mysqlSequencer struct {
```
AlCutter, line 425: super nit: `mySQL` / `newMySQLSequencer` below?

storage/aws/aws.go
```diff
@@ -447,36 +472,43 @@ func newSpannerSequencer(ctx context.Context, spannerDB string, maxOutstanding u
...
+			next BIGINT UNSIGNED NOT NULL,
+			PRIMARY KEY (id)
+		)`); err != nil {
+		return err
+	}
+	if _, err := s.dbPool.ExecContext(ctx,
+		`CREATE TABLE IF NOT EXISTS Seq(
+			id INT UNSIGNED NOT NULL,
+			seq BIGINT UNSIGNED NOT NULL,
+			v LONGBLOB,
```
AlCutter, line 488: Worth noting that `LONGBLOB` allows values up to 4GB here which may be overkill, `MEDIUMBLOB` would allow 2^24, and `BLOB` 2^16 bytes.

storage/aws/aws.go
```diff
@@ -565,140 +598,151 @@ func (s *spannerSequencer) assignEntries(ctx context.Context, entries []*tessera
...
-			}
-			entries = append(entries, b...)
-			seqsConsumed = append(seqsConsumed, seq)
-			orderCheck += int64(len(b))
-		}
-		if len(seqsConsumed) == 0 && !forceUpdate {
-			klog.V(1).Info("Found no rows to sequence")
-			return nil
-		}
+	seqsConsumed := []any{}
```
AlCutter, line 631: `[]any` rather than `[]uint64` because of something to do with passing these back as SQL args later?
Might be worth a comment here about the slightly unusual type...

storage/aws/aws.go
```diff
@@ -565,140 +598,151 @@ func (s *spannerSequencer) assignEntries(ctx context.Context, entries []*tessera
...
 
-	return didWork, nil
+	// consumeFunc was successful, so we can update our coordination row, and delete the row(s) for
+	// the then consumed entries.
+	if _, err := tx.ExecContext(ctx, "UPDATE IntCoord SET seq=? WHERE id=?", orderCheck, 0); err != nil {
+		return false, fmt.Errorf("update intcoord: %v", err)
+	}
+
+	if len(seqsConsumed) > 0 {
+		q := "DELETE FROM Seq WHERE id=? AND seq IN ( " + placeholder(len(seqsConsumed)) + " )"
```
AlCutter, line 672: Not for now, but worth looking at seeing how `seq IN (?, ?, ? ...)` performs vs, say, `seq BETWEEN ? AND ?` - in theory the latter _may_ be faster as it can do it in a single range scan vs. potentially having to do a sequential bunch of "pecks", but would need to be looked into once you have it all working.

AlCutter (approved):

storage/aws/aws.go
phbnf: I thought as well, I did this first, and then ran into unexpected errors. Sadly,  I can't reproduce them.

The tl;dr is that `types.NoSuchKey` is a type, it's not an `error` returned with `errors.New()`, as opposed to `gcs. ErrObjectNotExist`. `errors.As()` does a type comparison, whereas `errors.Is()` does value comparison: see https://go.dev/blog/go1.13-errors.

I left a comment around to explain the rationale and make sure nobody undoes this.

storage/aws/aws.go
phbnf: 4GB is big. I actually wonder what would happen with such a large bunch of entries and we should probably check.

2^23-1 is 16 MB which is also a bit big...

Having said that: 
The default DefaultBatchMaxSize is 256, so that's 16384KB/256=64KB.
The vast majority of CT entries are smaller than this, *but*, we know of large certificates, with a lot of SAN like https://crt.sh/?id=10751627 which is 371KB. So 65KB for BLOB won't work. I've left a TODO for us to run tests and check what works well.

Issue #312, opened 2024-11-15
CI setup for AWS


phbnf:
- [x] provide basic backend storage setup
  -  [ ] eventually, use Aurora Serverless instead of setting up a new AuroraDB every time
- [x] integrate with ECS + Fargate to run the conformance test (I haven't tested this yet with lambda functions, but did so from a VM)
   - [x] add a VPC between the lambda function and S3
   - [x]  set-up service accounts
   - [x] use IAM for database authentication
- [x] integrate with a build pipeilne
- [x] integrate with a deployment pipeline
- [x] connect the build pipeline to git

phbnf:
I spent a bunch of time looking into (`integrate with a build pipeline`, `integrate with a deployment pipeline`), and spoke a bit with @patflynn about various options.

We can:
 1. do everything in AWS, using CodeBuild, CodePipeline. It seems like a good idea since everything will be deployed on AWS anyways.
 2. do everything in GCP, in the same way that we do this for the GCP deployment (modulo that the conformance binary and hammer would run on AWS). There would be some wrinkles about a. which container registry we push the AWS conformance binary, b. cross cloud authentication (probably requires using secret keys, etc)
 3. do everything with GitHub action. Same wrinkles as for 2., but there's a lot open source bits that we can re-use to make this work.
 
 I'm going to move forward with 3.:
  - authentication with AWS is greatly simplified by https://github.com/aws-actions/configure-aws-credentials, and https://github.com/aws-actions/amazon-ecr-login
  - after playing with the three options, I come to the conclusion that GithubAction is by far the most frictionless option: a single `.yaml` file to rule them all, can easily pass data between steps, lowest learning curve, the team is already familiar with this
  - it's different than the GCP setup, but we already use Github Actions for a few things including some integration tests, so it's not breaking ground either.
  - we can always move the GCP setup to Github Actions later if we want consistency

PR #316, opened 2024-11-18
Conformance binary for AWS

phbnf:
Towards #24 and #312 

Verified that `docker build -f ./cmd/conformance/aws/Dockerfile .` works.

This PR uses password based authentication for now, which is good enough to start with. I've tested IAM based authentication, which works. Let's get all the terraform in with passwords, and then we can eventually move to IAM.

While we're there, get rid for verifiers in `signerFromFlags` since we don't use them anymore.

### DIFF
--- a/cmd/conformance/aws/main.go
+++ b/cmd/conformance/aws/main.go
@@ -131,10 +131,10 @@ func storageConfigFromFlags() aws.Config {
 	}
 
 	dbEndpoint := fmt.Sprintf("%s:%d", *dbHost, *dbPort)
-
 	dsn := fmt.Sprintf("%s:%s@tcp(%s)/%s?allowCleartextPasswords=true",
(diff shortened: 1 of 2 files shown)


cmd/conformance/aws/main.go
```diff
@@ -102,32 +107,48 @@ func main() {
...
+	}
+	if *dbUser == "" {
+		klog.Exit("--db_user must be set")
+	}
+	// Empty passord isn't an option with AuroraDB MySQL.
+	if *dbPassword == "" {
+		klog.Exit("--db_password must be set")
+	}
+
+	var dbEndpoint string = fmt.Sprintf("%s:%d", *dbHost, *dbPort)
```
AlCutter, line 133:
```suggestion
	dbEndpoint := fmt.Sprintf("%s:%d", *dbHost, *dbPort)
```

AlCutter (approved):

PR #317, opened 2024-11-18
delete unused verifier code

phbnf:
Followup from #308

### DIFF
--- a/cmd/conformance/gcp/main.go
+++ b/cmd/conformance/gcp/main.go
@@ -37,7 +37,6 @@ var (
 	listen            = flag.String("listen", ":2024", "Address:port to listen on")
 	spanner           = flag.String("spanner", "", "Spanner resource URI ('projects/.../...')")
 	signer            = flag.String("signer", "", "Note signer to use to sign checkpoints")
-	verifier          = flag.String("verifier", "", "Note verifier corresponding to --signer")
 	additionalSigners = []string{}
 )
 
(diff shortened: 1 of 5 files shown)


cmd/conformance/gcp/main.go
```diff
@@ -37,7 +37,6 @@ var (
 	listen            = flag.String("listen", ":2024", "Address:port to listen on")
 	spanner           = flag.String("spanner", "", "Spanner resource URI ('projects/.../...')")
 	signer            = flag.String("signer", "", "Note signer to use to sign checkpoints")
-	verifier          = flag.String("verifier", "", "Note verifier corresponding to --signer")
```
mhutchinson, line 40: You'll need to update (and possibly deploy) `deployment/modules/gcp/conformance/main.tf`

deployment/modules/gcp/cloudbuild/main.tf
```diff
@@ -94,7 +94,6 @@ resource "google_cloudbuild_trigger" "docker" {
       name   = "alpine/terragrunt"
       script = <<EOT
         export TESSERA_SIGNER=$(cat /workspace/key.sec)
```
phbnf, line 96: Hmm, I think this can go as well.

cmd/conformance/gcp/main.go
phbnf: Done - I think we should be good without deploying, right? Cloudbuild should just pick everything up?

mhutchinson: @AlCutter for the right answer - I can't recall which way this is deployed right now.

mhutchinson (approved):

cmd/conformance/gcp/main.go
AlCutter: So long as this config always matches the flags in the binary I think it should be good as the conformance test is currently [re-]created from scratch each time.

AlCutter (approved):

PR #321, opened 2024-11-19
AWS storage implementation tests

phbnf:
Towards #312 and #24.

This PR builds upon the GCP tests, and migrates them to MySQL and S3.

This file contains some MySQL and some non-MySQL tests. MySQL tests should only run if a MySQL database is available, inspired by the `storage/mysq/mysql_test.go` setup. I tried a bunch of setups with "TestMain" and "t.Clean", and settled on this setup with two helper methods, and disabling parallel tests.

This PR has two commits: one that moves tests, and one that migrates them to AWS to ease reviewing (separating them into two PRs is challenging because of package names and private functions).

### DIFF
--- a/.github/workflows/go_test.yml
+++ b/.github/workflows/go_test.yml
@@ -37,3 +37,22 @@ jobs:
           mysql -e "CREATE DATABASE IF NOT EXISTS $DB_DATABASE;" -u$DB_USER -p$DB_PASSWORD
       - name: Test with Go
         run: go test -v -race ./storage/mysql/... -is_mysql_test_optional=false
+
+  test-aws-mysql:
+    env:
+      DB_DATABASE: test_tessera
+      DB_USER: root
+      DB_PASSWORD: root
+
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Start MySQL
+        run: |
+          sudo /etc/init.d/mysql start
+          mysql -e "CREATE DATABASE IF NOT EXISTS $DB_DATABASE;" -u$DB_USER -p$DB_PASSWORD
+      - name: Test with Go
+        # Parallel tests are disabled for the MySQL test database to always be in a known state.
+        run: go test -p=1 -v -race ./storage/aws/... -is_mysql_test_optional=false
--- a/storage/aws/aws_test.go
+++ b/storage/aws/aws_test.go
@@ -1,0 +1,345 @@
+// Copyright 2024 The Tessera authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gcp
+
+import (
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"errors"
+	"fmt"
+	"os"
+	"reflect"
+	"sync"
+	"testing"
+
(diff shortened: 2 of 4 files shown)


storage/aws/aws_test.go
```diff
@@ -0,0 +1,410 @@
...
+
+var (
+	mySQLURI            = flag.String("mysql_uri", "root:root@tcp(localhost:3306)/test_tessera", "Connection string for a MySQL database")
+	isMySQLTestOptional = flag.Bool("is_mysql_test_optional", true, "Boolean value to control whether the MySQL test is optional")
+)
+
+// TestMain parses flags.
+func TestMain(m *testing.M) {
+	klog.InitFlags(nil)
+	flag.Parse()
```
AlCutter, line 55: You _may_ not need the call to `flag.Parse` here as `M.Run()` will parse flags if they've not already been parsed: https://cs.opensource.google/go/go/+/refs/tags/go1.23.3:src/testing/testing.go;l=1942-1945

storage/aws/aws_test.go
```diff
@@ -0,0 +1,410 @@
...
+	os.Exit(m.Run())
+}
+
+// canSkipMySQLTest checks if the test MySQL db is available and if not, if the test can be skipped.
```
AlCutter, line 59:
```suggestion
// canSkipMySQLTest checks if the test MySQL db is available and, if not, the test can be skipped.
```

storage/aws/aws_test.go
```diff
@@ -0,0 +1,410 @@
...
+//
+// Use this method before every MySQL test, and if it return true, exist the test.
```
AlCutter, line 61:
```suggestion
// Use this method before every MySQL test, and if it returns true, exit the test.
```

storage/aws/aws_test.go
```diff
@@ -0,0 +1,410 @@
...
+	if _, err := db.ExecContext(ctx, "DROP TABLE IF EXISTS `Seq`, `SeqCoord`, `IntCoord`"); err != nil {
+		t.Fatalf("failed to drop all tables: %v", err)
+	}
+}
+
+func TestMySQLSequencerAssignEntries(t *testing.T) {
+	ctx := context.Background()
+	if canSkipMySQLTest(t, ctx) {
+		klog.Warningf("MySQL not available, skipping %s", t.Name())
+		return
```
AlCutter, line 114: Could you use `t.Skip("MySQL not available)` here (& elsewhere)?

AlCutter (approved):

.github/workflows/go_test.yml
```diff
@@ -37,3 +37,22 @@ jobs:
...
+
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Start MySQL
+        run: |
+          sudo /etc/init.d/mysql start
+          mysql -e "CREATE DATABASE IF NOT EXISTS $DB_DATABASE;" -u$DB_USER -p$DB_PASSWORD
```
roger2hk, line 55: This is not the best way to start the MySQL service. I will update the one above soon.

https://docs.github.com/en/actions/use-cases-and-examples/using-containerized-services/creating-postgresql-service-containers

roger2hk: Ah, my bad. Please ignore my comment. That's for PostgreSQL. The MySQL is already pre-installed in the ubuntu-latest image.

storage/aws/aws_test.go
phbnf: Ah there is it! I was looking for t.Return.. Thanks, I've used this, and left the warning messages. Otherwise one needs to specify -v to realize that these tests are not running.

storage/aws/aws_test.go
```diff
@@ -0,0 +1,415 @@
...
+	"testing"
+
+	gcs "cloud.google.com/go/storage"
+	"github.com/aws/smithy-go"
+	"github.com/google/go-cmp/cmp"
+	tessera "github.com/transparency-dev/trillian-tessera"
+	"github.com/transparency-dev/trillian-tessera/api"
+	"github.com/transparency-dev/trillian-tessera/api/layout"
+	storage "github.com/transparency-dev/trillian-tessera/storage/internal"
+	"k8s.io/klog"
```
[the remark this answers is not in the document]
phbnf: and the aws file... thanks.

PR #323, opened 2024-11-19
Separate integration from publishing checkpoints

AlCutter:
This PR disconnects the act of integrating new entries into a tree, from the act of publishing a new checkpoint which publicly commits to that new tree state.

The rationale behind this change is two-fold:
1. Different storage implementations should be free to integrate in whatever fashion makes most sense for the infrastructure it runs on. E.g. it may make sense for some infrastructure to make  frequent smaller updates as opposed to a large batch.
2. A controlled and predictable cadence for making new checkpoints available helps the surrounding ecosystem to function well. E.g. capacity planning for services such as witnessing relies on having some idea of how many requests will be received.

Decoupling integration from publishing checkpoints helps free up implementations to be able to solve for both (1) and (2).


storage/posix/files.go
```diff
@@ -401,16 +438,58 @@ func (s *Storage) initialise(create bool) error {
...
 
-// readcheckpoint returns the latest stored checkpoint.
-func readCheckpoint(path string) ([]byte, error) {
-	return os.ReadFile(filepath.Join(path, layout.CheckpointPath))
+// readcheckpoint returns the latest stored private checkpoint.
+func (s *Storage) readCheckpoint() ([]byte, error) {
+	return os.ReadFile(filepath.Join(s.path, stateDir, layout.CheckpointPath))
+}
+
+// publishCheckpoint makes the most recently checkpoint from the stateDir available at the
```
mhutchinson, line 459:
```suggestion
// publishCheckpoint makes the most recent checkpoint from the stateDir available at the
```

mhutchinson: Comment could also explain how minStaleness will be applied.

storage/posix/files.go
```diff
@@ -401,16 +438,58 @@ func (s *Storage) initialise(create bool) error {
...
+// (public) tlog-tiles specified Checkpoint location.
+func (s *Storage) publishCheckpoint(minStaleness time.Duration) error {
+	// Lock the destination "published" checkpoint location:
+	cpF, err := lockFile(filepath.Join(s.path, layout.CheckpointPath))
+	if err != nil {
+		return fmt.Errorf("lockFile(%s): %v", layout.CheckpointPath, err)
+	}
+	defer func() {
+		if err := cpF.Close(); err != nil {
```
mhutchinson, line 468: Should this call `unlockFile` instead of close? Or just get rid of the unlockFile method because it's an alias...

storage/gcp/gcp_test.go
```diff
@@ -304,9 +305,72 @@ func TestBundleRoundtrip(t *testing.T) {
...
+			name:            "works ok",
+			cpModifiedAt:    time.Now().Add(-15 * time.Second),
+			publishInterval: 10 * time.Second,
+			wantUpdate:      true,
+		}, {
+			name:            "too soon, skip update",
+			cpModifiedAt:    time.Now().Add(-5 * time.Second),
+			publishInterval: 10 * time.Second,
+			wantUpdate:      false,
+		},
```
mhutchinson, line 335: Add a test for the case where no checkpoint exists yet?

storage/gcp/gcp_test.go
```diff
@@ -304,9 +305,72 @@ func TestBundleRoundtrip(t *testing.T) {
...
+				objStore:    m,
+				sequencer:   s,
+				entriesPath: layout.EntriesPath,
+				newCP:       func(size uint64, hash []byte) ([]byte, error) { return []byte(fmt.Sprintf("%d/%x,", size, hash)), nil },
+			}
+			// Call init so we've got a zero-sized checkpoint to work with.
+			if err := storage.init(ctx); err != nil {
+				t.Fatalf("storage.init: %v", err)
+			}
+			m.delete(layout.CheckpointPath)
```
mhutchinson, line 349: It looks like you're deleting the checkpoint here so that you can tell that something has been written a few lines down. This seems a bit hacky because the checkpoint would never be missing in reality. Is there a better way, or are you happy with this?

storage/gcp/gcp.go
```diff
@@ -647,6 +695,21 @@ func (s *spannerSequencer) consumeEntries(ctx context.Context, limit uint64, f c
 	return didWork, nil
 }
 
+// currentTree returns the size and root hash of the currently integrated tree.
+func (s *spannerSequencer) currentTree(ctx context.Context) (uint64, []byte, error) {
+	row, err := s.dbPool.Single().ReadRow(ctx, "IntCoord", spanner.Key{0}, []string{"seq", "rootHash"})
```
mhutchinson, line 700: More of an observation as stream of consciousness rather than a request for change.

I haven't looked through this code before, but it seems funky to me that the `IntCoord` table (I inferred this is for integration, not integers) is accessed via the sequencer. It looks like that ship has sailed and it would take me some more time to work out what the difference between `IntCoord` and `SeqCoord` is. I'm sure there are good reasons for it, but as a first-timer here my intuition was that sequencers would largely play around with `SeqCoord` and there would be some integrator objects that played around with `IntCoord`, with someone having the task of coordinating both of the coordinator tables. As it is, it appears the sequencer runs the whole show.

log.go
```diff
@@ -111,3 +111,10 @@ func WithPushback(maxOutstanding uint) func(*options.StorageOptions) {
 		o.PushbackMaxOutstanding = maxOutstanding
 	}
 }
+
+// WithCheckpointInterval allows the frequency at which Tessera will attempt to publish new checkpoints to be configured.
```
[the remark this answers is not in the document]
AlCutter: I've added a little bit more, PTAL, but if this is still coming up short please do make suggested edits - it's valuable that you're coming to it with "fresh eyes" as it were.

storage/internal/options.go
```diff
@@ -22,16 +22,18 @@ import (
...
 	defaults := &options.StorageOptions{
-		BatchMaxSize: DefaultBatchMaxSize,
-		BatchMaxAge:  DefaultBatchMaxAge,
-		EntriesPath:  layout.EntriesPath,
+		BatchMaxSize:       DefaultBatchMaxSize,
+		BatchMaxAge:        DefaultBatchMaxAge,
+		EntriesPath:        layout.EntriesPath,
+		CheckpointInterval: DefaultCheckpointInterval,
 	}
 	for _, opt := range opts {
```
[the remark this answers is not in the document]
AlCutter: I'm a bit on the fence with this one, it's tempting to clamp it to be `>= storage_integration_interval`, but, just for argument's sake, if there were a storage implementation which integrated once per week, would there be value in publishing once per hour to get fresh timestamped sigs and signal freshness?
_Possibly_?

OTOH there probably could be a (somewhat arbitrary) "politeness" lower bound we could clamp to _for now_? `1s`?

storage/posix/files.go
AlCutter: It should call `unlockFile`, this was an abstraction "leak" - fixed.

storage/posix/files.go
```diff
@@ -91,38 +110,53 @@ func (s *Storage) curTree() (uint64, error) {
 // operation on this file (even if it's a different FD) from this PID, or overwriting
 // of the file by *any* process breaks the lock.)
 func (s *Storage) lockCP() error {
```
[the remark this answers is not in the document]
AlCutter: Good call, I've tided all of the locking stuff up a bit, PTAL

storage/gcp/gcp.go
AlCutter: This is only because, currently, this is a synchronous integrating storage and so there isn't really the separation that's present in say `gcp`. This could easily be changed in the future, though.

storage/gcp/gcp_test.go
AlCutter: That's already implicitly covered - `init()` creates & publishes a zero-sized checkpoint if one doesn't already exist.

PR #330, opened 2024-11-21
AWS backend terraform + Readme

phbnf:
Towards #312

This PR introduces Terraform for AWS storage backend. More work is underway for the other components, see #312.

Putting aside the fact that this runs on AWS (RDS + S3) rather than GCP (Spanner + GCS) , other differences are:
 - AWS doesn't really have a concept of human readable project IDs, project IDs are numbers, so I've introduced a human readable "prefix" used across resources instead
 - Terragrunt on GCP uses GCS precondition to lock state. That's not an option (yet?) on AWS, it uses a DynamoDB database

### DIFF
--- a/deployment/modules/aws/storage/main.tf
+++ b/deployment/modules/aws/storage/main.tf
@@ -1,0 +1,51 @@
+terraform {
+  backend "s3" {}
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+      version = "5.76.0"
+    }
+  }
+}
+
+data "aws_caller_identity" "current" {}
+
+locals {
+  name = "${var.prefix_name}-${var.base_name}"
+}
+
+# Configure the AWS Provider
+provider "aws" {
+  region = var.region
+}
+
+# Resources
+
+## S3 Bucket
+resource "aws_s3_bucket" "log_bucket" {
+  bucket = "${local.name}-bucket"
+  force_destroy = var.ephemeral
+}
+
+## Aurora MySQL RDS database
+resource "aws_rds_cluster" "log_rds" {
+  apply_immediately       = true
+  cluster_identifier      = "${local.name}-cluster"
+  engine                  = "aurora-mysql"
+  # TODO(phboneff): make sure that we want to pin this
+  engine_version          = "8.0.mysql_aurora.3.05.2"
+  database_name           = "tessera"
+  master_username         = "root"
+  master_password         = "password"
+  skip_final_snapshot     = true
+  backup_retention_period = 0
+}
+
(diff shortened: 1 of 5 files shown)


deployment/modules/aws/storage/main.tf
```diff
@@ -0,0 +1,51 @@
...
+## Aurora MySQL RDS database
+resource "aws_rds_cluster" "log_rds" {
+  apply_immediately       = true
+  cluster_identifier      = "${local.name}-cluster"
+  engine                  = "aurora-mysql"
+  # TODO(phboneff): make sure that we want to pin this
+  engine_version          = "8.0.mysql_aurora.3.05.2"
+  database_name           = "tessera"
+  master_username         = "root"
+  master_password         = "password"
```
AlCutter, line 39: Worth a TODO here to use terraform for generating random strings?

deployment/modules/aws/storage/main.tf
```diff
@@ -0,0 +1,51 @@
...
+  skip_final_snapshot     = true
+  backup_retention_period = 0
+}
+
+resource "aws_rds_cluster_instance" "cluster_instances" {
+  count              = 1
+  identifier         = "${local.name}-writer-${count.index}"
+  cluster_identifier = aws_rds_cluster.log_rds.id
+  instance_class     = "db.r5.large"
```
AlCutter, line 48: Worth making this a variable? (maybe in the future/TODO?)

AlCutter (approved):

deployment/modules/aws/storage/main.tf
phbnf: My plan is to remove this altogether, and use IAM, see #330.
I've added a TODO here to make sure that doesn't slip away.

deployment/modules/aws/storage/main.tf
phbnf: Yeah, I'm not 100% sold yet on ~all the settings that we should use here, especially for the conformance test. My gut feeling is that we'll stick to Aurora, but maybe we'll move to serverless for conformance testing, but we could also use RDS without Aurora. One interesting data point for conformance testing, is that spinning up or turning down an Aurora instance takes ~5-10 minutes.

AlCutter: > One interesting data point for conformance testing, is that spinning up or turning down an Aurora instance takes ~5-10 minutes

... wow!

Yeah, makes sense to keep all the configs as it for the moment then, can make them variables later once you've decided.

AlCutter (approved):

PR #333, opened 2024-11-25
Pull dedupe out of queue and into personality mixin

### LINKED ISSUE #331: [Deduplication] look at extracting deduplication logic from sequencer queue into mixin
We still have some logic in the internal queue used for sequencing that performs some deduplication: https://github.com/transparency-dev/trillian-tessera/blob/main/storage/internal/queue.go#L102

This is a bit of a funny situation as it's very internal to Tessera, and not easily controllable by users. It also has some peculiarities because it only dedupes within an in-memory batch. This means the following code could lead to 1 or 2 entries being in the log, depending on whether a batch was flushed in between the calls:

```
s.Add(x)
s.Add(x)
```

Take a look at pulling out this deduplication into a mixin similar to the `Awaiter`. This will let personalities have more control over how this works, and Tessera will have less code, and be more predictable.

mhutchinson:
This allows the deduplication behaviour to be more precisely controlled by the personality. Previously deduping was always performed within a single batch flushed from the queue, and this could not be disabled. Now, deduping can be disabled, or made to apply across batches if a large enough cache size is set.

The conformance binaries have been updated to use this new mixin, as this is the closest to a no-op change I can make. We may want to look at making the cache size configurable and see how this affects performance.

This fixes #331.

### DIFF
--- a/cmd/conformance/aws/main.go
+++ b/cmd/conformance/aws/main.go
@@ -70,6 +70,7 @@ func main() {
 	if err != nil {
 		klog.Exitf("Failed to create new AWS storage: %v", err)
 	}
+	dedupe := tessera.NewInMemoryDedupe(storage.Add, 256)
 
 	// Expose a HTTP handler for the conformance test writes.
 	// This should accept arbitrary bytes POSTed to /add, and return an ascii
@@ -81,7 +82,7 @@ func main() {
 			return
 		}
 
-		idx, err := storage.Add(r.Context(), tessera.NewEntry(b))()
+		idx, err := dedupe.Add(r.Context(), tessera.NewEntry(b))()
 		if err != nil {
 			if errors.Is(err, tessera.ErrPushback) {
 				w.Header().Add("Retry-After", "1")
--- a/cmd/conformance/aws/main.go
+++ b/cmd/conformance/aws/main.go
@@ -70,7 +70,7 @@ func main() {
 	if err != nil {
 		klog.Exitf("Failed to create new AWS storage: %v", err)
 	}
-	dedupe := tessera.NewInMemoryDedupe(storage.Add, 256)
+	dedupedStorage := tessera.NewInMemoryDedupe(storage.Add, 256)
 
 	// Expose a HTTP handler for the conformance test writes.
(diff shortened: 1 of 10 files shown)


phbnf:
There's already some deduplication code that was written for CT, it's Best Effort deduplication that eventually converges on always returning the smallest index possible for a submitted leaf. The plan is to take what we can out of the static-ct repo later on, and put it back in this tessera repo. This PR will change how to reason about this, both for in terms of how we align  the existing and new deduplication APIs, and what happens if both deduplications are chained together (I think it will make convergence properties slightly different). Sounds good to me to move forward with it if you want this out before Alpha, and we can re-think how the personality mixins work with one another later?

mhutchinson:
> Sounds good to me to move forward with it if you want this out before Alpha, and we can re-think how the personality mixins work with one another later?

Great I'll merge this if I can get an approval. Until we hit 1.0 all parts of the API are still open to change. Ideally the core parts will change less, and the mixins are more open to change. So sounds like we're aligned that this mixin works for now but can be changed if needed.

FWIW I really like the approach I've taken here of making the mixins that modify the `Add` call have the same signature so you can use them as a functional decorator pattern. But if it needs to change then it can. 🙃

cmd/conformance/aws/main.go
```diff
@@ -81,7 +82,7 @@ func main() {
 			return
 		}
 
-		idx, err := storage.Add(r.Context(), tessera.NewEntry(b))()
+		idx, err := dedupe.Add(r.Context(), tessera.NewEntry(b))()
```
roger2hk, line 85: I find it a bit weird when calling the `dedupe.Add`. It sounds like it's adding the entry to the dedupe queue instead of the storage. Can we rename `dedupe` to something else like `dedupedStorage`?

roger2hk (approved):
I was expecting the dedupe mixin to be defined as one of the `*options.StorageOptions`. Using the decorator pattern seems to be a better idea.

cmd/conformance/aws/main.go
mhutchinson: Done.

PR #343, opened 2024-11-27
[Dedupe] Change in-memory dedupe to return function instead of object

mhutchinson:
Having an object with a single method left a large surface area to be documented (the object, its method, and the New method) and more cognitive load for users. This way is clearer that it simply wrap the Add function.


AlCutter (approved):

AlCutter:
Nice, I think this is cleaner for the mo'.

Issue #309, opened 2024-11-13
Add support for synchronous witnessing


AlCutter:
Tessera should make it as easy as possible for logs built using it to integrate with the growing public witness network.

Issue #198, opened 2024-09-03
CT Checkpoint updater


phbnf:
Towards #88 

CT checkpoints contain a timestamp, used amongst other things, to measure CT logs MMD. That means that a checkpoint needs to be updated every X amount of time, irregardless of whether or not new entries have been integrated.

Today, Tessera only updates checkpoints (with a new timestamp) when new entries are integrated. We need a new mechanism to allow CT logs to update their checkpoints, even if no new entries have been integrated.

**Option 1: Support this in personalities only, with a fully decoupled checkpoint writer.**
- That means that there will be at least two checkpoints writer: one from Tessera, one from the personality. We need a mechanism to ensure that Tessera's and the personality's checkpoint writers don't step on each other. Otherwise, a personality checkpoint writer might rollback a Tessera checkpoint update. 
  - For buckety storage systems supporting preconditions, we could rely on preconditions
  - For non buckety storage systems, that might be more of a problem, and might require the personality to get a lock on Tessera's internal storage systems.
- Another downside is that a personality would have to fiddle with Tessera's internal storage, which, even if it's just the checkpoint, isn't great. 
- If there are multiple servers, there will be multiple Tessera checkpoints writers as well, which is fine since Tessera make sure that they don't step on one another.
- There would also be multiple personality checkpoints, but it's less of a problem if these step over each other. They could be staggered over time for instance. Otherwise, preconditions - when they are a feature, would help.

**Option 2: Support this as a Tessera option**
- This is really a CT requirement (other checkpoints don't include timestamps), so it should probably be a ct_only option.
- Every Tessera storage system implementation would implement its own updater, and handle locks, shall they want to support this option. We might be able to provide a mixin, but it's not 100% straightforward.

----

Tessera's Alpha release will support CT on AWS and GCP, both of which use buckety storage systems, and allow for preconditions (as of a few days for AWS! https://aws.amazon.com/about-aws/whats-new/2024/08/amazon-s3-conditional-writes/). So option 1 should be enough to start with. But given the known limitations, we might as well go for option 2 directly? 

I'm happy to work on either implementation, but I'm interested to know what others think about this.

mhutchinson:
I'll just float an in-between option:

## Option 3

- Tessera has a function `CreateCheckpoint()`
- Personalities can call this at whatever interval their policy dictates where a new checkpoint needs to be created

[...]

PR #344, opened 2024-11-28
[MySQL] Separate integration from publishing checkpoints

AlCutter:
This PR disconnects the act of integrating new entries into a tree, from the act of publishing a new checkpoint which publicly commits to that new tree state for the MySQL storage implementation.

See #323 for more details on the rationale.

Helps with #309 and #198.

### DIFF
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -495,7 +495,7 @@ func (s *Storage) integrate(ctx context.Context, tx *sql.Tx, fromSeq uint64, ent
 		}
 	}
 
-	// Write new checkpoint.
+	// Write new tree state.
 	if err := s.writeTreeState(ctx, tx, newSize, newRoot); err != nil {
 		return fmt.Errorf("writeCheckpoint: %w", err)
 	}
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -22,27 +22,31 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/transparency-dev/merkle/rfc6962"
 	tessera "github.com/transparency-dev/trillian-tessera"
 	"github.com/transparency-dev/trillian-tessera/api"
 	options "github.com/transparency-dev/trillian-tessera/internal/options"
-	"github.com/transparency-dev/trillian-tessera/internal/parse"
-	"github.com/transparency-dev/trillian-tessera/storage/internal"
+	storage "github.com/transparency-dev/trillian-tessera/storage/internal"
 	"k8s.io/klog/v2"
 )
 
 const (
-	selectCheckpointByIDSQL          = "SELECT `note` FROM `Checkpoint` WHERE `id` = ?"
+	selectCheckpointByIDSQL          = "SELECT `note`, `published_at` FROM `Checkpoint` WHERE `id` = ?"
 	selectCheckpointByIDForUpdateSQL = selectCheckpointByIDSQL + " FOR UPDATE"
-	replaceCheckpointSQL             = "REPLACE INTO `Checkpoint` (`id`, `note`) VALUES (?, ?)"
+	replaceCheckpointSQL             = "REPLACE INTO `Checkpoint` (`id`, `note`, `published_at`) VALUES (?, ?, ?)"
+	selectTreeStateByIDSQL           = "SELECT `size`, `root` FROM `TreeState` WHERE `id` = ?"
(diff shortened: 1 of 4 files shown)


storage/mysql/schema.sql
```diff
@@ -14,12 +14,28 @@
...
+-- "Checkpoint" table stores a single row that records the latest _published_ checkpoint for the log.
+-- This is stored separately from the TreeState in order to enable publishing of commitments to updated tree states to happen
+-- on an indepentent timeframe to the internal updating of state.
 CREATE TABLE IF NOT EXISTS `Checkpoint` (
   -- id is expected to be always 0 to maintain a maximum of a single row.
   `id`    INT UNSIGNED NOT NULL,
   -- note is the text signed by one or more keys in the checkpoint format. See https://c2sp.org/tlog-checkpoint and https://c2sp.org/signed-note.
   `note`  MEDIUMBLOB NOT NULL,
+  -- published_at is the millisecond UNIX timestamp of when this row was written.
+  `published_at` BIGINT NOT NULL,
```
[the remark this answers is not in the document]
AlCutter: This is to keep it the same type, all the way down to `time.UnixMillis` in the code.

storage/mysql/mysql.go
```diff
@@ -113,24 +150,84 @@ func (s *Storage) ReadCheckpoint(ctx context.Context) ([]byte, error) {
...
 	return checkpoint, nil
 }
 
-// writeCheckpoint stores the log signed checkpoint.
-func (s *Storage) writeCheckpoint(ctx context.Context, tx *sql.Tx, size uint64, rootHash []byte) error {
-	rawCheckpoint, err := s.newCheckpoint(size, rootHash)
+// publishCheckpoint creates a new checkpoint for the given size and root hash, and stores it in the
+// Checkpoint table.
+func (s *Storage) publishCheckpoint(ctx context.Context, i time.Duration) error {
+	tx, err := s.db.Begin()
```
roger2hk, line 166: Do we want to use `s.db.BeginTx(ctx, nil)` to pass the `ctx`?

storage/mysql/mysql.go
```diff
@@ -113,24 +150,84 @@ func (s *Storage) ReadCheckpoint(ctx context.Context) ([]byte, error) {
...
 	}
 	return checkpoint, nil
 }
 
-// writeCheckpoint stores the log signed checkpoint.
-func (s *Storage) writeCheckpoint(ctx context.Context, tx *sql.Tx, size uint64, rootHash []byte) error {
-	rawCheckpoint, err := s.newCheckpoint(size, rootHash)
+// publishCheckpoint creates a new checkpoint for the given size and root hash, and stores it in the
+// Checkpoint table.
+func (s *Storage) publishCheckpoint(ctx context.Context, i time.Duration) error {
```
roger2hk, line 165: Rename `i` as `interval`? I took me a second to realise this is the checkpoint publish interval on L181.

roger2hk (approved):
Shall we align the wrapped error? Some of them are wrapped `%w` while some of them are not `%v`.

AlCutter:
> Shall we align the wrapped error? Some of them are wrapped `%w` while some of them are not `%v`.

No, don't use `%w` without a reason - only do it when you're explicitly making the decision to allow callers of the func to be able to be aware of implementation details  (https://go.dev/wiki/ErrorValueFAQ#i-am-already-using-fmterrorf-with-v-or-s-to-provide-context-for-an-error-when-should-i-switch-to-w).

storage/mysql/mysql.go
AlCutter: Good catch, ta!

roger2hk:
> > Shall we align the wrapped error? Some of them are wrapped `%w` while some of them are not `%v`.
> 
> No, don't use `%w` without a reason - only do it when you're explicitly making the decision to allow callers of the func to be able to be aware of implementation details (https://go.dev/wiki/ErrorValueFAQ#i-am-already-using-fmterrorf-with-v-or-s-to-provide-context-for-an-error-when-should-i-switch-to-w).

Right, I see some of the error messages are using `%w`. Shall we convert them back to `%v` in a follow up PR?

AlCutter:
> > > Shall we align the wrapped error? Some of them are wrapped `%w` while some of them are not `%v`.
> > 
> > 
> > No, don't use `%w` without a reason - only do it when you're explicitly making the decision to allow callers of the func to be able to be aware of implementation details (https://go.dev/wiki/ErrorValueFAQ#i-am-already-using-fmterrorf-with-v-or-s-to-provide-context-for-an-error-when-should-i-switch-to-w).
> 
> Right, I see some of the error messages are using `%w`. Shall we convert them back to `%v` in a follow up PR?

Yeah, there seems to be a mix throughout the file so probably needs some time to think about why they were added.

PR #345, opened 2024-11-28
Add a short README.md to MySQL storage

roger2hk:


storage/mysql/README.md
```diff
@@ -0,0 +1,45 @@
...
+func main() {
+    mysqlURI := "user:password@tcp(db:3306)/tessera"
+    db, err := sql.Open("mysql", mysqlURI)
+    if err != nil {
+        klog.Exitf("Failed to connect to DB: %v", err)
+    }
+
+    storage, err := mysql.New(ctx, db)
+    if err != nil {
+        klog.Exitf("Failed to create new MySQL storage: %v", err)
```
mhutchinson, line 34: Nit - you don't have an import for `klog`

mhutchinson: Either make sure it compiles, or make it clear that it's a snippet. This is 95% of the way to being a runnable sample.

mhutchinson (approved):

storage/mysql/README.md
roger2hk: Good catch. I've not only added the `klog` import but also mentioned this is a snippet.

PR #351, opened 2024-12-02
Pin terragrunt image version

AlCutter:


roger2hk (approved):

deployment/modules/gcp/cloudbuild/main.tf
```diff
@@ -116,10 +116,11 @@ resource "google_cloudbuild_trigger" "docker" {
     ## them in files under /workspace. These are needed for later steps.
     step {
       id       = "terraform_outputs"
-      name     = "alpine/terragrunt"
+      name   = "alpine/terragrunt:1.9.5"
       script   = <<EOT
         cd deployment/live/gcp/conformance/ci
         export TESSERA_SIGNER=$(cat /workspace/key.sec)
+        export TESSERA_CLOUD_RUN_DOCKER_IMAGE=unused,
```
roger2hk, line 123: Would `unset TESSERA_CLOUD_RUN_DOCKER_IMAGE` be better?

AlCutter: no, the reason it failed is because the terragrunt config expects to find that variable set - unfortunately I don't think it's possible to require env vars to be present only for _some_ terragrunt commands but not others.

roger2hk: What if you add the `DEFAULT` to `get_env()`?

`get_env(NAME, DEFAULT)`
https://terragrunt.gruntwork.io/docs/reference/built-in-functions/#get_env

AlCutter: That'll allow people trying to manually apply this themselves to get weird errors when it eventually gets to deploying CloudRun.

This seemed like the best tradeoff.

PR #356, opened 2024-12-02
[GCP Terraform] improved stability for deployment

mhutchinson:
Added missing API, and pinned the version of terraform to a known working version. 1.10.0 was recently released which breaks this in a hard to debug way.

### DIFF
--- a/deployment/modules/gcp/conformance/main.tf
+++ b/deployment/modules/gcp/conformance/main.tf
@@ -7,14 +7,19 @@ terraform {
       version = "6.1.0"
     }
   }
+
+  required_version = "= 1.9.8"
 }
 
 data "google_compute_default_service_account" "default" {
+  depends_on = [
+    google_project_service.compute_engine,
+  ]
 }
 
 locals {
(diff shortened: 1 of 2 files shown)


deployment/modules/gcp/conformance/main.tf
```diff
@@ -49,6 +54,11 @@ resource "google_project_service" "compute_engine" {
   disable_on_destroy = false
 }
 
+resource "google_project_service" "spanner_api" {
```
AlCutter, line 57: Does the spanner resource proper also need a `depends_on` link to this like you added for the `google_compute_default_service_account`?

AlCutter (approved):

deployment/modules/gcp/conformance/main.tf
mhutchinson: PTAL, I moved it to a different file and added the depends_on.

AlCutter: Cool, yeah, that's a better place.

PR #361, opened 2024-12-03
Update performance docs

AlCutter:
This PR replaces `docs/performance.md` with more of a high-level summary, and moves the more detailed information out to `PERFORMANCE.md` docs in each of the storage implementation directories.

The goal is to have the `docs/performance.md` give more of an overview of aspects which affect performance, and broad strokes on what to expect from each of the storage implementations.

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -1,8 +1,8 @@
 # Tessera Storage Performance
 
-Tessera is designed to scale up and down to meet the needs of most currently envisioned workloads in a cost-effective manner.
+Tessera is designed to scale to meet the needs of most currently envisioned workloads in a cost-effective manner.
 
-All storage backends currently implemented have been tested to meet the write-throughput of CT-scale loads without issue.
+All storage backends have been tested to meet the write-throughput of CT-scale loads without issue.
 The read API of Tessera based logs scales extremely well due to the immutable resource based approach used, which allows for:
 1. Aggressive caching to be applied, e.g. via CDN
 2. Horizontal scaling of read infrastructure (e.g. object storage)[^1]
@@ -21,12 +21,12 @@ binaries in various environments.
 Exact performance numbers are highly dependent on the exact infrastructure being used (e.g. storage type & locality, host resources
 of the machine(s) running the personality binary, network speed and weather, etc.)  If in doubt, you should run your own performance
 tests on infrastructure which is as close as possible to that which will ultimately be used to run the log in production.
-The [conformance binaries](/cmd/conformance) and [hammer tool](/internal/hammer) may be helpful for running such tests.
+The [conformance binaries](/cmd/conformance) and [hammer tool](/internal/hammer) are designed for this kind of performance testing.
 
 ### Deduplication
 
 Deduplicating incoming entries is a somewhat expensive operation in terms of both storage and throughput.
-Not all personality designs will require it, so Tessera is build such that you only incur these costs if they are necessary
+Not all personality designs will require it, so Tessera is built such that you only incur these costs if they are necessary
 for your design.
 
 Leaving deduplication disabled will greatly increase the throughput of the log, and decrease CPU and storage costs.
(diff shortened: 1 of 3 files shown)


docs/performance.md
```diff
@@ -1,194 +1,88 @@
 # Tessera Storage Performance
 
-This document describes the performance of each Trillian Tessera storage implementation under different conditions.
+Tessera is designed to scale up and down to meet the needs of most currently envisioned workloads in a cost-effective manner.
```
roger2hk, line 3: Do you mean scale out and in instead of up and down unless only one instance is allowed?

mhutchinson: I would just say that it is designed to be scalable. Whether scaling is done automatically and whether that's horizontally/vertically is outside the scope of this introductory sentence, IMO.

docs/performance.md
```diff
@@ -1,194 +1,88 @@
...
 
-> [!TIP]
-> The performance test result shows that Trillian Tessera can run on the free tier VM instance on GCP.
+All storage backends currently implemented have been tested to meet the write-throughput of CT-scale loads without issue.
```
mhutchinson, line 5:
```suggestion
All storage backends have been tested to meet the write-throughput of CT-scale loads without issue.
```

Cleaner and I think equivalent.

docs/performance.md
```diff
@@ -1,194 +1,88 @@
...
-   instance:~$ wget https://go.dev/dl/go1.23.0.linux-amd64.tar.gz
-   instance:~$ sudo rm -rf /usr/local/go && sudo tar -C /usr/local -xzf go1.23.0.linux-amd64.tar.gz
-   instance:~$ export PATH=$PATH:/usr/local/go/bin
-   instance:~$ go version
-   go version go1.23.0 linux/amd64
-   ```
+Exact performance numbers are highly dependent on the exact infrastructure being used (e.g. storage type & locality, host resources
+of the machine(s) running the personality binary, network speed and weather, etc.)  If in doubt, you should run your own performance
+tests on infrastructure which is as close as possible to that which will ultimately be used to run the log in production.
+The [conformance binaries](/cmd/conformance) and [hammer tool](/internal/hammer) may be helpful for running such tests.
```
mhutchinson, line 24:
```suggestion
The [conformance binaries](/cmd/conformance) and [hammer tool](/internal/hammer) are designed for this kind of performance testing.
```

docs/performance.md
```diff
@@ -1,194 +1,88 @@
...
 
-1. [Install Docker using the `apt` repository](https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository)
+### Deduplication
 
-1. Install Git
+Deduplicating incoming entries is a somewhat expensive operation in terms of both storage and throughput.
+Not all personality designs will require it, so Tessera is build such that you only incur these costs if they are necessary
```
mhutchinson, line 29:
```suggestion
Not all personality designs will require it, so Tessera is built such that you only incur these costs if they are necessary
```

docs/performance.md
```diff
@@ -1,194 +1,88 @@
...
+The main lever for cost vs performance on GCP is Spanner, in the form of "Performance Units" (PUs).
+PUs can be allocated in blocks of 100, and 1000 PUs is equivalent to 1 Spanner Server.
 
-   ```sh
-   hammer:~/trillian-tessera$ go run ./internal/hammer --log_public_key=transparency.dev/tessera/example+ae330e15+ASf4/L1zE859VqlfQgGzKy34l91Gl8W6wfwp+vKP62DW --log_url=http://10.128.0.3:2024 --max_read_ops=0 --num_writers=512 --max_write_ops=512
-   ```
+The table below shows some rough numbers of measured performance:
 
-### GCP Free Tier VM Instance + Cloud SQL (MySQL)
+| Spanner PUs | Num FEs | QPS no-dedup | QPS dedup |
```
mhutchinson, line 47: It's a bit odd that we're publishing the number for dedup when this code isn't availale for usage right now?

mhutchinson (approved):

docs/performance.md
AlCutter: I'll raise a PR with that code so folks can reproduce.

tracymiranda (approved):
This is great, thanks for updating!

PR #367, opened 2024-12-04
typo fix

phbnf:


AlCutter (approved):
Hmm, an ex-maple [tree] is presumably logs :?)

PR #382, opened 2024-12-05
[MySQL Conformance] fix tree size bug

### LINKED ISSUE #364: MySQL full tiles are not immutable
The Merkle tree and entries APIs for the MySQL and POSIX conformance examples behave differently, and I think the MySQL one violates the [spec](https://github.com/C2SP/C2SP/blob/main/tlog-tiles.md) ("This endpoint is immutable, so its caching headers SHOULD be long-lived.").

For each personality, I created 280 entries, so the second level 0 tile should be a partial tile with 24 entries:

```
$ for i in `seq 1 280` ; do curl -d "[${i}]" -H "Content-Type: application/data" -X POST ${WRITE_URL}add & done ; wait
```

For POSIX, the second full tile is not accessible yet:

```
$ curl -s ${READ_URL}tile/0/001 -o - 
404 page not found
```

Only the partial tile corresponding to the log size is available:

```
$ curl -i -s ${READ_URL}tile/0/001.p/24 
HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Length: 768
Content-Type: application/octet-stream
Last-Modified: Tue, 03 Dec 2024 22:17:40 GMT
Date: Tue, 03 Dec 2024 22:36:57 GMT
```

For MySQL, the second full tile is always accessible:

```
$ curl -i -s ${READ_URL}tile/0/001 
HTTP/1.1 200 OK
Date: Tue, 03 Dec 2024 22:37:46 GMT
Content-Length: 832
Content-Type: application/octet-stream
```

And it changes when I add new entries:

```
$ curl -s ${READ_URL}tile/0/001 -o - | sha256sum
47fde106d1f8dfe199a5a8a15b386a383d7293abc80adbd8edea11a0de1ae4d6  -
$ curl -d 'new entry 1' -H "Content-Type: application/data" -X POST ${WRITE_URL}add
280
$ curl -s ${READ_URL}tile/0/001 -o - | sha256sum
3f92a25712ae6a310522017f2aeeca00d96965efcd9ea66ea5b166d951395299  -
```

It's also curious that the `W` value doesn't matter at all when accessing partial tiles in MySQL:

```
$ curl -s ${READ_URL}tile/0/001.p/1 -o - | sha256sum
3f92a25712ae6a310522017f2aeeca00d96965efcd9ea66ea5b166d951395299  -
$ curl -s ${READ_URL}tile/0/001.p/24 -o - | sha256sum
3f92a25712ae6a310522017f2aeeca00d96965efcd9ea66ea5b166d951395299  -
```

but that at least is not inconsistent with the spec, just possibly a surprising difference for clients developing against it.

mhutchinson:
Major changes:
 - MySQL storage read methods return os.ErrNotExist when values aren't found
 - ReadTile returns an error if the user requests more data than we have available
 - Added tests for writing and reading data from tiles
 - Made tests hermetic (though slower) by resetting the DB for each test case

This got a bit bigger than intended.

This fixes #364.

### DIFF
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -143,7 +143,7 @@ func (s *Storage) maybeInitTree(ctx context.Context) error {
 }
 
 // ReadCheckpoint returns the latest stored checkpoint.
-// If the checkpoint is not found, nil is returned with no error.
+// If the checkpoint is not found, it returns os.ErrNotExist.
 func (s *Storage) ReadCheckpoint(ctx context.Context) ([]byte, error) {
 	row := s.db.QueryRowContext(ctx, selectCheckpointByIDSQL, checkpointID)
 	if err := row.Err(); err != nil {
@@ -208,7 +208,7 @@ type treeState struct {
 }
 
 // readTreeState returns the currently stored tree state information.
-// If there is no stored tree state, nil is returned with no error.
+// If there is no stored tree state, it returns os.ErrNotExist.
 func (s *Storage) readTreeState(ctx context.Context, tx *sql.Tx) (*treeState, error) {
 	row := tx.QueryRowContext(ctx, selectTreeStateByIDForUpdateSQL, treeStateID)
 	if err := row.Err(); err != nil {
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -259,13 +259,13 @@ func (s *Storage) ReadTile(ctx context.Context, level, index, minTreeSize uint64
 		return nil, fmt.Errorf("scan tile: %v", err)
 	}
 
-	minSize := partialTileSize(level, index, minTreeSize)
-	if minSize == 0 {
-		minSize = 256
+	requestedWidth := partialTileSize(level, index, minTreeSize)
+	if requestedWidth == 0 {
+		requestedWidth = 256
 	}
 	numEntries := uint64(len(tile) / 32)
 
-	if minSize > numEntries {
+	if requestedWidth > numEntries {
 		// If the user has requested a size larger than we have, they can't have it
 		return nil, os.ErrNotExist
 	}
--- a/storage/mysql/mysql.go
+++ b/storage/mysql/mysql.go
@@ -18,6 +18,7 @@ package mysql
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"database/sql"
 	"errors"
 	"fmt"
@@ -238,12 +239,9 @@ func (s *Storage) writeTreeState(ctx context.Context, tx *sql.Tx, size uint64, r
 // ReadTile returns a full tile or a partial tile at the given level, index and treeSize.
 // If the tile is not found, it returns os.ErrNotExist.
 //
-// TODO: Handle the following scenarios:
-// 1. Full tile request with full tile output: Return full tile.
-// 2. Full tile request with partial tile output: Return error.
-// 3. Partial tile request with full/larger partial tile output: Return trimmed partial tile with correct tile width.
-// 4. Partial tile request with partial tile (same width) output: Return partial tile.
-// 5. Partial tile request with smaller partial tile output: Return error.
+// Note that if a partial tile is requested, but a larger tile is available, this
+// will return the largest tile available. This could be trimmed to return only the
+// number of entries specifically requested if this behaviour becomes problematic.
 func (s *Storage) ReadTile(ctx context.Context, level, index, minTreeSize uint64) ([]byte, error) {
 	row := s.db.QueryRowContext(ctx, selectSubtreeByLevelAndIndexSQL, level, index)
 	if err := row.Err(); err != nil {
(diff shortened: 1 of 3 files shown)


roger2hk:
There is also a bug in the `mysql_test/TestTileRoundTrip`. The `nodeIndex` should be replaced with `nodeIndex+1` when passing into the `s.ReadTile`.

https://github.com/transparency-dev/trillian-tessera/blob/26d856727a8823585e99f9b381731476a549b987/storage/mysql/mysql_test.go#L288-L289

storage/mysql/mysql.go
```diff
@@ -277,7 +287,7 @@ func (s *Storage) writeTile(ctx context.Context, tx *sql.Tx, level, index uint64
 }
 
 // ReadEntryBundle returns the log entries at the given index.
-// If the entry bundle is not found, nil is returned with no error.
+// If the entry bundle is not found, it returns os.ErrNotFound.
```
roger2hk, line 290:
```suggestion
// If the entry bundle is not found, it returns os.ErrNotExist.
```

mhutchinson:
@roger2hk @AlCutter PTAL - this is passing the tests now.

storage/mysql/mysql.go
```diff
@@ -252,20 +253,37 @@ func (s *Storage) ReadTile(ctx context.Context, level, index, width uint64) ([]b
...
+			return nil, os.ErrNotExist
 		}
 
 		return nil, fmt.Errorf("scan tile: %v", err)
 	}
 
-	// Return nil when returning a partial tile on a full tile request.
-	if width == 256 && uint64(len(tile)/32) != width {
-		return nil, nil
+	minSize := partialTileSize(level, index, minTreeSize)
```
roger2hk, line 262: The partial tile size calculated from the minimum tree size seems to be the same as the `width` from the URL. Shall we rename `minSize` back to `width`?

mhutchinson: Renamed as agreed 1:1

storage/mysql/mysql.go
```diff
@@ -252,20 +253,37 @@ func (s *Storage) ReadTile(ctx context.Context, level, index, width uint64) ([]b
...
 	return tile, nil
 }
 
+// partialTileSize returns the expected number of leaves in a tile at the given location within
+// a tree of the specified logSize, or 0 if the tile is expected to be fully populated.
+func partialTileSize(level, index, logSize uint64) uint64 {
+	sizeAtLevel := logSize >> (level * 8)
+	fullTiles := sizeAtLevel / 256
+	if index < fullTiles {
+		return 0
```
AlCutter, line 282: You could return 256 here and get rid of the `if` on L263?
`partialTileSize` --> `tileSize`?

storage/mysql/mysql.go
```diff
@@ -252,20 +253,37 @@ func (s *Storage) ReadTile(ctx context.Context, level, index, width uint64) ([]b
...
 	}
 
-	// Return nil when returning a partial tile on a full tile request.
-	if width == 256 && uint64(len(tile)/32) != width {
-		return nil, nil
+	requestedWidth := partialTileSize(level, index, minTreeSize)
+	if requestedWidth == 0 {
+		requestedWidth = 256
+	}
+	numEntries := uint64(len(tile) / 32)
```
AlCutter, line 266: can use `sha256.Size` in place of `32` for fewer magic constants (in test, too)

storage/mysql/mysql.go
```diff
@@ -234,16 +235,16 @@ func (s *Storage) writeTreeState(ctx context.Context, tx *sql.Tx, size uint64, r
...
+// If the tile is not found, it returns os.ErrNotExist.
 //
 // TODO: Handle the following scenarios:
 // 1. Full tile request with full tile output: Return full tile.
 // 2. Full tile request with partial tile output: Return error.
 // 3. Partial tile request with full/larger partial tile output: Return trimmed partial tile with correct tile width.
 // 4. Partial tile request with partial tile (same width) output: Return partial tile.
 // 5. Partial tile request with smaller partial tile output: Return error.
-func (s *Storage) ReadTile(ctx context.Context, level, index, width uint64) ([]byte, error) {
+func (s *Storage) ReadTile(ctx context.Context, level, index, minTreeSize uint64) ([]byte, error) {
```
AlCutter, line 247: I think some of these TODOs can be marked TODONE, just 3 and 4 to ponder?

storage/mysql/mysql_test.go
```diff
@@ -163,6 +164,93 @@ func TestNew(t *testing.T) {
...
+	}
+
+	for _, test := range []struct {
+		name                   string
+		level, index, treeSize uint64
+		wantEntries            int
+		wantNotFound           bool
+	}{
+		{
+			name:  "really too small but that's ok",
```
AlCutter, line 195: Trying to come up with a crisp description of what's being tested:
"treeSize << actual size, but all requested entries are present"?

storage/mysql/mysql.go
roger2hk: The logic now combines both `2` and `5` together. Do we want to return different error?

roger2hk: Ah, please ignore my comment. My assumption was wrong. The tree size is provided by the user but not the internal tree size.

storage/mysql/mysql_test.go
mhutchinson: changed as discussed 1:1

roger2hk (approved):

storage/mysql/mysql.go
mhutchinson: Thanks!

mhutchinson (commented):
Done!

PR #385, opened 2024-12-05
AWS Design doc.

phbnf:
See the two stage commits.

Let's update the checkpoint logic in a followup PR or with #372 directly.

### DIFF
--- a/storage/aws/README.md
+++ b/storage/aws/README.md
@@ -1,12 +1,12 @@
-# GCP Design
+# AWS Design
 
-This document describes how the storage implementation for running Tessera on Google Cloud
+This document describes how the storage implementation for running Tessera on Amazon Web Services
 is intended to work.
 
 ## Overview
 
-This design takes advantage of GCS for long term storage and low cost & complexity serving of read traffic, 
-but leverages something more transactional for coordinating the cluster.
+This design takes advantage of S3 for long term storage and low cost & complexity serving of read traffic,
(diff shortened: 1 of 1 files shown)


storage/aws/README.md
```diff
@@ -1 +1,112 @@
...
+
+## Dedup
+
+Two experimental implementations have been tested which uses either Aurora MySQL,
+or a local bbolt database to store the `<identity_hash>` --> `sequence` mapping.
+They work well, but call for further stress testing and cost analysis.
+
+### Alternatives considered
+
+Other transactional storage systems are available on GCP, e.g. Redshift, RDS or
```
AlCutter, line 91: /GCP/AWS/

AlCutter (approved):

storage/aws/README.md
```diff
@@ -1 +1,112 @@
...
+
+### `SeqCoord`
+A table with a single row which is used to keep track of the next assignable sequence number.
+
+<ins>Schema:</ins>
+```
+id INT UNSIGNED NOT NULL,
+next BIGINT UNSIGNED NOT NULL,
+PRIMARY KEY (id)
+```
```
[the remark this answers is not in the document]
phbnf: But then we might move them out of aws.go if we initialize the database somewhere else. I'll just remove them.

storage/aws/README.md
phbnf: Ah! thanks.

PR #393, opened 2024-12-06
Storage API uses partial width rather than logSize

AlCutter:
This PR replaces the use of `logSize` on the exported storage APIs with the explicit desired partial tile/entry bundle size.

There are different classes of users of these APIs which will naturally favour one these approaches (e.g. clients must have a checkpoint in their possession and so always know the log's size), but whereas it's trivial to go from `logSize` to the implied partial tile/entry bundle width, the inverse is not.

This PR redresses the imbalance by flipping over to using partial tile widths where appropriate, and exporting the function for calculating partial tile widths from tile coordinate and log size.

### DIFF
--- a/api/layout/example_test.go
+++ b/api/layout/example_test.go
@@ -28,19 +28,19 @@ func ExampleNodeCoordsToTileAddress() {
 }
 
 func ExampleTilePath() {
-	tilePath := layout.TilePath(0, 1234067, 315921160)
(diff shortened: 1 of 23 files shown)


mhutchinson (approved):

mhutchinson:
There's a conflict between this and #398, but it's an easy fix.

PR #394, opened 2024-12-06
Storage implementations enforce minimum checkpoint intervals

AlCutter:
This PR defines minimum checkpoint publication intervals for the storage implementations, and they will now return an error if an invalid interval is set via the `WithCheckpointInterval` option.

### DIFF
--- a/storage/aws/aws.go
+++ b/storage/aws/aws.go
@@ -59,8 +59,10 @@ import (
 )
 
 const (
-	logContType  = "application/octet-stream"
-	ckptContType = "text/plain; charset=utf-8"
+	entryBundleSize       = 256
+	logContType           = "application/octet-stream"
+	ckptContType          = "text/plain; charset=utf-8"
+	minCheckpointInterval = 1 * time.Second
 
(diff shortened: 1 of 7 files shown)


storage/aws/aws.go
```diff
@@ -59,9 +59,10 @@ import (
 )
 
 const (
-	entryBundleSize = 256
-	logContType     = "application/octet-stream"
-	ckptContType    = "text/plain; charset=utf-8"
+	entryBundleSize       = 256
+	logContType           = "application/octet-stream"
+	ckptContType          = "text/plain; charset=utf-8"
+	minCheckpointInterval = 1 * time.Second
```
roger2hk, line 65:
```suggestion
	minCheckpointInterval = time.Second
```
Remove `1 * ` to align with the rest of the storage.

roger2hk (approved):

roger2hk (approved):

quinnburke:
Is there a particular reason there is a minCheckpointInterval (and that it is defaulted to 1 second)? I can't find anything about this in the code.

PR #398, opened 2024-12-09
[MySQL] complete entry bundle implementation

### LINKED ISSUE #364: MySQL full tiles are not immutable

mhutchinson:
MySQL API now checks that the number of entries returned in a bundle is not smaller than the number requested. It does this by parsing the entry bundle. This could be optimized by storing the bundle size in the table.

Now this is fixed, the conformance personality is updated to set cache headers on the entry bundles to allow aggressive caching. Also fixed the checkpoint implementation to never cache, and handle not found properly.

This fixes #364.

### DIFF
--- a/cmd/conformance/mysql/main.go
+++ b/cmd/conformance/mysql/main.go
@@ -143,15 +143,19 @@ func configureTilesReadAPI(mux *http.ServeMux, storage *mysql.Storage) {
 	mux.HandleFunc("GET /checkpoint", func(w http.ResponseWriter, r *http.Request) {
 		checkpoint, err := storage.ReadCheckpoint(r.Context())
 		if err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				w.WriteHeader(http.StatusNotFound)
+				return
+			}
 			klog.Errorf("/checkpoint: %v", err)
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
-		if checkpoint == nil {
-			w.WriteHeader(http.StatusNotFound)
-			return
-		}
 
+		// Don't cache checkpoints as the endpoint refreshes regularly.
+		// A personality that wanted to _could_ set a small cache time here which was no higher
+		// than the checkpoint publish interval.
+		w.Header().Set("Cache-Control", "no-cache")
 		if _, err := w.Write(checkpoint); err != nil {
 			klog.Errorf("/checkpoint: %v", err)
 			return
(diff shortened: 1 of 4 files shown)


mhutchinson:
@roger2hk PTAL - I redid this to use the table to store the number of entries. The implementation you reviewed parsed the bundle, but this isn't safe because CT has a different bundle format.

AlCutter (approved):

roger2hk (approved):

roger2hk:
> @roger2hk PTAL - I redid this to use the table to store the number of entries. The implementation you reviewed parsed the bundle, but this isn't safe because CT has a different bundle format.

Good catch!

PR #401, opened 2024-12-09
Just use layout.PartialTileSize

AlCutter:
This PR fixes a bug introduced in #393 where an incorrect value for entry bundle width is calculated, which will frequently result in an entry bundle incorrectly not being found.


mhutchinson (approved):
Put something in the commit message that this fixes the non-deterministic failure with posix integration test.

PR #416, opened 2024-12-12
Add cache headers to POSIX version too

lapo-luchini:
Add cache headers to POSIX version too.

### DIFF
--- a/cmd/conformance/posix/main.go
+++ b/cmd/conformance/posix/main.go
@@ -49,6 +49,13 @@ func init() {
 	})
 }
 
+func addCacheHeaders(value string, fs http.Handler) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add("Cache-Control", value)
+		fs.ServeHTTP(w, r)
+	}
+}
+
 func main() {
 	klog.InitFlags(nil)
 	flag.Parse()
@@ -84,7 +91,10 @@ func main() {
 	})
 	// Proxy all GET requests to the filesystem as a lightweight file server.
 	// This makes it easier to test this implementation from another machine.
-	http.Handle("GET /", http.FileServer(http.Dir(*storageDir)))
+	fs := http.FileServer(http.Dir(*storageDir))
+	http.Handle("GET /checkpoint", addCacheHeaders("no-cache", fs))
+	http.Handle("GET /tile/", addCacheHeaders("public, max-age=31536000, immutable", fs))
+	http.Handle("GET /", fs)
 
 	// TODO(mhutchinson): Change the listen flag to just a port, or fix up this address formatting
 	klog.Infof("Environment variables useful for accessing this log:\n"+
--- a/cmd/conformance/posix/main.go
+++ b/cmd/conformance/posix/main.go
@@ -93,7 +93,7 @@ func main() {
 	// This makes it easier to test this implementation from another machine.
 	fs := http.FileServer(http.Dir(*storageDir))
 	http.Handle("GET /checkpoint", addCacheHeaders("no-cache", fs))
-	http.Handle("GET /tile/", addCacheHeaders("public, max-age=31536000, immutable", fs))
+	http.Handle("GET /tile/", addCacheHeaders("max-age=31536000, immutable", fs))
 	http.Handle("GET /", fs)
 
 	// TODO(mhutchinson): Change the listen flag to just a port, or fix up this address formatting


cmd/conformance/posix/main.go
```diff
@@ -84,7 +91,10 @@ func main() {
 	})
 	// Proxy all GET requests to the filesystem as a lightweight file server.
 	// This makes it easier to test this implementation from another machine.
-	http.Handle("GET /", http.FileServer(http.Dir(*storageDir)))
+	fs := http.FileServer(http.Dir(*storageDir))
+	http.Handle("GET /checkpoint", addCacheHeaders("no-cache", fs))
+	http.Handle("GET /tile/", addCacheHeaders("public, max-age=31536000, immutable", fs))
```
AlCutter, line 96: I'm far from an expert on this, but from reading the [mdn docs on this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) the other day, it might be fine/better to leave `public` out - it's the default anyway for non-authorized resources, it seems like it's only needed if you want to make resources retrieved when using an `Authorization` header be cached in a public cache.

AlCutter (commented):
Thanks!

Just a small nit/question inline.

cmd/conformance/posix/main.go
lapo-luchini: Oh, I copied that from the mysql implementation (c30d02cb4c4d82b0dc2f6c2423815714b12bb559), you might want to change that there too.

AlCutter: Aha, will do - thanks!

lapo-luchini:
This has the problem of marking `immutable` also the `/tile/` sub-folders (which contain a list of files which is mutable).
I'm not sure this is a real problem, but might need some more complex path matching. (note that some of the headers wre added by my reverse proxy)

```
% xh localhost:2025/tile/0/
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Date: Thu, 12 Dec 2024 15:21:54 GMT
Last-Modified: Mon, 09 Dec 2024 22:47:14 GMT
Transfer-Encoding: chunked

<pre>
<a href="000">000</a>
<a href="000.p/">000.p/</a>
<a href="001">001</a>
<a href="001.p/">001.p/</a>
<a href="002">002</a>
<a href="002.p/">002.p/</a>
[omissis]
</pre>
```

AlCutter:
> This has the problem of marking `immutable` also the `/tile/` sub-folders (which contain a list of files which is mutable). I'm not sure this is a real problem, but might need some more complex path matching. (note that some of the headers wre added by my reverse proxy)
> 
> ```
> % xh localhost:2025/tile/0/
> HTTP/1.1 200 OK
> Cache-Control: public, max-age=31536000, immutable
> Content-Type: text/html; charset=utf-8
> Date: Thu, 12 Dec 2024 15:21:54 GMT
> Last-Modified: Mon, 09 Dec 2024 22:47:14 GMT
> Transfer-Encoding: chunked
> 
> <pre>
> <a href="000">000</a>
> <a href="000.p/">000.p/</a>
> <a href="001">001</a>
> <a href="001.p/">001.p/</a>
> <a href="002">002</a>
> <a href="002.p/">002.p/</a>
> [omissis]
> </pre>
> ```

Those partial tiles (`.p/...`) are mutable only in the sense that the tlog-tiles spec permits the log implementation to later remove them once a full tile at the same coordinates is present, so I _think_ the only issue this causes is that a cache may hang on to partial tile resources longer than it strictly needs to.

cmd/conformance/posix/main.go
AlCutter: https://github.com/transparency-dev/trillian-tessera/pull/417

lapo-luchini:
> I _think_ the only issue this causes is that a cache may hang on to partial tile resources longer than it strictly needs to.

On a fresh install I onlt see these elements, tough, but soon some `006` and `007.p` will show up. (but hidden until the reverse proxy cache expires)

```
<pre>
<a href="000">000</a>
<a href="000.p/">000.p/</a>
<a href="001">001</a>
<a href="001.p/">001.p/</a>
<a href="002">002</a>
<a href="002.p/">002.p/</a>
<a href="003">003</a>
<a href="003.p/">003.p/</a>
<a href="004">004</a>
<a href="004.p/">004.p/</a>
<a href="005">005</a>
<a href="005.p/">005.p/</a>
<a href="006.p/">006.p/</a>
</pre>
```

cmd/conformance/posix/main.go
lapo-luchini: Done.

AlCutter:
> > I _think_ the only issue this causes is that a cache may hang on to partial tile resources longer than it strictly needs to.
> 
> On a fresh install I onlt see these elements, tough, but soon some `006` and `007.p` will show up. (but hidden until the reverse proxy cache expires)
> 
> ```
> <pre>
> <a href="000">000</a>
> <a href="000.p/">000.p/</a>
> <a href="001">001</a>
> <a href="001.p/">001.p/</a>
> <a href="002">002</a>
> <a href="002.p/">002.p/</a>
> <a href="003">003</a>
> <a href="003.p/">003.p/</a>
> <a href="004">004</a>
> <a href="004.p/">004.p/</a>
> <a href="005">005</a>
> <a href="005.p/">005.p/</a>
> <a href="006.p/">006.p/</a>
> </pre>
> ```

Right - if you delete the log and later create a brand new log served at the same URL via a cache it's going to be _not good_ on various fronts :)

This _shouldn't ever_ happen with a production log (one URL <-> one log), but I'm wondering if we should lower those `max-age` values in these conformance binaries to limit damage if someone does blindly copy & paste.

I think we can merge this now, at least it brings all the conformance FEs into line, and then figure out whether to revisit the age.

AlCutter (approved):

PR #424, opened 2024-12-17
[Hammer] Throttle has mutex guarding changes

mhutchinson:
This should avoid the weird number that can be reported when changing the limits at runtime. This also cleans up the readability of the main supply loop, avoiding the need to have a label to break to.

### DIFF
--- a/internal/hammer/hammer.go
+++ b/internal/hammer/hammer.go
@@ -408,13 +408,16 @@ func NewThrottle(opsPerSecond int) *Throttle {
 }
 
 type Throttle struct {
-	opsPerSecond int
 	tokenChan    chan bool
+	mu           sync.Mutex
+	opsPerSecond int
 
 	oversupply int
 }
 
 func (t *Throttle) Increase() {
+	t.mu.Lock()
+	defer t.mu.Unlock()
 	tokenCount := t.opsPerSecond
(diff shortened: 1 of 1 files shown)


internal/hammer/hammer.go
```diff
@@ -443,20 +448,27 @@ func (t *Throttle) Run(ctx context.Context) {
...
-			}
+			ctx, cancel := context.WithTimeout(ctx, interval)
+			t.supplyTokens(ctx)
+			cancel()
+		}
+	}
+}
+
+func (t *Throttle) supplyTokens(ctx context.Context) {
+	t.mu.Lock()
```
AlCutter, line 459: Ok, so worst case the UI gets blocked for `interval` while the lock is held here and we're sleeping in the `select` because the channel is full? Which seems fine.

AlCutter (approved):

internal/hammer/hammer.go
mhutchinson: Yeah that's exactly right. I tested this by setting a large throttle value and no consumers and it makes the UI feel laggy in that situation, but it works. If we wanted to fix that we could make this throttle more complicated to allow a pending update message to be left, but that's out of scope for now.

PR #428, opened 2024-12-19
Open AWS storage to being used with non-AWS MySQL and S3 services

AlCutter:
This PR enables the `aws` storage implementation to be used with `MySQL` and `S3` compatible storage services outside of the `AWS` environment.

Such usage is unsupported; folks doing this must verify that things are working correctly themselves. That said, this change has been (very minimally and quickly) tested locally with MariaDB and a local MinIO server running in docker.

### DIFF
--- a/cmd/conformance/aws/main.go
+++ b/cmd/conformance/aws/main.go
@@ -48,7 +48,6 @@ var (
 	s3Endpoint        = flag.String("s3_endpoint", "", "Endpoint for custom non-AWS S3 service")
 	s3AccessKeyID     = flag.String("s3_access_key", "", "Access key ID for custom non-AWS S3 service")
 	s3SecretAccessKey = flag.String("s3_secret", "", "Secret access key for custom non-AWS S3 service")
-	s3UseSSL          = flag.Bool("s3_use_ssl", false, "Whether to use SSL for custom non-AWS S3 service")
 	signer            = flag.String("signer", "", "Note signer to use to sign checkpoints")
 	publishInterval   = flag.Duration("publish_interval", 3*time.Second, "How frequently to publish updated checkpoints")
 	additionalSigners = []string{}
--- a/cmd/conformance/aws/main.go
+++ b/cmd/conformance/aws/main.go
@@ -24,6 +24,9 @@ import (
 	"net/http"
 	"time"
 
+	aaws "github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
 	tessera "github.com/transparency-dev/trillian-tessera"
 	"github.com/transparency-dev/trillian-tessera/storage/aws"
 	"golang.org/x/mod/sumdb/note"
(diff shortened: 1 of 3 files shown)


mhutchinson (approved):

cmd/conformance/aws/main.go
```diff
@@ -138,8 +144,27 @@ func storageConfigFromFlags() aws.Config {
 		*dbUser, *dbPassword, dbEndpoint, *dbName,
 	)
 
+	// Configure to use MinIO Server
+	var awsConfig *aaws.Config
+	var s3Opts func(o *s3.Options)
+	if *s3Endpoint != "" {
+		const defaultRegion = "us-east-1"
```
phbnf, line 151: It looks to me that this will always set the region to `us-east-1` as soon as someone sets s3Endpoint. That will then set `SDKConfig` which means that `LoadDefaultConfig` will be skipped, thus skipping other values that would be in the default config. At the moment, we only use `LoadDefaultConfig` for S3 but that might not always be the case, so let's disambiguate this to avoid leaving something to trip on later? Note,  that I don't know how `LoadDefaultConfig` works with MinIO, but it looks like this might be tricky if someone uses both AWS and the new flags. That should not be done in the first place, but if the code doesn't prevent it (should it?!), then let's make sure there's no abiguity?

storage/aws/aws.go
```diff
@@ -110,6 +110,17 @@ type consumeFunc func(ctx context.Context, from uint64, entries []storage.Sequen
...
 // Config holds AWS project and resource configuration for a storage instance.
 type Config struct {
+	// SDKConfig is an optional AWS config to use when configuring service clients, e.g. to
+	// use non-AWS S3 or MySQL services.
+	//
+	// If nil, the value from config.LoadDefaultConfig() will be used - this is the only
+	// supported configuration.
+	SDKConfig *aws.Config
+	// S3Options is an optional function which can be used to configure the S3 library.
+	// This is primarily useful when configuring the use of non-AWS S3 or MySQL services.
```
phbnf, line 122: I think that S3Options has no impact on MySQL, right? The comment says otherwise.

cmd/conformance/aws/main.go
AlCutter: That's intentional - the idea here is that you only ever set `--s3_endpoint` if you're going "off-piste" and running outside of AWS, in which case the aws default config no longer applies.

storage/aws/aws.go
```diff
@@ -873,3 +891,13 @@ func (s *s3Storage) lastModified(ctx context.Context, obj string) (time.Time, er
 
 	return *r.LastModified, r.Body.Close()
 }
+
+func printDragonsWarning() {
+	d := `H4sIAFZYZGcAA01QMQ7EIAzbeYXV5UCqkq1bf2IFtpNuPalj334hFQdkwLGNAwBzyXnKitOiqTYj
```
phbnf, line 896: <3

storage/aws/aws.go
```diff
@@ -873,3 +891,13 @@ func (s *s3Storage) lastModified(ctx context.Context, obj string) (time.Time, er
...
+B7ZGplWEwZhZqxZ1aKuswcD0AA4GXPUhI0MEpSd5Ow09vJ+m6rVtF6m0GDccYXDZEdp9N/g1H9Pf
+Qu80vNj7tiOe0lkdc8hwZK9YxavT0+FTP++vU6DUKvpEOr1+VGTk3IBXKSX9AHz5xXRgAQAA`
+	g, _ := base64.StdEncoding.DecodeString(d)
+	r, _ := gzip.NewReader(bytes.NewReader(g))
+	t, _ := io.ReadAll(r)
+	klog.Infof("Running in non-AWS mode: here be dragons!\n%s", t)
```
phbnf, line 902: Maybe point at the docs?

storage/aws/aws.go
AlCutter: Strictly speaking you're right, but intention of the comment is to make it clear that if you're setting these options you're outside of AWS and the supported config; i.e. it's only really useful to use `S3Options` if you're configuring _the binary_ to use non-AWS services.

cmd/conformance/aws/main.go
phbnf: Things I'm thinking of:
  - If someone uses a local MySQL instance, they're "off piste", but might not experience the same behavior depending on whether they specify the S3 endpoint using the flags, or `LoadDefaultConfig`, which is..... fine because they're off piste?
 - Right now `LoadDefaultConfig` is only called for S3, but my plan was to change this such that it can use IAM authentication for MySQL instead of user and password. We might have to reorganise things then, but seems fine to me with the current code.

phbnf (approved):

cmd/conformance/aws/main.go
AlCutter: Cool, yeah - I _think_ this is only a problem if they want to use "half AWS, half off"(?) which is definitely a non-supported config, so, as you say, probably ok that it doesn't work.

I'm trying to avoid having the amazon SDK "inadvertently" populate the client with AWS credentials, endpoint URLs, or anything else it does "by default" if you're in off-piste mode.

storage/aws/aws.go
AlCutter: Couldn't pass up the option :)

PR #433, opened 2025-01-06
[Architecture] Switch over to a common impl with separate drivers

mhutchinson:
The functionality is the same, but now all personalities have a common wrapper around a driver instead of having separate implementations directly in their hand.

The big advantage of the main API being this shape is that Tessera now "feels" like a single API, with drivers to bind it to different environments. This is more similar to canonical usage of sql/db package, which many go developers are familiar with.

Another advantage realized in this change is that deduplication is now folded into the core Appender object, instead of functionally floating around alongside it and needing to be handled by the personalities.


mhutchinson:
Raised this as a draft to allow conversation to first happen about the high level approach before we get into the nuts and bolts of this change.

AlCutter (approved):

PR #438, opened 2025-01-09
Relax go version requirements

AlCutter:
This PR relaxes the go version requirement to allow users of the module to use any patch version of 1.22, with a local toolchain recommendation of 1.22.10.

Rationale similar to google/trillian#3713


AlCutter:
Ah, nm, we can't because grpc: 
```
go: go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc@v0.58.0 requires go@1.22.7, but 1.22.0 is requested
```

dnwe:
@AlCutter yeah, this is the unfortunate behaviour of the `go` directive — you get forced to use the maximum found across all of your dependencies, so as soon as someone has set a higher version you all have to bump up to match. I don't really know why the Go team permitted/required a full major.minor.patch as it has led to a lot of confusion amongst module developers

Issue #436, opened 2025-01-09
Add support for migration lifecycle mode


AlCutter:
**Description**

Tessera should support migration of log data between storage implementations so that operators are not "locked in" by early choices about infrastructure.

PR #441, opened 2025-01-10
Experimental migration support for POSIX

AlCutter:
This PR adds experimental migration support for migrating into POSIX logs.

Towards #436

### DIFF
--- a/api/layout/paths.go
+++ b/api/layout/paths.go
@@ -40,52 +40,128 @@ func EntriesPathForLogIndex(seq, logSize uint64) string {
 	return EntriesPath(seq/EntryBundleWidth, PartialTileSize(0, seq, logSize))
 }
 
-// RangeInfo represents a range of entries/hashes across one or more entry bundles/tiles.
-type RangeInfo struct {
-	// StartIndex is the index of the first entry bundle the range touches.
-	StartIndex uint64
-	// StartPartial is non-zero if the bundle at StartIndex is expected to be partial.
-	StartPartial uint8
-	// StartFirst is the index of the first entry in the StartIndex bundle to be used.
-	StartFirst uint8
-
-	// EndIndex is the index of the final entry bundle containing part of the range.
-	EndIndex uint64
-	// EndPartial is non-zero if the bundle at EndIndex is expected to be partial.
-	EndPartial uint8
-	// EndN is the number of entries from the EndIndex bundle to be used. Zero means "all".
-	EndN uint8
-}
-
-// Range calculates which bundles, and elements contained within, are necessary to cover the
-// provided [from, from+N) range of entries.
+// Range calculates the addresses of bundles, and elements contained within, that are necessary to
+// cover the provided [from, from+N) range of entries, and returns an interator which provides
+// this information for a single bundle at a time.
 //
-// If N=0, an empty slice will be returned.
-// If [from, from+N) are covered within a single bundle, a single BundleSlice will be returned.
-// Otherwise two BundleSlices will be returned: the first will contain information about the
-// first bundle of the range, and the second information about the final bundle.
-func Range(from, N, treeSize uint64) (RangeInfo, error) {
+// There are several corner-cases this function deals with, where the requested range:
+// - is spread across multiple bundles; "start" bundle, zero or more "middle" bundles of size 256; zero or one "end" bundles which may be full or partial.
+// - is entirely contained within a single bundle.
+// - may not start or end aligned with bundle boundaries.
+//
+// The comments on this func and the RangeInfo struct below refer to entries and entry bundles, but
+// because tiles and entry bundles are the same width, this function works for ranges across either.
+func Range(from, N, treeSize uint64) (RangeIterator, error) {
 	endInc := from + N - 1
 	switch {
 	case N == 0:
-		return RangeInfo{}, fmt.Errorf("empty range")
+		return RangeIterator{}, fmt.Errorf("empty range")
 	case endInc >= treeSize:
-		return RangeInfo{}, fmt.Errorf("range [%d, %d) is beyond treeSize (%d)", from, endInc, treeSize)
+		return RangeIterator{}, fmt.Errorf("range [%d, %d) is beyond treeSize (%d)", from, endInc, treeSize)
 	}
 
-	r := RangeInfo{
-		StartIndex: from / EntryBundleWidth,
-		StartFirst: uint8(from % EntryBundleWidth),
-		EndIndex:   endInc / EntryBundleWidth,
-		EndN:       uint8((endInc)%EntryBundleWidth) + 1,
+	r := RangeIterator{
+		currentIndex: (from / EntryBundleWidth) - 1,
+		startIndex:   from / EntryBundleWidth,
+		startFirst:   int(from % EntryBundleWidth),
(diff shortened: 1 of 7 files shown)


api/layout/paths.go
```diff
@@ -40,6 +40,54 @@ func EntriesPathForLogIndex(seq, logSize uint64) string {
 	return EntriesPath(seq/EntryBundleWidth, PartialTileSize(0, seq, logSize))
 }
 
+// RangeInfo represents a range of entries/hashes across one or more entry bundles/tiles.
+type RangeInfo struct {
+	// StartIndex is the index of the first entry bundle the range touches.
+	StartIndex uint64
+	// StartPartial is non-zero if the bundle at StartIndex is expected to be partial.
```
mhutchinson, line 47: Add something that this value references the first index into the bundle? i.e. `(256 * StartIndex) + StartPartial` is the index that this starts from?

At the moment the docs suggest this `StartPartial` is basically a boolean in disguise.

api/layout/paths_test.go
```diff
@@ -306,3 +308,90 @@ func TestParseTileLevelIndexPartial(t *testing.T) {
 		})
 	}
 }
+
+func TestRange(t *testing.T) {
+	for _, test := range []struct {
+		from, N, treeSize uint64
+		desc              string
+		want              []RangeInfo
+		wantErr           bool
```
mhutchinson, line 317: Looks like this ain't used.

cmd/experimental/migrate/internal/migrate.go
```diff
@@ -0,0 +1,185 @@
...
+	// sourceSize is the size of the source log.
+	sourceSize uint64
+	// sourceRoot is the root hash of the source log at sourceSize.
+	sourceRoot []byte
+
+	// todo contains work items to be completed.
+	todo chan bundle
+
+	// bundlesToMigrate is the total number of entry bundles which need to be copied.
+	bundlesToMigrate uint64
```
mhutchinson, line 47: Access to this field looks racy

mhutchinson (commented):
Still reviewing, but here's some comments.

api/layout/paths_test.go
AlCutter: Good spot, removed

cmd/experimental/migrate/internal/migrate.go
AlCutter: You're right - changed the approach

api/layout/paths.go
```diff
@@ -40,6 +41,71 @@ func EntriesPathForLogIndex(seq, logSize uint64) string {
 	return EntriesPath(seq/EntryBundleWidth, PartialTileSize(0, seq, logSize))
 }
 
+// Range returns an iterator over a list of RangeInfo structs which describe the bundles/tiles
+// necessary to cover the specified range of individual entries/hashes `[from, min(from+N, treeSize) )`.
+//
+// If from >= treeSize or N == 0, the returned iterator will yeild no elements.
```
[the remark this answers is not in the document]
AlCutter: ta!

roger2hk (commented):
Adding a `README.md` in `cmd/experimental/migrate` would be helpful to tell how to try this migration tool.

AlCutter:
> Adding a `README.md` in `cmd/experimental/migrate` would be helpful to tell how to try this migration tool.

Good call, but I'll add that in a follow up soon, this PR is already a bit large...

cmd/experimental/migrate/internal/migrate.go
mhutchinson: It doesn't even need to be a field now. Can just be a local variable?

cmd/experimental/migrate/internal/migrate.go
```diff
@@ -0,0 +1,191 @@
...
+// bundle represents the address of an individual entry bundle.
+type bundle struct {
+	Index   uint64
+	Partial uint8
+}
+
+// MigrationStorage describes the required functionality from the target storage driver.
+type MigrationStorage interface {
+	// SetEntryBundle is called to store the provided entry bundle bytes at the given coordinates.
+	SetEntryBundle(ctx context.Context, index uint64, partial uint8, bundle []byte) error
```
mhutchinson, line 60: It seems like there's an implicit assumption that a driver exposing this method will perform some timely integration ?

cmd/experimental/migrate/internal/migrate.go
AlCutter: done

cmd/experimental/migrate/internal/migrate.go
```diff
@@ -0,0 +1,189 @@
...
+//
+// Only the entry bundles are copied as the target storage is expected to integrate them and recalculate the root.
+// This is done to ensure the correctness of both the source log as well as the copy process itself.
+//
+// A call to this function will block until either the copying is done, or an error has occurred.
+// It is an error if the resource copying completes ok but the resulting root hash does not match the provided sourceRoot.
+func Migrate(ctx context.Context, numWorkers int, sourceSize uint64, sourceRoot []byte, getEntries client.EntryBundleFetcherFunc, storage MigrationStorage) error {
+	klog.Infof("Starting migration; source size %d root %x", sourceSize, sourceRoot)
+
+	// TODO store state & resume
```
mhutchinson, line 79: Naked TODO considered harmful. Before submit, can we have this fixed, a name attached, or an issue created and linked?

cmd/experimental/migrate/internal/migrate.go
```diff
@@ -0,0 +1,189 @@
...
+	// init
+	targetSize, err := m.storage.Size(ctx)
+	if err != nil {
+		return fmt.Errorf("Size: %v", err)
+	}
+	if targetSize > sourceSize {
+		return fmt.Errorf("Target size %d > source size %d", targetSize, sourceSize)
+	}
+	if targetSize == sourceSize {
+		return nil
```
mhutchinson, line 97: Should this check that the hashes match? It didn't "do" anything, but I would expect a successful call to Migrate to mean that the log now matches the source log.

cmd/experimental/migrate/internal/migrate.go
mhutchinson: Looking at the usage of this, docs should specify that this _must_ allow out-of-order bundle writes. This is an option whenever numWorkers > 1. In this case, it means that a failure and recovery must also allow the same bundle to be written again. Should this allow overwriting? Should it complain if there is any difference?

AlCutter: Yup, added that to the comment on the interface

cmd/experimental/migrate/internal/migrate.go
AlCutter: Yeah, it probably should, otherwise there's a "crack" if, e.g., someone does a migrate which fails with a root mismatch, scratches their head, re-runs it (without deleting the target data first), then it'll "succeed".

Done.

cmd/experimental/migrate/internal/migrate.go
AlCutter: I've removed the TODO - currently if the job crashes and is restarted, it'll pick up from the size of the locally integrated tree. That's likely good enough for now.

mhutchinson (approved):

PR #453, opened 2025-01-16
Output conformance bucket and use in GCB

AlCutter:
This PR should get the hammer working again in GCP CI.


roger2hk (approved):

AlCutter:
CloudBuild config applied.

Issue #450, opened 2025-01-16
Persisted state must have a compatibility version


mhutchinson:
Anywhere where we persist state that needs to be read from the running binary, we have a potential issue with data compatibility. Fortunately, the majority of the data that is written and read has its serialization format fully defined by the tlog-tiles spec, which means that this data can’t change.

There is some data which is not defined by this spec, and this data requires some consideration.

Each of the implementations stores some state that is read at runtime. The requirement to tick an implementation off the list below is that a version must be serialized in this state, and that version must be checked for compatibility at startup. Out of scope is how migration between versions happens.

- [x] AWS
- [x] GCP
- [x] MySQL
- [x] POSIX

mhutchinson:
Done!

PR #457, opened 2025-01-22
[POSIX] Ensure compatibility version

mhutchinson:
This will upgrade any existing unversioned logs to have the current version. Towards #450

Regenerated the testdata to confirm this works and checked in the .state
directory, with lock files and all. We may want to remove lock files
from disk when they are unlocked, but that's beyond the scope of this
PR.


storage/posix/files.go
```diff
@@ -418,6 +429,32 @@ type treeState struct {
...
+func (s *Storage) ensureVersion(version uint16) error {
+	versionFile := filepath.Join(s.path, stateDir, "version")
+
+	if _, err := os.Stat(versionFile); errors.Is(err, os.ErrNotExist) {
+		klog.V(1).Infof("No version file exists, creating")
+		data := make([]byte, 2)
+		binary.LittleEndian.PutUint16(data, version)
+		if err := createExclusive(versionFile, data); err != nil {
+			return fmt.Errorf("failed to create version file: %v", err)
+		}
```
AlCutter, line 443: Could pop a `return nil` in here, and then lose the final `else` and out-dent that code as the "happy path"?

storage/posix/files.go
```diff
@@ -418,6 +429,32 @@ type treeState struct {
...
 
+// ensureVersion will fail if the compatibility version stored in the state directory
+// is not the expected version. If no file exists, then it is created with the expected version.
+func (s *Storage) ensureVersion(version uint16) error {
+	versionFile := filepath.Join(s.path, stateDir, "version")
+
+	if _, err := os.Stat(versionFile); errors.Is(err, os.ErrNotExist) {
+		klog.V(1).Infof("No version file exists, creating")
+		data := make([]byte, 2)
+		binary.LittleEndian.PutUint16(data, version)
```
AlCutter, line 440: I think I'd go for [in order] either ASCII decimal (nice for debugging: "what schema version does your ./state/version file say?") or BigEndian (for consistency with the related C2SP specs, while acknowledging that this file obvs isn't covered by the spec).

AlCutter (approved):

storage/posix/files.go
mhutchinson: Done.

storage/posix/files.go
mhutchinson: Used ASCII.

AlCutter:
I think you'll need to re-create the `version` file too?

mhutchinson:
> I think you'll need to re-create the `version` file too?

I did :-)

AlCutter (approved):

PR #458, opened 2025-01-22
[MySQL] Check that the DB schema is compatible

mhutchinson:
The version is written at schema creation, and checked on startup. If they are different, it fails. Towards #450.

### DIFF
--- a/storage/mysql/schema.sql
+++ b/storage/mysql/schema.sql
@@ -21,7 +21,7 @@ CREATE TABLE IF NOT EXISTS Tessera (
   -- id is expected to be always 0 to maintain a maximum of a single row.
   `id`                   TINYINT UNSIGNED NOT NULL,
   -- compatibilityVersion is the version of this schema and the data within it.
-  `compatibilityVersion` TINYINT UNSIGNED NOT NULL,
+  `compatibilityVersion` BIGINT UNSIGNED NOT NULL,
   PRIMARY KEY (`id`)
 );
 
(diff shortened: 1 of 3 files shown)


storage/mysql/schema.sql
```diff
@@ -14,6 +14,19 @@
...
 -- MySQL version of the Trillian Tessera database schema.
 
+-- "Tessera" table stores a single row that is the version of this schema
+-- and the data formats within it. This is read at startup to prevent Tessera
+-- running against a database with an incompatible format.
+CREATE TABLE IF NOT EXISTS Tessera (
+  -- id is expected to be always 0 to maintain a maximum of a single row.
+  `id`                   TINYINT UNSIGNED NOT NULL,
+  -- compatibilityVersion is the version of this schema and the data within it.
+  `compatibilityVersion` TINYINT UNSIGNED NOT NULL,
```
AlCutter, line 24: Given there's only every going to be one row in here, I'd be tempted to just make the `compatibilityVersion` type an int64 - I agree there are unlikely to ever be more than 256 schema/encoding changes, but the larger size might open doors to dealing with unexpected situations (e.g. using individual bits as "feature flags" etc.)?

AlCutter (approved):

roger2hk (commented):
Please also add the new table in the schema section in `storage/mysql/DESIGN.md`.

storage/mysql/schema.sql
mhutchinson: Done.

mhutchinson:
> Please also add the new table in the schema section in `storage/mysql/DESIGN.md`.

Good spot - done.

PR #480, opened 2025-02-11
Lock initialise to prevent concurrent initialisation 

### LINKED ISSUE #476: Idempotent posix.New without explicit initialization
**Description**

When creating the Tessera POSIX storage, a `create` bool must be set to `true` once and only once. It is the deployer's responsibility to know if the directory has already been initialized. This complicates deployment, as the deployer needs to maintain a one-off job for tree creation and the long-running deployment.

I was curious what happens when `posix.New(create=true)` is repeatedly called, to see if I received an error reinitializing the tree, but I did not receive an error. Repeated calls to `posix.New()` with `create=true` leaves the tree in a broken state, with tiles and entries left over from previous trees.\
Digging into `Storage.initialise`, `os.MkdirAll` does not throw an error if the directory exists, and `Storage.writeTreeState()`/`createExclusive()` won't either as `os.WriteFile()` writes a temporary file that only exists for a moment and `os.Rename` will overwrite an existing file.

## Request(s)

Could `posix.New(...)` be idempotent, where the tree is implicitly created if it doesn't exist and repeated calls do not reinitialize the tree? For example, when `posix.New(...)` is called, could `Storage.initialise` check if a directory already exists and for the required state files, and skip initialization if those exist?

If it can't be idempotent, can `posix.New(...)` throw an error if the tree is already initialized, e.g. if the directory exists? This would prevent accidentally corrupting the log, though deployment would still be more complex, so idempotencey would be preferred.

haydentherapper:
If two tasks attempted to initialise an empty checkpoint, and one task
stalled while the other began accepting traffic, then the tree state and
checkpoint would be invalidated by the stalled task. This guarantees
initialisation only once per log.

Fixes #476

### DIFF
--- a/storage/posix/files.go
+++ b/storage/posix/files.go
@@ -424,7 +424,7 @@ func (a *appender) initialise() error {
 	// - The mutex `Lock()` ensures that multiple concurrent calls to this function within a task are serialised.
 	// - The POSIX `lockFile()` ensures that distinct tasks are serialised.
 	a.s.mu.Lock()
-	unlock, err := a.s.lockFile("initialise.lock")
+	unlock, err := a.s.lockFile("treeState.lock")
 	if err != nil {
 		panic(err)
 	}
@@ -499,7 +499,7 @@ func (s *Storage) ensureVersion(version uint16) error {
 func (s *Storage) writeTreeState(size uint64, root []byte) error {
 	raw, err := json.Marshal(treeState{Size: size, Root: root})
 	if err != nil {
-		return fmt.Errorf("json.Marshal: %v", err)
+		return fmt.Errorf("error in Marshal: %v", err)
 	}
 
 	if err := s.overwrite(filepath.Join(stateDir, "treeState"), raw); err != nil {
@@ -513,11 +513,11 @@ func (s *Storage) readTreeState() (uint64, []byte, error) {
 	p := filepath.Join(s.path, stateDir, "treeState")
 	raw, err := os.ReadFile(p)
 	if err != nil {
-		return 0, nil, fmt.Errorf("os.ReadFile(%q): %w", p, err)
+		return 0, nil, fmt.Errorf("error in ReadFile(%q): %w", p, err)
 	}
 	ts := &treeState{}
 	if err := json.Unmarshal(raw, ts); err != nil {
-		return 0, nil, fmt.Errorf("json.Unmarshal: %v", err)
+		return 0, nil, fmt.Errorf("error in Unmarshal: %v", err)
 	}
 	return ts.Size, ts.Root, nil
 }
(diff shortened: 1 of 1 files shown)


AlCutter:
Actually - please can you hold off on this PR for a bit?
There's a very large refactoring going on at the mo' and it's going to be tricky to merge!

haydentherapper:
Sure! I pushed up my changes locking initialise while it's running, now initialisation should only happen once per log. I'll rebase once the refactor is done.

AlCutter:
I ended up having to make some similar changes to eliminate the `create` param from the c'tor as part of the large refactor in #484 (that func now has to conform to an API).

Have a look and see whether this does what you need or if we can extract and apply bits from this PR (I'm afraid it's unlikely to be just a rebase!)

haydentherapper:
The refactor solves initialisation without `create`! There's still a chance of a race-y initialisation, so I copied in the change to lock while initializing. Also fixed a few inconsequential lint errors.

storage/posix/files.go
```diff
@@ -416,6 +416,25 @@ func (lrs *logResourceStorage) writeBundle(_ context.Context, index uint64, part
...
 func (a *appender) initialise() error {
+	// Idempotent: If folder exists, nothing happens.
+	if err := os.MkdirAll(filepath.Join(a.s.path, stateDir), dirPerm); err != nil {
+		return fmt.Errorf("failed to create log directory: %q", err)
+	}
+	// Double locking:
+	// - The mutex `Lock()` ensures that multiple concurrent calls to this function within a task are serialised.
+	// - The POSIX `lockFile()` ensures that distinct tasks are serialised.
+	a.s.mu.Lock()
+	unlock, err := a.s.lockFile("initialise.lock")
```
AlCutter, line 427: We can probably reuse the `treeState.lock` file here rather than add a new one.

storage/posix/files.go
```diff
@@ -483,7 +499,7 @@ func (s *Storage) ensureVersion(version uint16) error {
 func (s *Storage) writeTreeState(size uint64, root []byte) error {
 	raw, err := json.Marshal(treeState{Size: size, Root: root})
 	if err != nil {
-		return fmt.Errorf("Marshal: %v", err)
+		return fmt.Errorf("json.Marshal: %v", err)
```
AlCutter, line 502: nit: I'd only really add these package qualifiers if there's ambiguity about where in the method this error could've come from.

storage/posix/files.go
```diff
@@ -728,7 +763,7 @@ func (m *MigrationStorage) State(_ context.Context) (uint64, []byte, error) {
 func (m *MigrationStorage) buildTree(ctx context.Context, targetSize uint64) error {
 	// Double locking:
 	// - The mutex `Lock()` ensures that multiple concurrent calls to this function within a task are serialised.
-	// - The POSIX `lockForTreeUpdate()` ensures that distinct tasks are serialised.
+	// - The POSIX `lockFile()` ensures that distinct tasks are serialised.
```
AlCutter, line 766: good spot, ta!

AlCutter (commented):
Thanks for shifting this atop the refactor, looks good, just a couple of small suggestions inline.

storage/posix/files.go
haydentherapper: Done.

storage/posix/files.go
AlCutter: Aaah, I see!

AlCutter (approved):

AlCutter:
Nice one, thanks!

PR #481, opened 2025-02-11
Tighten up POSIX file operations

AlCutter:
This PR tightens the semantics of the file operations performed by the POSIX storage implementation.

It splits "writing files" across 3 narrower funcs which are then called by the storage driver depending on exactly which behaviour is intended.

### DIFF
--- a/storage/posix/files.go
+++ b/storage/posix/files.go
@@ -541,10 +541,20 @@ func (s *Storage) publishCheckpoint(minStaleness time.Duration) error {
 // data & close the file.
 func createExclusive(f string, d []byte) error {
 	tmpName := f + ".temp"
-	if err := os.WriteFile(tmpName, d, filePerm); err != nil {
-		return fmt.Errorf("unable to write data to temporary file: %w", err)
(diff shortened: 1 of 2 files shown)


storage/posix/files.go
```diff
@@ -528,28 +528,88 @@ func (s *Storage) publishCheckpoint(minStaleness time.Duration) error {
...
+}
+
+// createIdempotent atomically writes the provided data to a file at the provided path.
+// An error will be returned if a file already exists in the named location AND that file's
+// contents is not identical to the provided data.
+func createIdempotent(p string, d []byte) error {
+	if err := createExclusive(p, d); err != nil {
+		if errors.Is(err, os.ErrExist) {
+			if r, err := os.ReadFile(p); err != nil {
+
```
roger2hk, line 587: Is this line intended to be empty?

roger2hk (approved):

storage/posix/files.go
AlCutter: Oops, no - removed.

haydentherapper (commented):
Looks great!

Issue #474, opened 2025-02-04
[MySQL] Migration lifecycle support


AlCutter:
**Description**

The MySQL storage implementation supports the migration lifecycle mode.

Issue #473, opened 2025-02-04
[AWS] Migration lifecycle support


AlCutter:
**Description**

The AWS storage supports the migration lifecycle mode.

Issue #472, opened 2025-02-04
[POSIX] Migration lifecycle support


AlCutter:
**Description**

The POSIX storage implementation supports the migration lifecycle mode.

Issue #471, opened 2025-02-04
[GCP] Migration lifecycle support


AlCutter:
**Description**

The GCP storage implementation supports the migration lifecycle mode.

PR #482, opened 2025-02-12
Migrate storage implementations to Appender lifecycle model

AlCutter:
This PR moves all the of the storage implementations fully over to the lifecycle model introduced in #433.

The functionality related to managing an append-only log, which was previously present on the drivers themselves, is moved to a new `Appender` type. While there have been a few small refactors and tweaks to make the transition easier/cleaner, no functional changes in the way each of the drivers manages append-only logs is intended.

`StorageOptions` functionality has been moved into `tessera` and merged with the new `appendOptions` struct introduced in #443.

Towards #471 #472 #473 #474

### DIFF
--- a/cmd/conformance/aws/main.go
+++ b/cmd/conformance/aws/main.go
@@ -79,10 +79,10 @@ func main() {
 		tessera.WithBatching(1024, time.Second),
 		tessera.WithPushback(10*4096),
 		tessera.WithAppendDeduplication(tessera.InMemoryDedupe(256)))
-	addFn := appender.Add
 	if err != nil {
 		klog.Exit(err)
 	}
+	addFn := appender.Add
 
 	// Expose a HTTP handler for the conformance test writes.
 	// This should accept arbitrary bytes POSTed to /add, and return an ascii


lifecycle.go
```diff
@@ -59,47 +63,72 @@ type Appender struct {
...
+type NewCPFunc func(size uint64, hash []byte) ([]byte, error)
+
+// ParseCPFunc is the signature of a function which knows how to verify and parse checkpoints.
+type ParseCPFunc func(raw []byte) (*f_log.Checkpoint, error)
+
+// EntriesPathFunc is the signature of a function which knows how to format entry bundle paths.
+type EntriesPathFunc func(n uint64, p uint8) string
+
+// AppendOptions holds optional settings for all storage implementations.
+type AppendOptions struct {
```
mhutchinson, line 105: I'm not immediately convinced that all of these need to be publicly exposed from the API?

AlCutter: I've removed the func types for now, we can figure out whether/how to reintroduce in a less polluting way later on.

`AppendOptions` needs to be exported so that storage impls can accept it, but again I think we can return to this and see if there's a better way/place - the primary goal of _this_ PR is just to get us over the mark of broadly adopting the lifecycle API changes as it's blocking other work.

ct_only.go
```diff
@@ -63,8 +62,8 @@ func convertCTEntry(e *ctonly.Entry) *Entry {
 }
 
 // WithCTLayout instructs the underlying storage to use a Static CT API compatible scheme for layout.
-func WithCTLayout() func(*options.StorageOptions) {
-	return func(opts *options.StorageOptions) {
+func WithCTLayout() func(*AppendOptions) {
```
mhutchinson, line 65: A TODO would be ace here to indicate that this is known to be temporary. This option will need to be passed into every lifecycle state. As discussed in person, maybe moving this down to the drivers is a sensible strategy.

lifecycle.go
```diff
@@ -59,47 +62,61 @@ type Appender struct {
...
-		o.addDecorators = decorators
+func WithAppendDeduplication(decorators ...func(AddFn) AddFn) func(*AppendOptions) {
+	return func(o *AppendOptions) {
+		o.AddDecorators = decorators
 	}
 }
 
-type appendOptions struct {
-	addDecorators []func(AddFn) AddFn
+// AppendOptions holds optional settings for all storage implementations.
```
mhutchinson, line 90: Nit, not all of these are optional. At least, `NewCP` seems pretty required.

mhutchinson (approved):

PR #488, opened 2025-02-17
Defined witness policy configuration

mhutchinson:
This allows the required witnesses to be defined and the theshold
policies that apply within each group. Arbitrarily nested structures can
be built, each with different numbers of signatures.

Each WitnessGroup provides the URLs at which the witness can be reached
to perform witnessing, and a function that determines if the group is
satisfied.

This format is consistent with the only other known witness policy
configuration format out there:
https://git.glasklar.is/sigsum/core/sigsum-go/-/blob/main/doc/policy.md

Towards #309.

### DIFF
--- a/witness.go
+++ b/witness.go
@@ -31,10 +31,10 @@ type policyComponent interface {
 	// witnesses involved in this policy component.
 	Satisfied(cp []byte) bool
 
-	// Urls returns the URLs for requesting a counter signature from all
+	// URLs returns the URLs for requesting a counter signature from all
 	// witnesses that are involved in determining the satisfaction of this
 	// PolicyComponent.
-	Urls() []*url.URL
+	URLs() []*url.URL
 }
 
 // NewWitness returns a Witness given a verifier key and the root URL for where this
@@ -45,7 +45,7 @@ func NewWitness(vkey string, witnessRoot *url.URL) (Witness, error) {
 		return Witness{}, err
 	}
 	// "key hash" MUST be a lowercase hex-encoded SHA-256 hash of a 32-byte Ed25519 public key.
-	// This expression cuts off the identity name and hash
+	// This expression cuts off the identity name and hash.
 	key64 := strings.SplitAfterN(vkey, "+", 3)[2]
 	key, err := base64.StdEncoding.DecodeString(key64)
 	if err != nil {
@@ -62,7 +62,7 @@ func NewWitness(vkey string, witnessRoot *url.URL) (Witness, error) {
 }
 
 // Witness represents a single witness that can be reached in order to perform a witnessing operation.
-// The Urls() method returns the URL where it can be reached for witnessing, and the Satisfied method
+// The URLs() method returns the URL where it can be reached for witnessing, and the Satisfied method
 // provides a predicate to check whether this witness has signed a checkpoint.
 type Witness struct {
 	Key note.Verifier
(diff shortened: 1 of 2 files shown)


witness.go
```diff
@@ -0,0 +1,145 @@
...
+// single Witness, or a WitnessGroup.
+type policyComponent interface {
+	// Satisfied returns true if the checkpoint is signed by the quorum of
+	// witnesses involved in this policy component.
+	Satisfied(cp []byte) bool
+
+	// Urls returns the URLs for requesting a counter signature from all
+	// witnesses that are involved in determining the satisfaction of this
+	// PolicyComponent.
+	Urls() []*url.URL
```
roger2hk, line 37:
```suggestion
	URLs() []*url.URL
```

witness.go
```diff
@@ -0,0 +1,145 @@
...
+		Url: u,
+	}, err
+}
+
+// Witness represents a single witness that can be reached in order to perform a witnessing operation.
+// The Urls() method returns the URL where it can be reached for witnessing, and the Satisfied method
+// provides a predicate to check whether this witness has signed a checkpoint.
+type Witness struct {
+	Key note.Verifier
+	Url *url.URL
```
roger2hk, line 69:
```suggestion
	URL *url.URL
```

witness.go
```diff
@@ -0,0 +1,145 @@
...
+
+// NewWitness returns a Witness given a verifier key and the root URL for where this
+// witness can be reached.
+func NewWitness(vkey string, witnessRoot *url.URL) (Witness, error) {
+	v, err := note.NewVerifier(vkey)
+	if err != nil {
+		return Witness{}, err
+	}
+	// "key hash" MUST be a lowercase hex-encoded SHA-256 hash of a 32-byte Ed25519 public key.
+	// This expression cuts off the identity name and hash
```
roger2hk, line 48:
```suggestion
	// This expression cuts off the identity name and hash.
```

witness.go
```diff
@@ -0,0 +1,145 @@
...
+// Satisfied returns true if the checkpoint provided is signed by this witness.
+// This will return false if there is no signature, and also if the
+// checkpoint cannot be read as a valid note. It is up to the caller to ensure
+// that the input value represents a valid note.
+func (w Witness) Satisfied(cp []byte) bool {
+	n, err := note.Open(cp, note.VerifierList(w.Key))
+	if err != nil {
+		return false
+	}
+	return len(n.Sigs) == 1
```
roger2hk, line 81: Could the number of signature in the checkpoint be larger than `1` in this scenario?

roger2hk (approved):

witness.go
mhutchinson: D'oh, I had this _right_ at first, and then for some reason thought the Go policy was CamelCase for these things. I'm a muppet. Good catch.

witness.go
mhutchinson: No, because we only provide a single verifier in the verifier list.

roger2hk (approved):

witness.go
```diff
@@ -0,0 +1,145 @@
...
+	// "key hash" MUST be a lowercase hex-encoded SHA-256 hash of a 32-byte Ed25519 public key.
+	// This expression cuts off the identity name and hash.
+	key64 := strings.SplitAfterN(vkey, "+", 3)[2]
+	key, err := base64.StdEncoding.DecodeString(key64)
+	if err != nil {
+		return Witness{}, err
+	}
+	h := sha256.Sum256(key)
+
+	u := witnessRoot.JoinPath(fmt.Sprintf("/%x/add", h))
```
haydentherapper, line 56: Should this be `/add-checkpoint`, from https://github.com/C2SP/C2SP/blob/main/tlog-witness.md#add-checkpoint?

mhutchinson: Good spot. I'd already got this in the follow-up PR that I'm still working on, but I appreciate the attention to detail: https://github.com/transparency-dev/trillian-tessera/compare/main...mhutchinson:trillian-tessera:b309-witnessConfig?expand=1

PR #494, opened 2025-02-25
Implementing witnessing code and API

mhutchinson:
Towards #309.

### DIFF
--- a/internal/witness/witness.go
+++ b/internal/witness/witness.go
@@ -131,7 +131,6 @@ func (wg WitnessGateway) Witness(ctx context.Context, cp []byte) ([]byte, error)
 			err = errors.Join(err, r.err)
 			continue
 		}
-		fmt.Printf("got this sig: %q", r.sig)
 		// Add new signature to the new note we're building
 		sigBlock.Write(r.sig)
 
--- a/internal/witness/witness.go
+++ b/internal/witness/witness.go
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package storage
+// Package witness contains the implementation for sending out a checkpoint to witnesses
+// and retrieving sufficient signatures to satisfy a policy.
+package witness
 
 import (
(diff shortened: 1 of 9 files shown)


storage/internal/witness.go
```diff
@@ -0,0 +1,242 @@
...
+			statusCode: httpResp.StatusCode,
+			body:       rb,
+			headers:    httpResp.Header,
+		}, nil
+	}
+	for _, u := range urls {
+		witnesses = append(witnesses, witness{
+			url:        u,
+			size:       0,
+			post:       postFnImpl,
```
AlCutter, line 71: It kinda feels like this func should have the url baked in here, and the func signature be something like `func(context.Context, string) (...)`, or will the `witness` ever need to change `url` before passing it into `post`?

AlCutter (approved):

mhutchinson:
I've written tests for the main Gateway code. I'll take a look at the comments tomorrow to see if any of them can be addressed here without making this any bigger.

storage/internal/witness.go
mhutchinson: If this was an external API then I'd agree more strongly. I just tried changing it and the soft blocker is that if we lost the URL then the witness has no identifier, which makes any errors it returns ambiguous. Keeping the URL there allows us to say which witness failed.

internal/witness/witness.go
```diff
@@ -130,9 +131,9 @@ func (wg WitnessGateway) Witness(ctx context.Context, cp []byte) ([]byte, error)
 			err = errors.Join(err, r.err)
 			continue
 		}
+		fmt.Printf("got this sig: %q", r.sig)
```
AlCutter, line 134: Did you mean to leave this in?

AlCutter (approved):

internal/witness/witness.go
mhutchinson: Ummm... <snip> leave what in?

PR #495, opened 2025-02-25
Add Migration lifecycle support

AlCutter:
This PR adds `MigrationTarget` support to the lifecycle/driver API, and updates the POSIX and GCP implementations to match.

Towards #414 #471 #472

### DIFF
--- a/lifecycle.go
+++ b/lifecycle.go
@@ -201,9 +201,23 @@ func resolveAppendOptions(opts ...func(*AppendOptions)) *AppendOptions {
 	return defaults
 }
 
-type MigrationTarget struct {
-	SetBundle        func(ctx context.Context, idx uint64, bundle []byte) error
-	AwaitIntegration func(ctx context.Context, size uint64) ([]byte, error)
+// MigrationTarget describes the contract of the Migration lifecycle.
+//
+// This lifecycle mode is used to migrate C2SP tlog-tiles and static-ct
+// compliant logs into Tessera.
+type MigrationTarget interface {
+	// SetEntryBundle stores the provided serialised entry bundle at the location implied by the provided
+	// entry bundle index and partial size.
+	//
+	// Bundles may be set in any order (not just consecutively), and the implementation should integrate
+	// them into the local tree in the most efficient way possible.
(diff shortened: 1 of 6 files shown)


lifecycle.go
```diff
@@ -200,3 +200,50 @@ func resolveAppendOptions(opts ...func(*AppendOptions)) *AppendOptions {
...
+// MigrationTarget describes the contract of the Migration lifecycle.
+//
+// This lifecycle mode is used to migrate C2SP tlog-tiles and static-ct
+// compliant logs into Tessera.
+type MigrationTarget interface {
+	// SetEntryBundle stores the provided serialised entry bundle at the location implied by the provided
+	// entry bundle index and partial size.
+	//
+	// Bundles may be set in any order (not just consecutively), and the implementation should integrate
+	// them into the local tree in the most efficient way possible.
```
mhutchinson, line 213: What happens if the same bundle is set twice?

1) Overwrite
2) Silently fail
3) Noisily fail
4) compare and fail if not equal

There are probably other strategies. This feels important to be on the interface to make all implementations do the same thing.

lifecycle.go
```diff
@@ -200,3 +200,50 @@ func resolveAppendOptions(opts ...func(*AppendOptions)) *AppendOptions {
...
+	// and should return the locally calculated root hash derived from the integration of the contents of
+	// entry bundles set using SetEntryBundle above.
+	AwaitIntegration(ctx context.Context, size uint64) ([]byte, error)
+	// IntegratedSize returns the current size of the locally integrated log.
+	IntegratedSize(ctx context.Context) (uint64, error)
+}
+
+// BundleProcessorFunc is a function which knows how to turn a serialised entry bundle into a slice of
+// []byte representing each of the entries within the bundle.
+type BundleProcessorFunc func(entryBundle []byte) ([][]byte, error)
```
mhutchinson, line 225: `UnbundlerFunc` would be more in-theme.

mhutchinson (approved):

lifecycle.go
AlCutter: Added language to the comment.

Issue #497, opened 2025-02-26
[GCP] spanner transactions do not always retry due to unwrapped errors


bobcallaway:
**Description**

I was playing around with tessera and the local spanner emulator (which only supports one concurrent transaction at a time) and I got this fatal error:

`2025/02/23 16:49:56 failed to flush batch: failed to read SeqCoord: spanner: code = "Aborted", desc = "Transaction 251 aborted due to active transaction 250. The emulator only supports one transaction at a time.", requestID = "1.5119b3de91110e38.1.1.642.1"`

The [spanner API docs](https://pkg.go.dev/cloud.google.com/go/spanner#hdr-Aborted_transactions-ReadWriteTransaction) state that the transaction will be automatically retried when the `Aborted` code is returned, which isn't happening because in some situation the underlying error is being masked through string conversion instead of being wrapped: e.g.

https://github.com/transparency-dev/trillian-tessera/blob/4c294a9a09d440bc22d438814aee75149e17af36/storage/gcp/gcp.go#L773

PR #498, opened 2025-02-26
wrap errors from spanner calls inside txn

bobcallaway:
Fixes: #497 

I only wrapped the calls within a transaction that were calling spanner APIs so that those return codes would get bubbled up to the transaction layer... without a mock of spanner this will be hard to test.

### DIFF
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -770,11 +770,11 @@ func (s *spannerSequencer) assignEntries(ctx context.Context, entries []*tessera
 		// First we need to grab the next available sequence number from the SeqCoord table.
 		row, err := txn.ReadRowWithOptions(ctx, "SeqCoord", spanner.Key{0}, []string{"id", "next"}, &spanner.ReadOptions{LockHint: spannerpb.ReadRequest_LOCK_HINT_EXCLUSIVE})
 		if err != nil {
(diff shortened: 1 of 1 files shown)


storage/gcp/gcp.go
```diff
@@ -770,11 +770,11 @@ func (s *spannerSequencer) assignEntries(ctx context.Context, entries []*tessera
 		// First we need to grab the next available sequence number from the SeqCoord table.
 		row, err := txn.ReadRowWithOptions(ctx, "SeqCoord", spanner.Key{0}, []string{"id", "next"}, &spanner.ReadOptions{LockHint: spannerpb.ReadRequest_LOCK_HINT_EXCLUSIVE})
 		if err != nil {
-			return fmt.Errorf("failed to read SeqCoord: %v", err)
+			return fmt.Errorf("failed to read SeqCoord: %w", err)
```
AlCutter, line 773: I _suspect_ this is the only location where the `%w` will make any difference - I think all the others are going to be local ops and so can't really `Abort`.

bobcallaway: do you want this changed or is this just a comment? :)

AlCutter (approved):

AlCutter:
Ta!

Issue #469, opened 2025-02-04
Anti-spam support


AlCutter:
**Description**

Tessera should support a _best-effort_ deduplication mechanism which may optionally be used to help mitigate the effects of abuse.

Ideally, we'll have idiomatic/native drivers for the storage infrastructure Tessera currently supports, but also provide the ability for users of the Tessera library to implement their own dedupe drivers.

Strong/atomic/transactional deduplication is explicitly out of scope. We have previously found this to be expensive in cost & throughput, and would restrict the types of storage infrastructure Tessera could support. This goes against the philosophy of the Tessera approach.

patflynn:
I like this framing of the dedupe work because it helps me reason about what solution is good enough to address the spam risk.  I'd love to see some concrete spam risks from a small set of personality implementations to help understand what their common needs are.

@phbnf what do you think about using this framing and documenting spam risks for CT as the *parent* FR for CT dedupe?

PR #504, opened 2025-02-27
Add antispam support to migration lifecycle

AlCutter:
This PR adds an antispam option to the Migration lifecycle, and shows its use in the GCP migrate tool.

Towards #469

### DIFF
--- a/migrate_lifecycle.go
+++ b/migrate_lifecycle.go
@@ -90,3 +90,12 @@ func (o MigrationOptions) EntriesPath() func(uint64, uint8) string {
 func (o MigrationOptions) Followers() []func(context.Context, LogReader) {
 	return o.followers
 }
+
+func (o *MigrationOptions) WithAntispam(as Antispam) *MigrationOptions {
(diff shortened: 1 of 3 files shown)


migrate_lifecycle.go
```diff
@@ -90,3 +90,12 @@ func (o MigrationOptions) EntriesPath() func(uint64, uint8) string {
 func (o MigrationOptions) Followers() []func(context.Context, LogReader) {
 	return o.followers
 }
+
+func (o *MigrationOptions) WithAntispam(as Antispam) *MigrationOptions {
```
mhutchinson, line 94: This one would very much benefit from a brief comment to highlight that this doesn't prevent duplicate entries being added. This feature is intended to be used to pre-populate a DB for anti-spam, which likely only makes sense if you will transition this log to be an append-only log in the future. If you're strictly mirroring for ever and ever, then no need to use this.

mhutchinson (approved):

PR #505, opened 2025-02-27
[Hammer] support more verbose checkpoint logging

mhutchinson:
If the checkpoint size isn't updating, this logging will at least allow visibility that the main loop is still running.


roger2hk (approved):
Thanks, Martin!

PR #508, opened 2025-02-28
Antispam with lookup experiment

AlCutter:
This PR augments the appender future with a bool representing whether the entry was a dupe, and adapts the GCP conformance frontend to show how action can be taken on the FE in response to dupes - e.g. reading the original entry from the log in order to return extra information.


cmd/conformance/gcp/main.go
```diff
@@ -104,8 +110,39 @@ func main() {
...
+				w.WriteHeader(http.StatusInternalServerError)
+				_, _ = w.Write([]byte("invalid checkpoint"))
+				return
+			}
+			treeSize, err := strconv.ParseUint(bits[1], 10, 64)
+			if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
+				_, _ = w.Write([]byte(err.Error()))
+				return
+			}
```
mhutchinson, line 133: Consider doing this properly. This encourages people to do unsafe things with checkpoints.

append_lifecycle.go
```diff
@@ -54,7 +54,15 @@ type AddFn func(ctx context.Context, entry *Entry) IndexFuture
...
 // Implementations of this func are likely to be "futures", or a promise to return this data at
 // some point in the future, and as such will block when called if the data isn't yet available.
-type IndexFuture func() (uint64, error)
+type IndexFuture func() (Index, error)
+
+// Index represents a durably assigned index for some entry.
+type Index struct {
+	// Index is the location in the log to which a particular entry has been assigned.
+	Index uint64
+	// IsDup is true if Index represents a previously assignment for the same entry.
```
mhutchinson, line 63: "previously assignment" needs some rewriting. Could also do with some docs explaining that this will only be returned if AntiSpam is enabled etc. I still have some internal conflict about how dupe and anti-spam mean the same thing, but if you only know one term the other is hard to discover.

mhutchinson (approved):
I don't love this, but I can see why we need it.

phbnf (approved):

PR #509, opened 2025-03-03
Witnessing: Prevent the log from updating to an invalid size

haydentherapper:
This prevents a malicious witness from updating the log witness state to an invalid size that is larger than the log size.

### DIFF
--- a/internal/witness/witness.go
+++ b/internal/witness/witness.go
@@ -35,7 +35,7 @@ import (
 	"golang.org/x/mod/sumdb/note"
 )
 
-var PolicyNotSatisfiedErr = errors.New("witness policy was not satisfied")
+var ErrPolicyNotSatisfied = errors.New("witness policy was not satisfied")
 
 // NewWitnessGateway returns a WitnessGateway that will send out new checkpoints to witnesses
 // in the group, and will ensure that the policy is satisfied before returning. All outbound
@@ -143,7 +143,7 @@ func (wg *WitnessGateway) Witness(ctx context.Context, cp []byte) ([]byte, error
 	}
 
 	// We can only get here if all witnesses have returned and we're still not satisfied.
-	return sigBlock.Bytes(), errors.Join(PolicyNotSatisfiedErr, err)
+	return sigBlock.Bytes(), errors.Join(ErrPolicyNotSatisfied, err)
 }
 
 type consistencyFuture func() ([][]byte, error)
@@ -255,8 +255,12 @@ func (w *witness) update(ctx context.Context, cp []byte, size uint64, fetchProof
 			if err != nil {
 				return nil, fmt.Errorf("witness at %q replied with x.tlog.size but body %q could not be parsed as decimal", w.url, bodyStr)
 			}
+			// These cases should not happen unless the witness is misbehaving.
+			// w.size <= newWitSize <= size must always be true.
+			if newWitSize > size {
+				return nil, fmt.Errorf("witness at %q replied with x.tlog.size %d, larger than log size %d", w.url, newWitSize, size)
+			}
 			if newWitSize < w.size {
-				// This case should not happen unless the witness is misbehaving
 				return nil, fmt.Errorf("witness at %q replied with x.tlog.size %d, smaller than known size %d", w.url, newWitSize, w.size)
 			}
 			w.size = newWitSize


haydentherapper:
cc @mhutchinson

mhutchinson:
What's the context behind this PR? Have you deployed this and seen strange behaviour? 

The mechanism you're updating here is what allows a log to update its state of what size this process last observed from the witness. This size starts at zero each time Tessera starts, so we must be able to handle larger sizes. It must also be able to handle another Tessera instance having updated the witness since the last time this process did it, so zero is not a special case.

I can't see any real situation where the witness can have replied with a larger size and now has a smaller size. A witness rewinding its view of the log size is misbehaving. What's the situation you have seen?

haydentherapper:
Ah, I had a different use case in mind. I’ve been playing around with the witness code you’ve added for BT, and simulated a log updating only a subset of witnesses. Then, when the log attempts to update a different subset of witnesses with a later checkpoint, the consistency proof verification will fail since the log and witness are out of sync and the witness has an older view of the log. It then replies with its last verified size, which is less than the current size of the log, and the log can then compute a consistency proof between that older size and its current size.

haydentherapper:
It sounds like there’s two cases we have to handle: one where the log process just started up and is 0, where as you noted the witness will always have a larger size, and another case where the witness has fallen behind where the log is at, which would be a smaller size than the log size.

mhutchinson:
Let's introduce some terminology because there's some subtlety that is hard to communicate without new symbols. It reminds me of mind games where one ends up with sentences like "I know X, and he I know that she knows X, but does she know that I know X?!". I don't love this terminology, but the important thing is that the first letter is always the actor with the state:

- Ls: The latest size of the log `L`
- WLs: The size of the latest checkpoint signed by a witness `W` for log `L` (i.e. the witnesses view of L)
- LWs: The size that log `L` thinks witness `W` holds for it (i.e. the logs view of the witnesses' view of L)

The invariant here is `Ls >= WLs >= LWs`. In an ecosystem where the log hasn't updated in some time, then all of these should become equal.

 The relationship between Ls and WLs is clear: if the witness has signed a larger checkpoint than the log now claims, then the log is not behaving in an append-only manner
 
The relationship between WLs and LWs is almost as clear; if LWs is only updated in response to W responding with a new WLs then there should be no situation where LWs > WLs. There is one caveat to that, which is that a witness responding with `text/x.tlog.size` can cause the log to update its view, and this response is not authenticated. This means that a log that talked to the wrong witness, or an intentional bad actor, could cause the log to update its state to be ahead of what is possible.

I've got a suggestion for your PR that I'll add as a comment on the code that might make it a useful extra check.

internal/witness/witness.go
```diff
@@ -224,9 +224,9 @@ func (w witness) update(ctx context.Context, cp []byte, size uint64) ([]byte, er
 			if err != nil {
 				return nil, fmt.Errorf("witness at %q replied with x.tlog.size but body %q could not be parsed as decimal", w.url, bodyStr)
 			}
-			if newWitSize < int64(w.size) {
+			if newWitSize > int64(w.size) {
```
mhutchinson, line 227:
```suggestion
			if newWitSize > int64(size) {
```

If you duplicated this block (i.e. leave the original if block unchanged) and then made this change to compare agains the size we just sent out (which should be the largest possible size for this log) then it would at least prevent one category of errors where a malicious actor could MitM the connection and respond with MaxUint64.

mhutchinson:
See also #510. I suspect if you rebase this PR as it stands right now onto that, then the new test will fail.

haydentherapper:
@mhutchinson, thanks for chatting through this. In playing around with the WitnessGateway, I set `w.size` to the latest checkpoint size (`Ls` from your comment), which I see was misunderstanding the relationship between the various sizes. I made the suggested change and added a clarifying comment.

Also fixed a minor linter complaint.

mhutchinson (approved):

mhutchinson:
I tried to get this in today but CodeQL is angry. I've made #529 that should make this trivial to fix if you rebase. I can merge this on Monday, or maybe you can get @AlCutter or @roger2hk to get this in tomorrow before any more obstacles come along :-)

haydentherapper:
@mhutchinson rebased

mhutchinson (approved):

PR #510, opened 2025-03-03
Fix witnessed size state

mhutchinson:
Previously the state was being lost. Even though this was unintentional, it worked, but involved constantly querying the witness as if it was at size 0.

This change makes the size sticky for the life of the process by correctly using pointers instead of copying structs. It also updates the size on success, which was missing earlier.

### DIFF
--- a/internal/witness/witness_test.go
+++ b/internal/witness/witness_test.go
@@ -368,7 +368,7 @@ func TestWitnessStateEvolution(t *testing.T) {
 			if !bytes.HasPrefix(body, []byte("old 34840403")) {
 				t.Fatalf("expected body to start with old 34840403 but got\n%v", string(body))
 			}
-			// End of test; we don't even both constructing a valid response here
+			// End of test; we don't even bother constructing a valid response here
 		}
 		count++
 	}))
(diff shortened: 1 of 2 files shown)


internal/witness/witness_test.go
```diff
@@ -324,6 +325,88 @@ func TestWitness_UpdateResponse(t *testing.T) {
...
+			_, _ = w.Write(sigForSigner(t, cp, wit1_skey))
+		case 2:
+			body, err := io.ReadAll(r.Body)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if !bytes.HasPrefix(body, []byte("old 34840403")) {
+				t.Fatalf("expected body to start with old 34840403 but got\n%v", string(body))
+			}
+			// End of test; we don't even both constructing a valid response here
```
AlCutter, line 371:
```suggestion
			// End of test; we don't even bother constructing a valid response here
```

internal/witness/witness_test.go
```diff
@@ -324,6 +325,88 @@ func TestWitness_UpdateResponse(t *testing.T) {
...
+
+			_, _ = w.Write(sigForSigner(t, cp, wit1_skey))
+		case 2:
+			body, err := io.ReadAll(r.Body)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if !bytes.HasPrefix(body, []byte("old 34840403")) {
+				t.Fatalf("expected body to start with old 34840403 but got\n%v", string(body))
+			}
```
AlCutter, line 370: Do we hit here? I can see two calls to `g.Witness` below which I imagine would hit `count==0` and then `count==`?

internal/witness/witness_test.go
```diff
@@ -324,6 +325,88 @@ func TestWitness_UpdateResponse(t *testing.T) {
...
+		return nil
+	}
+
+	g := NewWitnessGateway(group, ts.Client(), fetchProof)
+	_, err = g.Witness(ctx, logSignedCheckpoint)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, _ = g.Witness(ctx, logSignedCheckpoint)
```
AlCutter, line 398: Could do with a comment as to why it's not important to check the error here

AlCutter (approved):

PR #514, opened 2025-03-06
[Witness] construct proofs in witness

mhutchinson:
This switches away from the previous idea which was to pass in a proof
fetcher. Now a tile fetcher is passed in, and the proofs are constructed
in the witness code.

Now that real proofs are being generated, we need to use a real log in
the witness tests. The POSIX log in testdata is used for this purpose.

The unsafe method for parsing checkpoints has also been extended to
return hashes, which was needed to support this work.

Towards #309.

### DIFF
--- a/internal/witness/witness.go
+++ b/internal/witness/witness.go
@@ -222,7 +222,7 @@ func (w *witness) update(ctx context.Context, cp []byte, size uint64, fetchProof
 	}
 	rb, err := io.ReadAll(httpResp.Body)
 	if err != nil {
-		return nil, fmt.Errorf("failed to ready body from witness at %q: %v", w.url, err)
+		return nil, fmt.Errorf("failed to read body from witness at %q: %v", w.url, err)
 	}
 	_ = httpResp.Body.Close()
 
--- a/internal/witness/witness.go
+++ b/internal/witness/witness.go
@@ -92,6 +92,7 @@ func (wg *WitnessGateway) Witness(ctx context.Context, cp []byte) ([]byte, error
 	}
 	pf := sharedConsistencyProofFetcher{
 		pb:      pb,
+		toSize:  size,
 		results: make(map[uint64]consistencyFuture),
 	}
 
@@ -152,12 +153,16 @@ type consistencyFuture func() ([][]byte, error)
 // of the witnesses are of the same size, and thus require the same proof.
 type sharedConsistencyProofFetcher struct {
 	pb      *client.ProofBuilder
+	toSize  uint64
 	mu      sync.Mutex
 	results map[uint64]consistencyFuture
 }
 
-// ConsistencyProof fetches a consistency proof, reusing any results from parallel requests.
+// ConsistencyProof constructs a consistency proof, reusing any results from parallel requests.
 func (pf *sharedConsistencyProofFetcher) ConsistencyProof(ctx context.Context, smaller, larger uint64) ([][]byte, error) {
+	if larger != pf.toSize {
+		return nil, fmt.Errorf("required larger size to be %d but was given %d", pf.toSize, larger)
+	}
 	var f consistencyFuture
 	var ok bool
 	pf.mu.Lock()
@@ -209,7 +214,7 @@ func (w *witness) update(ctx context.Context, cp []byte, size uint64, fetchProof
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, w.url, strings.NewReader(body))
 	if err != nil {
-		return nil, fmt.Errorf("failed to post to witness at %q: %v", w.url, err)
+		return nil, fmt.Errorf("failed to construct request to %q: %v", w.url, err)
 	}
 	httpResp, err := w.client.Do(req)
 	if err != nil {
@@ -217,7 +222,7 @@ func (w *witness) update(ctx context.Context, cp []byte, size uint64, fetchProof
 	}
 	rb, err := io.ReadAll(httpResp.Body)
 	if err != nil {
-		return nil, fmt.Errorf("failed to post to witness at %q: %v", w.url, err)
+		return nil, fmt.Errorf("failed to ready body from witness at %q: %v", w.url, err)
 	}
 	_ = httpResp.Body.Close()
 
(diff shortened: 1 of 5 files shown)


internal/witness/witness.go
```diff
@@ -152,16 +145,31 @@ func (wg *WitnessGateway) Witness(ctx context.Context, cp []byte) ([]byte, error
...
+// of the witnesses are of the same size, and thus require the same proof.
+type sharedConsistencyProofFetcher struct {
+	pb      *client.ProofBuilder
+	mu      sync.Mutex
+	results map[uint64]consistencyFuture
 }
 
-// postFn wraps http.Post in order to allow for easier testing.
-type postFn func(ctx context.Context, url, body string) (pr postResponse, err error)
+// ConsistencyProof fetches a consistency proof, reusing any results from parallel requests.
```
AlCutter, line 159:
```suggestion
// ConsistencyProof constructs a consistency proof, reusing any results from parallel requests.
```

internal/witness/witness.go
```diff
@@ -152,16 +145,31 @@ func (wg *WitnessGateway) Witness(ctx context.Context, cp []byte) ([]byte, error
...
+func (pf *sharedConsistencyProofFetcher) ConsistencyProof(ctx context.Context, smaller, larger uint64) ([][]byte, error) {
+	var f consistencyFuture
+	var ok bool
+	pf.mu.Lock()
+	if f, ok = pf.results[smaller]; !ok {
+		f = sync.OnceValues(func() ([][]byte, error) {
+			return pf.pb.ConsistencyProof(ctx, smaller, larger)
+		})
+		pf.results[smaller] = f
+	}
```
AlCutter, line 169: This keying of the cache will obviously break if anyone _somehow_ called this method with a different sized `larger` than all the other callers.

I suspect that at the moment this probably can't happen(?), but I wonder if it's worth requiring `larger` be part of the `sharedConsistencyProofFetcher` and validated with each call to `ConsistencyProof`, or widening the key to be a string like e.g. `"<smaller>:<larger>"` to defend against manually-effected baldness for future us if the surrounding code materially changes?

internal/witness/witness.go
```diff
@@ -196,16 +207,30 @@ func (w *witness) update(ctx context.Context, cp []byte, size uint64) ([]byte, e
...
+		return nil, fmt.Errorf("failed to post to witness at %q: %v", w.url, err)
+	}
+	httpResp, err := w.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to post to witness at %q: %v", w.url, err)
+	}
+	rb, err := io.ReadAll(httpResp.Body)
 	if err != nil {
 		return nil, fmt.Errorf("failed to post to witness at %q: %v", w.url, err)
 	}
```
AlCutter, line 221: Worth disambiguating each of these error conditions in the msgs?

internal/witness/witness.go
```diff
@@ -196,16 +207,30 @@ func (w *witness) update(ctx context.Context, cp []byte, size uint64) ([]byte, e
...
-		signed := append(cp, resp.body...)
+		// Concatenate the signature to the checkpoint passed in and verify it is valid.
+		// append is tempting here but is dangerous because it can modify `cp` and race with other
+		// witnesses, causing signatures to be swapped. cp must not be modified.
+		signed := make([]byte, len(cp)+len(rb))
+		copy(signed, cp)
+		copy(signed[len(cp):], rb)
 		if n, err := note.Open(signed, note.VerifierList(w.verifier)); err != nil {
-			return nil, fmt.Errorf("witness at %q replied with invalid signature: %q", w.url, resp.body)
+			return nil, fmt.Errorf("witness %q at %q replied with invalid signature: %q\nconstructed note: %q\n%v", w.verifier.Name(), w.url, rb, string(signed), err)
```
AlCutter, line 233: nit: might be easy to miss the error message in the log - worth modifying to `...note: %q\nerror: %v`?

internal/witness/witness_test.go
```diff
@@ -22,34 +22,34 @@ import (
...
 const (
-	log_vkey    = "LilLog+b65b2501+Af40T+NgLuQzeKqU1mbUL4pcmQwCVDK67QSmdJ3Q8LTl"
-	log_skey    = "PRIVATE+KEY+LilLog+b65b2501+AbKaNq0e8nx6WOuOH0eYAgPeKPtk8KM3fZBhwr5qzo+p"
+	log_vkey    = "example.com/log/testdata+33d7b496+AeHTu4Q3hEIMHNqc6fASMsq3rKNx280NI+oO5xCFkkSx"
 	wit1_vkey   = "Wit1+55ee4561+AVhZSmQj9+SoL+p/nN0Hh76xXmF7QcHfytUrI1XfSClk"
 	wit1_skey   = "PRIVATE+KEY+Wit1+55ee4561+AeadRiG7XM4XiieCHzD8lxysXMwcViy5nYsoXURWGrlE"
 	wit2_vkey   = "Wit2+85ecc407+AWVbwFJte9wMQIPSnEnj4KibeO6vSIOEDUTDp3o63c2x"
 	wit2_skey   = "PRIVATE+KEY+Wit2+85ecc407+AfPTvxw5eUcqSgivo2vaiC7JPOMUZ/9baHPSDrWqgdGm"
 	witBad_vkey = "WitBad+b82b4b16+AY5FLOcqxs5lD+OpC6cVTrxsyNJktaCGYHNfnE5vKBQX"
 	witBad_skey = "PRIVATE+KEY+WitBad+b82b4b16+AYSil2PKfSN1a0LhdbzmK1uXqDFZbp+P1OyR54k3gdJY"
```
AlCutter, line 44: Maybe for a follow-up, but just spotted the naming convention for the `consts` here is a bit unusual (https://google.github.io/styleguide/go/decisions.html#constant-names)

AlCutter (approved):
Makes sense given the context, just a new suggestions/nits in-line.

internal/witness/witness.go
mhutchinson: Done.

internal/witness/witness_test.go
mhutchinson: Fixed

internal/witness/witness.go
AlCutter: You won't get a line number unless there's a `panic` - you'll only see the line number of the place where this (probably, by then, wrapped multiple times) error is ultimately passed into `log.Infof`.

internal/witness/witness.go
```diff
@@ -209,15 +214,15 @@ func (w *witness) update(ctx context.Context, cp []byte, size uint64, fetchProof
...
+		return nil, fmt.Errorf("failed to construct request to %q: %v", w.url, err)
 	}
 	httpResp, err := w.client.Do(req)
 	if err != nil {
 		return nil, fmt.Errorf("failed to post to witness at %q: %v", w.url, err)
 	}
 	rb, err := io.ReadAll(httpResp.Body)
 	if err != nil {
-		return nil, fmt.Errorf("failed to post to witness at %q: %v", w.url, err)
+		return nil, fmt.Errorf("failed to ready body from witness at %q: %v", w.url, err)
```
AlCutter, line 225: nit: /ready/read/

AlCutter (approved):
one tiny typo :)

AlCutter (approved):

PR #515, opened 2025-03-06
[Witnessing] Use witnessing in all drivers

mhutchinson:
All drivers now take set up witnessing as configured in the AppendOptions. If no options are provided then this is still set up, but does nothing. Towards #309.  This required changing the witness tests package to avoid a package cycle; the posix implementation now uses the witness, but the witness tests use posix.

This is based on #514, and that needs to be merged first.

### DIFF
--- a/internal/witness/witness_test.go
+++ b/internal/witness/witness_test.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package witness
+package witness_test
 
 import (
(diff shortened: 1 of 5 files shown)


AlCutter (approved):
LGTM, it'd be nice to see if we can move this up to the options/tessera layer later on so storage impls don't need to care other than calling `newCP`, but I see that it's tricky to get there at the mo'.

Worth adding an issue or TODO to keep track?

Issue #470, opened 2025-02-04
[GCP] Anti-spam support


AlCutter:
**Description**

Native anti-spam support for GCP, likely using Spanner.

PR #521, opened 2025-03-11
[GCP] Add persistent anti-spam support to terraform

AlCutter:
This PR adds support for:
 - the GCP terraform to optionally create infrastructure used by the GCP antispam implementation.
 - the conformance terraform to optionally create and use the GCP antispam support.

As a side-effect, the naming scheme for the Spanner databases has changed very slightly.
 
Towards #470

### DIFF
--- a/deployment/modules/gcp/gcs/main.tf
+++ b/deployment/modules/gcp/gcs/main.tf
@@ -62,7 +62,7 @@ resource "google_spanner_instance" "log_spanner" {
 
 resource "google_spanner_database" "log_db" {
   instance = google_spanner_instance.log_spanner.name
-  name     = "${var.base_name}-db"
+  name     = "${var.base_name}"
 
   deletion_protection = !var.ephemeral
 }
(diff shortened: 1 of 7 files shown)


deployment/modules/gcp/gcs/variables.tf
```diff
@@ -29,7 +29,12 @@ variable "log_writer_members" {
   type        = list(any)
 }
 
+variable "create_antispam" {
+  description = "Set to true to create the infrastructure required by the GCP antispam implementation"
```
roger2hk, line 33:
```suggestion
  description = "Set to true to create the infrastructure required by the GCP antispam implementation."
```

deployment/modules/gcp/gcs/main.tf
```diff
@@ -67,10 +67,27 @@ resource "google_spanner_database" "log_db" {
   deletion_protection = !var.ephemeral
 }
 
+resource "google_spanner_database" "log_antispam_db" {
+  count    = var.create_antispam ? 1 : 0
+  instance = google_spanner_instance.log_spanner.name
+  name     = "${var.base_name}-db_dedup"
```
roger2hk, line 73: Should the `dedup` in the database name be replaced with `antispam`? I find it a bit weird to have both `-` and `_` in the name.

```suggestion
  name     = "${var.base_name}-antispam-db"
```

roger2hk (approved):

mhutchinson (approved):

deployment/modules/gcp/gcs/main.tf
AlCutter: Yeah, you're not wrong, let me see how much breakage that's going to cause :)

PR #523, opened 2025-03-12
Add CT support for migration

AlCutter:
This PR adds necessary hooks and implementations for TesseraCT to implement a CT-specific tool for migrating [static-ct](https://c2sp.org/static-ct) compliant logs.

Towards #495

### DIFF
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -108,6 +108,7 @@ func NewAppendOptions() *AppendOptions {
 		batchMaxSize:           DefaultBatchMaxSize,
 		batchMaxAge:            DefaultBatchMaxAge,
 		entriesPath:            layout.EntriesPath,
+		bundleIDHasher:         defaultIDHasher,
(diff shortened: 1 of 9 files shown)


append_lifecycle.go
```diff
@@ -108,6 +108,7 @@ func NewAppendOptions() *AppendOptions {
 		batchMaxSize:           DefaultBatchMaxSize,
 		batchMaxAge:            DefaultBatchMaxAge,
 		entriesPath:            layout.EntriesPath,
+		bundleIDHasher:         defaultIDHasher,
```
roger2hk, line 111: I couldn't leave a comment on L100. Should the `defaultIDHasher` in `WithAntispam()` be updated to `o. bundleIDHasher`?

AlCutter: Yes - great spot, thanks!

roger2hk (approved):

Issue #341, opened 2024-11-27
Provide an API for integration done so personalities know when they can shutdown


mhutchinson:
At the moment personalities have no way of knowing whether there is still important background work happening to integrate sequenced entries into the log. For AWS/GCP, this means that a personality could accept a bunch of writes, sequence them, and then shut down. These sequenced entries would not be available via the read API until a storage instance is created again, but the personality has no way of knowing when this work is complete if it didn't keep its own record of the last issued sequence number.

Options:
 - Storage has a `Shutdown` method (prior art: https://pkg.go.dev/net/http#Server.Shutdown).
 - We already pass a `Context` into `storage.New` - we could cancel this and also provide a `Run` method or similar in the storage that exits after this context cancellation, but only after integration has happened
 - When exiting the personality instantiates a draining storage (lifecycle doc: https://github.com/transparency-dev/trillian-tessera/blob/main/docs/design/lifecycle.md#draining)

mhutchinson:
Looks like I accidentally closed this. The main feature still needs to be added.

PR #524, opened 2025-03-12
Implement Shutdown method on Appender

mhutchinson:
Towards #341.

This API allows users of an appender to ensure that everything they've sent to this object is committed to before the process terminates.

Out of scope for this change is to make it so that it globally waits for items added via any other appenders to become integrated. The plan for this is to implement a draining lifecycle state, which can better ensure that the queue stays empty.

Raising as a draft because it's late in the day and I'd like to have another look tomorrow morning to make sure this isn't bonkers. There's a bit of duplication here that I'd like to make sure is reasonable. OTOH, raising the PR now so that the tests can run and it can be looked at in principle if reviewers get the time.

### DIFF
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -191,8 +191,8 @@ func (s *Storage) Appender(ctx context.Context, opts *tessera.AppendOptions) (*t
 	}
 
 	go a.sequencerJob(ctx)
-	ctx, cancel := context.WithCancel(ctx)
-	a.done = ctx.Done()
+	cctx, cancel := context.WithCancel(ctx)
+	a.done = cctx.Done()
 	go a.publisherJob(ctx, cancel, opts.CheckpointInterval())
 
 	reader := &LogReader{
@@ -278,8 +278,8 @@ func (a *Appender) Shutdown(ctx context.Context) error {
 	if err := a.queue.Close(ctx); err != nil {
 		return err
 	}
-	// At this point sequencing and integration is done, now to make sure a large
-	// enough checkpoint has been issued.
+	// At this point sequencing is done, now to block until a large enough tree
+	// is constructed and a checkpoint has been issued.
 	maxIndex := a.largestIssued.Load()
 	if maxIndex == 0 {
 		// special case no work done
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -287,28 +287,32 @@ func (a *Appender) Shutdown(ctx context.Context) error {
 		<-a.done
 		return nil
 	}
+	sleepTime := 0 * time.Millisecond
 	for {
-		cp, err := a.logStore.getCheckpoint(ctx)
-		if err != nil && !errors.Is(err, gcs.ErrObjectNotExist) {
-			return err
-		}
-		if err == nil {
-			_, size, _, err := parse.CheckpointUnsafe(cp)
-			if err != nil {
-				return err
-			}
-			klog.V(1).Infof("Shutting down, waiting for checkpoint committing to size %d (current checkpoint is %d)", maxIndex, size)
-			if size > maxIndex {
-				close(a.cpUpdated)
-				<-a.done
-				return nil
-			}
-		}
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
 		default:
-			time.Sleep(100 * time.Millisecond)
+			time.Sleep(sleepTime)
+		}
+		sleepTime = 100 * time.Millisecond // after the first time, ensure we sleep in any other loops
+
+		cp, err := a.logStore.getCheckpoint(ctx)
+		if err != nil {
+			if !errors.Is(err, gcs.ErrObjectNotExist) {
+				return err
+			}
+			continue
+		}
+		_, size, _, err := parse.CheckpointUnsafe(cp)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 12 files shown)


storage/gcp/gcp.go
```diff
@@ -221,11 +225,92 @@ type Appender struct {
...
+//
+// After this returns, any calls to Add will fail.
+func (a *Appender) Shutdown(ctx context.Context) error {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	a.stopped = true
+	if err := a.queue.Close(ctx); err != nil {
+		return err
+	}
+	// At this point sequencing and integration is done, now to make sure a large
```
AlCutter, line 281: Is this comment correct?
I think that anything already in the `Queue` will have had sequence numbers durably assigned, and nothing else can be added, but integration may not have happened yet?

storage/gcp/gcp.go
```diff
@@ -221,11 +225,92 @@ type Appender struct {
...
+	// enough checkpoint has been issued.
+	maxIndex := a.largestIssued.Load()
+	if maxIndex == 0 {
+		// special case no work done
+		close(a.cpUpdated)
+		<-a.done
```
AlCutter, line 287: You're racing the `publisherTask` goroutine in taking from the `cpUpdated` channel here, so the special case could occur over there (I believe the Go runtime can do context switching to another goroutine on function calls).

I think the worst that could happen if you just let it handle it there is that it'll publish one more CP and then exit due to the closed `cpUpdated` channel?

I imagine the same in the other storage impls.

storage/gcp/gcp.go
```diff
@@ -221,11 +225,92 @@ type Appender struct {
...
+		return nil
+	}
+	for {
+		cp, err := a.logStore.getCheckpoint(ctx)
+		if err != nil && !errors.Is(err, gcs.ErrObjectNotExist) {
+			return err
+		}
+		if err == nil {
```
AlCutter, line 295: The indented block for `err == nil` is a bit unusual, could avoid it with something like this here:

```suggestion
		if err != nil {
		  if !errors.Is(err, gcs.ErrObjectNotExist) {
			return err
		  }
		}
		
```

storage/gcp/gcp.go
```diff
@@ -221,11 +225,92 @@ type Appender struct {
...
+			_, size, _, err := parse.CheckpointUnsafe(cp)
+			if err != nil {
+				return err
+			}
+			klog.V(1).Infof("Shutting down, waiting for checkpoint committing to size %d (current checkpoint is %d)", maxIndex, size)
+			if size > maxIndex {
+				close(a.cpUpdated)
+				<-a.done
```
AlCutter, line 303: Racy here too

storage/internal/queue.go
```diff
@@ -54,23 +63,24 @@ type FlushFunc func(ctx context.Context, entries []*tessera.Entry) error
 // the same order as they were added, when either the oldest entry in the queue has been there
 // for maxAge, or the size of the queue reaches maxSize.
 func NewQueue(ctx context.Context, maxAge time.Duration, maxSize uint, f FlushFunc) *Queue {
+	ctx, cancel := context.WithCancel(ctx)
```
AlCutter, line 66: You're shadowing the passed-in ctx here, which means the context-with-cancel is passed in to the worker thread on L107 rather than the outer context, is that ok?

Maybe rename to `cctx`?

storage/internal/queue.go
```diff
@@ -103,6 +126,30 @@ func (q *Queue) Add(ctx context.Context, e *tessera.Entry) tessera.IndexFuture {
...
+func (q *Queue) Close(ctx context.Context) error {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+	q.closed = true
+
+	// Flush and close the buffer, which will trigger workers
+	// to process the entries.
+	if err := q.buf.Flush(); err != nil {
+		return err
+	}
```
AlCutter, line 140: I think you can drop the call to `Flush` as `Close` will flush before marking the buffer closed: https://pkg.go.dev/github.com/globocom/go-buffer#Buffer.Close

AlCutter (commented):
I added some comments on the `aws` storage impl, I think they'll apply to the other impls too so I figured I'd stop here and let you ponder those before I continue.

storage/gcp/gcp.go
mhutchinson: Ditto my response from above then. I'll leave the comment open in case remediation is needed on this race and I'm missing it.

storage/internal/queue.go
mhutchinson: Yeah I think this is safer. Done.

storage/gcp/gcp.go
mhutchinson: Done.

storage/gcp/gcp.go
mhutchinson: There's a comment on the channel field that we pull from. Discussed 1:1 and you are happy that's reasonable.

AlCutter (approved):

storage/gcp/gcp.go
mhutchinson: I've convinced myself that this works, but the comment was wrong. I've updated the comment.

mhutchinson:
Replaced this with #525 which is more contained. Closing this one. It can stay about for historical purposes if we need to do any invasive closing of channels.

PR #525, opened 2025-03-13
[Shutdown] Implement Shutdown in a single location

### LINKED ISSUE #341: Provide an API for integration done so personalities know when they can shutdown

mhutchinson:
This is a revised attempt at Shutdown that is intended to replace #524.
This implementation doesn't have the granular integration with the
channels in each implementation. Instead, it is implemented at the top
level of the appender lifecycle, and each driver doesn't need to have
very similar duplicated code for handling this.

The previous attempt was also trying to make it so that Shutdown would
close out channels and kill goroutines etc. This approach makes it clear
that the way to do this is to cancel the original context passed into
the appender.

Fixes #341.

### DIFF
--- a/cmd/examples/posix-oneshot/main.go
+++ b/cmd/examples/posix-oneshot/main.go
@@ -115,13 +115,15 @@ func main() {
 	for _, entry := range indexFutures {
 		seq, _, err := await.Await(ctx, entry.f)
 		if err != nil {
-			klog.Exitf("failed to sequence %q: %q", entry.name, err)
+			klog.Exitf("Failed to sequence %q: %q", entry.name, err)
 		}
 		klog.Infof("%d: %v", seq, entry.name)
 	}
 
 	// 2) shutdown the appender
-	shutdown(ctx)
+	if err := shutdown(ctx); err != nil {
+		klog.Exitf("Failed to shut down cleanly: %v", err)
+	}
 }
 
 // Read log private key from file or environment variable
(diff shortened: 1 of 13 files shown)


cmd/examples/posix-oneshot/main.go
```diff
@@ -119,7 +120,8 @@ func main() {
 		klog.Infof("%d: %v", seq, entry.name)
 	}
 
-	// All futures have been resolved, which means the log is built and we can allow the process to terminate. Goodbye!
+	// 2) shutdown the appender
+	shutdown(ctx)
```
AlCutter, line 124: check/log err returned?

AlCutter (approved):

cmd/examples/posix-oneshot/main.go
mhutchinson: I thought it looked a bit bare. Done.

PR #542, opened 2025-03-18
Move migration into lifecycle

AlCutter:
This PR moves the migration functionality into the MigrationTarget lifecycle, rather than in a separate function.


storage/gcp/gcp.go
```diff
@@ -1029,7 +1029,7 @@ func (s *gcsStorage) lastModified(ctx context.Context, obj string) (time.Time, e
 }
 
 // MigrationTarget creates a new GCP storage for the MigrationTarget lifecycle mode.
-func (s *Storage) MigrationTarget(ctx context.Context, opts *tessera.MigrationOptions) (tessera.MigrationTarget, tessera.LogReader, error) {
+func (s *Storage) MigrationTarget(ctx context.Context, opts *tessera.MigrationOptions) (tessera.MigrationWriter, tessera.LogReader, error) {
```
mhutchinson, line 1032: Looks like the method name needs updating, but maybe you left that to avoid catching me out.

AlCutter: Done

mhutchinson (approved):

PR #544, opened 2025-03-19
Add richer Follower interface enabling clean shutdown & stats

AlCutter:
This PR defines a `Follower` interface which enables Tessera to know when it's safe to exit, e.g. during migration, we need to potentially wait for the antispam index to be fully populated.

Towards #436

### DIFF
--- a/migrate_lifecycle.go
+++ b/migrate_lifecycle.go
@@ -63,7 +63,7 @@ func NewMigrationTarget(ctx context.Context, d Driver, opts *MigrationOptions) (
 		return nil, fmt.Errorf("failed to init MigrationTarget lifecycle: %v", err)
 	}
 	for _, f := range opts.followers {
-		go f(ctx, r)
+		go f.Follow(ctx, r)
 	}
 	return &MigrationTarget{
 		writer: mw,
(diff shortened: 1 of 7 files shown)


migrate_lifecycle.go
```diff
@@ -108,16 +104,16 @@ func (o *MigrationOptions) LeafHasher() func([]byte) ([][]byte, error) {
 // of the source tree and so no attempt is made to reject/deduplicate entries.
 func (o *MigrationOptions) WithAntispam(as Antispam) *MigrationOptions {
 	if as != nil {
-		o.followers = append(o.followers, func(ctx context.Context, lr LogReader) {
-			as.Populate(ctx, lr, o.bundleIDHasher)
-		})
+		o.followers = append(o.followers, as.Follower(o.bundleIDHasher))
```
mhutchinson, line 107: Can you add a comment to `bundleIDHasher` that this must always be immutable after struct initialization? An easy trap now would be to add a feature to allow that to become mutable, which would make this `WithAntispam` method behave differently depending on order of calls.

mhutchinson (approved):

mhutchinson (approved):
Consider adding tests for this.

AlCutter:
> Consider adding tests for this.

Yup, I'll come back with a follow-up PR so as not to block this one.

PR #555, opened 2025-03-25
Standardize logging when publishing a checkpoint

haydentherapper:
All storage backends will log when a new checkpoint is published.

Tiny change to standardize logging message for integration as well.

### DIFF
--- a/storage/posix/files.go
+++ b/storage/posix/files.go
@@ -343,7 +343,7 @@ func doIntegrate(ctx context.Context, fromSeq uint64, leafHashes [][]byte, ls *l
 		}
 	}
 
-	klog.Infof("New tree state: %d, %x", newSize, newRoot)
+	klog.Infof("New tree: %d, %x", newSize, newRoot)
 
 	return newSize, newRoot, nil
 }
@@ -591,7 +591,8 @@ func (a *appender) publishCheckpoint(minStaleness time.Duration) error {
 	if err := a.s.overwrite(layout.CheckpointPath, cpRaw); err != nil {
 		return fmt.Errorf("overwrite(%s): %v", layout.CheckpointPath, err)
 	}
-	klog.Infof("Published latest checkpoint")
+
+	klog.V(2).Infof("Published latest checkpoint: %d, %x", size, root)
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


storage/posix/files.go
```diff
@@ -591,7 +591,6 @@ func (a *appender) publishCheckpoint(minStaleness time.Duration) error {
 	if err := a.s.overwrite(layout.CheckpointPath, cpRaw); err != nil {
 		return fmt.Errorf("overwrite(%s): %v", layout.CheckpointPath, err)
 	}
-	klog.Infof("Published latest checkpoint")
```
AlCutter, line 594: Rather than remove it entirely, how about making it a `klog.V(1).Infof` (or even `V(2)`) and include the tree size in the logged line?

roger2hk: AWS: https://github.com/transparency-dev/trillian-tessera/blob/a60c8d23e801e11eb7511a623d144d6a7da3d0d4/storage/aws/aws.go#L760

GCP: https://github.com/transparency-dev/trillian-tessera/blob/a60c8d23e801e11eb7511a623d144d6a7da3d0d4/storage/gcp/gcp.go#L629

roger2hk: The same info logging cannot be found in AWS and GCP storage implementation of `publishCheckpoint`.

haydentherapper: ^ was the reason I chose to delete it, and there's already logging for when a new tree state is published with the new tree size - https://github.com/transparency-dev/trillian-tessera/blob/a60c8d23e801e11eb7511a623d144d6a7da3d0d4/storage/posix/files.go#L346

AlCutter: Integration != Publishing in Tessera, and in fact these are asynchronous processes.
 
Tessera is not yet at a point in its maturity cycle where I'm happy for debug logging to be entirely removed, if there's a disparity between events logged in different storage backends I'd argue for _adding_ similar logging in the other storage implementations as opposed to removing, as @roger2hk's comment above also suggests.

I _am_ happy for debug logging to have its level moved up so that it can be enabled/disabled, which was the suggestion.

haydentherapper: @AlCutter Added logging for checkpoint publishing for all backends.

AlCutter: Ace, thanks!

AlCutter (approved):

PR #560, opened 2025-03-28
Don't fail PR check for benchmarking

AlCutter:
This PR stops the benchmark action from recording a check fail on PRs if the performance delta is over the threshold.

While it's useful to track benchmark history, the "fails" we've seen are almost all false positives due to the environment in which the benchmark is running. So, for now at least, this PR will remove the likely undeserved red X from PR checks, which should remove a bit of cognitive load from reviewers ("ah, _that_ failure is ok, but not _this_ one", etc.)


roger2hk (approved):
I wish there are three different statuses. 

✅
⚠️
❌

AlCutter:
In theory it should drop a comment in the review, but I'm not sure I've seen it actually do that.

Issue #526, opened 2025-03-13
[AWS] Anti-spam support


AlCutter:
**Description**

Native anti-spam support for AWS, likely using Aurora.

PR #568, opened 2025-03-31
[AWS] Antispam

mhutchinson:
This adds experimental support for antispam (dedupe) for AWS-hosted
logs. The hashes and indices of seen entries are stored in a MySQL
database.

This is based heavily on the GCP Spanner version, and still maintains
its core way of operation (i.e. filling in the duplicates by using the
Follower pattern).

Towards #526.

### DIFF
--- a/storage/aws/antispam/aws.go
+++ b/storage/aws/antispam/aws.go
@@ -1,0 +1,437 @@
+// Copyright 2025 The Tessera authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package aws contains an AWS-based antispam implementation for Tessera.
+//
+// A MySQL database provides a mechanism for maintaining an index of
+// hash --> log position for detecting duplicate submissions.
+package aws
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"strings"
+	"sync/atomic"
+	"time"
+
+	tessera "github.com/transparency-dev/trillian-tessera"
+	"github.com/transparency-dev/trillian-tessera/api/layout"
+	"k8s.io/klog/v2"
+
+	_ "github.com/go-sql-driver/mysql"
+)
+
+const (
+	DefaultMaxBatchSize      = 64
+	DefaultPushbackThreshold = 1024
+
+	// SchemaCompatibilityVersion represents the expected version (e.g. layout & serialisation) of stored data.
+	//
+	// A binary built with a given version of the Tessera library is compatible with stored data created by a different version
+	// of the library if and only if this value is the same as the compatibilityVersion stored in the Tessera table.
+	//
+	// NOTE: if changing this version, you need to consider whether end-users are going to update their schema instances to be
+	// compatible with the new format, and provide a means to do it if so.
+	SchemaCompatibilityVersion = 1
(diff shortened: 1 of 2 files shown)


storage/aws/antispam/aws.go
```diff
@@ -0,0 +1,437 @@
...
+					vals := make([]any, 0, 2*len(curEntries))
+					for i, e := range curEntries {
+						sqlStr += "(?, ?),"
+						vals = append(vals, e, curIndex+uint64(i))
+					}
+					sqlStr = strings.TrimSuffix(sqlStr, ",")
+
+					_, err := f.as.dbPool.ExecContext(ctx, sqlStr, vals...)
+					if err != nil {
+						return fmt.Errorf("failed to insert into AntispamIDSeq with query %q: %v", sqlStr, err)
```
AlCutter, line 400: Could potentially avoid the `TrimSuffix`:

```suggestion
					args := make([]string, 0, len(curEntries))
					vals := make([]any, 0, 2*len(curEntries))
					for i, e := range curEntries {
						args = append(args, "(?, ?)")
						vals = append(vals, e, curIndex+uint64(i))
					}
					sqlStr := fmt.Sprintf("INSERT IGNORE INTO AntispamIDSeq (h, idx) VALUES %s", strings.Join(args, ","))

					if _, err := f.as.dbPool.ExecContext(ctx, sqlStr, vals...); err != nil {
						return fmt.Errorf("failed to insert into AntispamIDSeq with query %q: %v", sqlStr, err)
```

storage/aws/antispam/aws.go
```diff
@@ -0,0 +1,437 @@
...
+					batch = append(batch, c)
+				}
+				curEntries = batch
+				curIndex = followFrom
+
+				// Store antispam entries.
+				//
+				// Note that we're writing the antispam entries outside of the transaction here. The reason is because we absolutely do not want
+				// the transaction to fail if there's already an entry for the same hash in the IDSeq table.
+				//
```
AlCutter, line 383: You don't have this problem in MySQL since you're using `INSERT IGNORE`, I wonder if it might be best to just insert these rows in the `tx` you have open for now and then look at other approaches if there's a performance optimsation which could be done, etc.?

storage/aws/antispam/aws.go
```diff
@@ -0,0 +1,437 @@
...
+	if _, err := s.dbPool.ExecContext(ctx,
+		`CREATE TABLE IF NOT EXISTS AntispamFollowCoord (
+			id INT UNSIGNED NOT NULL,
+			nextIdx BIGINT UNSIGNED NOT NULL,
+			PRIMARY KEY (id)
+		)`); err != nil {
+		return err
+	}
+	// Set default values for a newly initialised schema - these rows being present are a precondition for
+	// sequencing and integration to occur.
```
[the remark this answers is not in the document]
mhutchinson: Done.

storage/aws/antispam/aws.go
```diff
@@ -0,0 +1,437 @@
...
+	// Note that this will only succeed if no row exists, so there's no danger
+	// of "resetting" an existing log.
```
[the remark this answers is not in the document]
mhutchinson: Done.

storage/aws/antispam/aws.go
```diff
@@ -0,0 +1,437 @@
...
+		for workDone := true; workDone; {
+			err = func() error {
+				tx, err := f.as.dbPool.BeginTx(ctx, &sql.TxOptions{
+					ReadOnly: false,
+				})
+				if err != nil {
+					return err
+				}
+
+				row := tx.QueryRowContext(ctx, "SELECT nextIdx FROM AntispamFollowCoord WHERE id = 0")
```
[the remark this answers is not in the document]
mhutchinson: Good call.

storage/aws/antispam/aws.go
mhutchinson: Thanks for confirming this. This was something I wanted to talk to you about, but forgot because of the break between starting this PR and finishing it.

storage/aws/antispam/aws.go
mhutchinson: Yeah this is a bit tidier.

PR #572, opened 2025-04-02
Fixed go.mod file to remove exclude

mhutchinson:
Exclude directives break running commands in the repository in the
normal go run way:

```shell
go run github.com/transparency-dev/trillian-tessera/cmd/examples/posix-oneshot@main --storage_dir=/tmp/mylog --initialise

The go.mod file for the module providing named packages contains one or
more exclude directives. It must not contain directives that would cause
it to be interpreted differently than if it were the main module.
```

I've run `go mod tidy` after this and it seems OK. We'll need to keep an
eye for dependabot making a mess.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/transparency-dev/trillian-tessera
 
 go 1.24.0
+
 require (
 	cloud.google.com/go/spanner v1.78.0
 	cloud.google.com/go/storage v1.51.0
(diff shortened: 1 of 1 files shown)


AlCutter (approved):

mhutchinson:
> The exclude directives were added in #435 which was not a Dependabot update.

It wasn't a dependabot update, but was done to fix a mess that a dependabot update was making.

PR #578, opened 2025-04-03
Move witnessing fully into lifecycle layer

### LINKED ISSUE #516: Clean up how drivers set up witnessing
#515 finishes the implementation of witnessing. However, the diff in that PR feels a bit messy. A cleaner way to implement this might be to make the `NewCP` method of options automatically configure the decorators around this method. This would require more plumbing in order to make the `ReadTile` method available at that layer.

AlCutter:
This PR consolidates responsibility for dealing with getting checkpoints witnessed in the lifecycle layer, removing the need for storage implementations to be aware of it.

This is a continuation of the approach taken for creating checkpoints and signing them.

Down the line we may want to add an (optional) mechanism to thread custom `http.Clients` in, currently applications could achieve something similar by configuring `http.DefaultClient`.

Fixes #516

### DIFF
--- a/internal/witness/witness.go
+++ b/internal/witness/witness.go
@@ -36,8 +36,24 @@ import (
 
 var ErrPolicyNotSatisfied = errors.New("witness policy was not satisfied")
 
+// WitnessGroup defines a group of witnesses, and a threshold of
+// signatures that must be met for this group to be satisfied.
+// Witnesses within a group should be fungible, e.g. all of the Armored
+// Witness devices form a logical group, and N should be picked to
+// represent a threshold of the quorum. For some users this will be a
+// simple majority, but other strategies are available.
+// N must be <= len(WitnessKeys).
 type WitnessGroup interface {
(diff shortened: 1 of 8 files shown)


internal/witness/witness.go
```diff
@@ -29,19 +29,23 @@ import (
...
 	"github.com/transparency-dev/trillian-tessera/internal/parse"
 	"golang.org/x/mod/sumdb/note"
 )
 
 var ErrPolicyNotSatisfied = errors.New("witness policy was not satisfied")
 
+type WitnessGroup interface {
+	Satisfied(cp []byte) bool
+	Endpoints() map[string]note.Verifier
+}
```
mhutchinson, line 42: Doc strings?

mhutchinson (approved):

internal/witness/witness.go
AlCutter: Added Doc Martin's

PR #579, opened 2025-04-03
Add antispam overview

AlCutter:
This PR adds a high-level design overview of Tessera's antispam mechanism.

#469 #155

### DIFF
--- a/docs/design/antispam.md
+++ b/docs/design/antispam.md
@@ -8,19 +8,22 @@ resubmitting entries which are already present in the log.
 From a transparency perspective, this is unlikely to be a problem - an entry is _discoverable_ if it's
 present in a log; this property doesn't diminish if there are further copies of it the log.
 
-However, from a cost & operational point of view, particularly for large scale logs, it would be
-preferable to try to minimise the number of duplicate entries, and this is what Tessera's anti-spam
-mechanism is intended to do.
+However, particularly for large scale logs, there is a potentially significant cost to both the log operator
+(in terms of egress & storage), and to monitors/verifiers within the ecosystem (in terms of ingress, and possibly
+compute & storage, too) to allowing the unfetter addition of duplicate entries to a log.
+
+It would, therefore, be preferable to try to minimise the number of duplicate entries, and this is what
+Tessera's anti-spam mechanism is intended to do.
 
 Tessera anti-spam provides _best effort_ support for minimising duplicate entries submitted to a log.
 
 ## Non-goals
 
 Tessera's anti-spam mechanisms are _explicitly not_ intended to provide strong/atomic deduplication, nor any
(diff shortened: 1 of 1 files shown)


docs/design/antispam.md
```diff
@@ -0,0 +1,138 @@
...
+preferable to try to minimise the number of duplicate entries, and this is what Tessera's anti-spam
+mechanism is intended to do.
+
+Tessera anti-spam provides _best effort_ support for minimising duplicate entries submitted to a log.
+
+## Non-goals
+
+Tessera's anti-spam mechanisms are _explicitly not_ intended to provide strong/atomic deduplication, nor any
+guarantee about uniqueness of entries in a log. These properties are not generally required, but are
+expensive and have a profound impact on log performance.
```
mhutchinson, line 21:
```suggestion
guarantee about uniqueness of entries in a log. These properties are not generally required, are
expensive, and have a profound impact on log performance.
```

docs/design/antispam.md
```diff
@@ -0,0 +1,138 @@
...
+
+If your transparency application needs such properties you will need to handle this "upstream" of Tessera.
```
mhutchinson, line 23: I would just say in the personality instead of upstream.

docs/design/antispam.md
```diff
@@ -0,0 +1,138 @@
...
+  persistent:B -[Lookup]-> T:dedupdata
+  populater:T -[Insert]-> B:dedupdata
+
+  integrate:B -[Write]-> T:bundles
+  bundles:B -[Read]- T:f2
+  f2:L -- R:f1
+  f1:L --> R:populater
+```
+
+If the process of following the log falls too far behind a configurable threshold, the decorator will start
```
mhutchinson, line 72: Maybe link to the configurable threshold?

docs/design/antispam.md
```diff
@@ -0,0 +1,138 @@
...
+1. Normal operation
+2. Degraded anti-spam
+3. Degraded integration
+
+As currently implemented, the Tessera anti-spam implementation consists of a layered defense: 
+
+1. An in-memory deduplication cache, followed by:
+2. A persistent deduplication index.
+
+This gives fast and cheap protection against (T1) in all failure scenarios, and falls back to the more costly interaction
```
mhutchinson, line 94: You may have elided this for brevity, but in reality the in-memory version may allow N copies through, where N is the number of Tessera Appender instances you have (i.e. probably the number of personality instances).

mhutchinson (approved):

docs/design/antispam.md
AlCutter: I did consciously leave it out, but it is important so I've added a note about that.

PR #580, opened 2025-04-04
Add initial OTel instrumentation to Tessera

AlCutter:
This PR adds some basic OTel span instrumentation to Tessera (appender lifecycle, GCP antispam, and client).

This should give us a feel for fit, and allow for a bit of experimentation.


append_lifecycle.go
```diff
@@ -44,6 +47,41 @@ const (
...
+	appenderAddsTotal    metric.Int64Counter
+	appenderAddHistogram metric.Int64Histogram
+	appenderTreeSize     metric.Int64Gauge
+)
+
+func init() {
+	var err error
+
+	appenderAddsTotal, err = meter.Int64Counter(
+		"appender.add.calls",
```
mhutchinson, line 60: I suspect we should use `tessera.*` as a namespace. This:
a) makes it easy for personalities operators to find all of the metrics we are exporting
b) avoids conflict on the remote chance that they are tracking how many times they, e.g. call the appender.

mhutchinson: Or to go full [guidelines](https://opentelemetry.io/docs/specs/semconv/general/naming/#recommendations-for-application-developers) (and a bit Java): `dev.transparency.tessera.*`. Seems a bit much for my tastes, but them's the recommendations.

internal/otel/cast.go
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package otel
+
+import "math"
+
+// Clamp64 casts a uint64 to an int64, clamping it at MaxInt64 if the value is above.
+//
+// Intended only for coverting Tessera uint64 internal values to int64 for us with
```
mhutchinson, line 21:
```suggestion
// Intended only for coverting Tessera uint64 internal values to int64 for use with
```

internal/otel/cast.go
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package otel
+
+import "math"
+
+// Clamp64 casts a uint64 to an int64, clamping it at MaxInt64 if the value is above.
+//
+// Intended only for converting Tessera uint64 internal values to int64 for us with
```
mhutchinson, line 21:
```suggestion
// Intended only for converting Tessera uint64 internal values to int64 for use with
```

mhutchinson (approved):

Issue #570, opened 2025-04-01
Metrics for drivers


haydentherapper:
For any connections that Tessera manages for its drivers, like for AWS and GCP, it would be helpful to provide metrics to track availability and latency.

I think this issue is relevant to all drivers, and am filing this since we've mostly been working with the GCP driver, but some storage backends that don't manage outbound connections, like POSIX, might not need metrics.

I would suggest using Prometheus for the metrics. Tessera can manage creation of the metrics and writing to the metrics, and the personality using Tessera can either provide a registry on initialization of the driver to register the metrics, or Tessera can provide a function that exposes the metrics for the personality to register.

AlCutter:
Yup, metrics is on the list of things we're looking at currently.

I'm trying out OpenTelemetry currently, as this seems to be the thing that observability is standardising around (OpenCensus is dead, Prom supports OTEL, GCP libs are instrumented with OTel, etc.)

AlCutter:
@haydentherapper, I think there's now a reasonable coverage of metrics & traces for the appender Tessera API (should be some for an OTel aware `client` too, although I've not tested that).

Take a look at the `cmd/conformance/gcp/*.go` files to see how to enable the export of metrics into GCP from your own binary, once you've done something similar you should see metrics (and traces, if configured) appearing in your GCP project's `Monitoring` section.

I'm going to close this as done for now, but let me know if there are other specific metrics you have a concrete need for which are not already covered!

PR #581, opened 2025-04-07
Add metrics around integration

AlCutter:
This PR adds metrics which offer insights into how integration is performing.

Towards #570

### DIFF
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -190,7 +190,7 @@ func NewAppender(ctx context.Context, d Driver, opts *AppendOptions) (*Appender,
 	for i := len(opts.addDecorators) - 1; i >= 0; i-- {
 		a.Add = opts.addDecorators[i](a.Add)
 	}
-	sd := newIntegrationStats()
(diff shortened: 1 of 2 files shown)


append_lifecycle.go
```diff
@@ -174,9 +207,61 @@ func NewAppender(ctx context.Context, d Driver, opts *AppendOptions) (*Appender,
...
+type integrationStats struct {
+	sample atomic.Pointer[idxAt]
+}
+
+func newIntegrationStats() *integrationStats {
+	return &integrationStats{}
+}
+
+func (i *integrationStats) saw(idx uint64) {
+	i.sample.CompareAndSwap(nil, &idxAt{idx: idx, at: time.Now()})
```
mhutchinson, line 224: This will only work the first time, and then `i.sample` will be non-nil and this will always fail... is that intentional?

AlCutter: It gets reset below in the `latency` func - I've added a bunch of comments to make this clearer.

mhutchinson (approved):

PR #587, opened 2025-04-08
Add follower metrics

AlCutter:
This PR adds a couple of `Follower`-related metrics:
* Position (by follower name): which exposes the highest index each follower has consumed.
* Lag (by follower name): how many entries behind the currently integrated tree each follower each follower is.

Towards #570

### DIFF
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -236,9 +236,9 @@ func followerStats(ctx context.Context, f Follower, size func(context.Context) (
 		case <-t.C:
 		}
 
-		idx, err := f.Position(ctx)
+		idx, err := f.EntriesProcessed(ctx)
 		if err != nil {
-			klog.Errorf("followerStats: follower %q Position(): %v", name, err)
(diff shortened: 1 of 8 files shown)


append_lifecycle.go
```diff
@@ -207,6 +226,31 @@ func NewAppender(ctx context.Context, d Driver, opts *AppendOptions) (*Appender,
...
+	name := f.Name()
+	t := time.NewTicker(200 * time.Millisecond)
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-t.C:
+		}
+
+		idx, err := f.EntriesProcessed(ctx)
```
mhutchinson, line 239:
```suggestion
		count, err := f.EntriesProcessed(ctx)
```

If I was really pedantic...

AlCutter: if I _were_ really pedantic... 🤣

mhutchinson (approved):

PR #592, opened 2025-04-14
[AWS Antispam] End-to-end antispam for conformance

mhutchinson:
Antispam DB can now be created via the Terraform, and the conformance
command takes flags that allow antispam to be enabled and configured
with this DB.

Docs have been added to the end of the codelab instructions to document
how to deploy this.

Some performance testing was performed and results look good: with 1200
write workers doing 1200 writes per second, with 30% dupe chance, the
log was keeping up, growing by around 800 leaves per second.

```
Read (8 workers): Current max: 20/s. Oversupply in last second: 0
Write (1200 workers): Current max: 1200/s. Oversupply in last second: 0
TreeSize: 199416 (Δ 745qps over 30s)
Time-in-queue: 6ms/411ms/1056ms (min/avg/max)
Observed-time-to-integrate: 43ms/1291ms/4346ms (min/avg/max)
```

There are a couple of fixes in here for state/transaction management to
be more resilient when following the DB. Previously TXs were sometimes
not closed, which would lock the table until TX expiry.

Towards #526.

### DIFF
--- a/cmd/conformance/aws/main.go
+++ b/cmd/conformance/aws/main.go
@@ -91,7 +91,7 @@ func main() {
 		}
 		antispam, err = aws_as.NewAntispam(ctx, antispamMysqlConfig().FormatDSN(), asOpts)
 		if err != nil {
-			klog.Exitf("Failed to create new GCP antispam storage: %v", err)
+			klog.Exitf("Failed to create new AWS antispam storage: %v", err)
 		}
 	}
 	appender, shutdown, _, err := tessera.NewAppender(ctx, driver, tessera.NewAppendOptions().
--- a/cmd/conformance/aws/main.go
+++ b/cmd/conformance/aws/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/go-sql-driver/mysql"
 	tessera "github.com/transparency-dev/trillian-tessera"
 	"github.com/transparency-dev/trillian-tessera/storage/aws"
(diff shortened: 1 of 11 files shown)


mhutchinson:
Moved this to draft while I consider whether having the antispam in the same DB as the other tables is feature or bug.

cmd/conformance/aws/main.go
```diff
@@ -78,12 +82,24 @@ func main() {
...
+	var antispam tessera.Antispam
+	// Persistent antispam is currently experimental, so there's no documentation yet!
+	if *antispamEnable {
+		asOpts := aws_as.AntispamOpts{
+			MaxBatchSize:      64,
+			PushbackThreshold: 1024,
+		}
+		antispam, err = aws_as.NewAntispam(ctx, antispamMysqlConfig().FormatDSN(), asOpts)
+		if err != nil {
+			klog.Exitf("Failed to create new GCP antispam storage: %v", err)
```
AlCutter, line 94:
```suggestion
			klog.Exitf("Failed to create new AWS antispam storage: %v", err)
```

AlCutter (approved):
Might be worth enabling by default in the `live/aws/conformance/ci` config so it's exercised in CI tests similar to the GCP impl?

cmd/conformance/aws/main.go
mhutchinson: Done.

Issue #588, opened 2025-04-09
Monitoring should make it clear when the log is quiescent


mhutchinson:
For #574 I'm attempting to document how freezing a log will work. The natural instructions are:
 1. Stop writing to the Appender (probably via a flag flip)
 2. Ensure an Appender is instantiated, and thus processing sequenced/integrated entries in the background
 3. Wait until it is done

The third step is not clear to define to the user at the moment. The monitoring added via #570 gets us a lot of the way there for a running log that is being written to, but for a new process that is brought up (e.g. because it was restarted due to a flag flip), I don't think there's sufficient information to know when all entries have been published and there is no more background sequencing/integrating/publishing work to do.

PR #600, opened 2025-04-15
Add NextIndex metric

AlCutter:
This PR adds a metric which exposes the next index which will be used for sequencing.

Such a metric helps determine whether a log is fully drained, critical information when switching lifecycle modes. 

Towards #588

### DIFF
--- a/lifecycle.go
+++ b/lifecycle.go
@@ -64,11 +64,12 @@ type LogReader interface {
 	// log should be used. If in doubt, use ReadCheckpoint instead.
 	IntegratedSize(ctx context.Context) (uint64, error)
 
-	// PendingCount returns the number of sequenced but not-yet-integrated entries.
+	// NextIndex returns the first as-yet unassigned index.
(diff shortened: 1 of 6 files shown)


lifecycle.go
```diff
@@ -64,6 +64,13 @@ type LogReader interface {
 	// log should be used. If in doubt, use ReadCheckpoint instead.
 	IntegratedSize(ctx context.Context) (uint64, error)
 
+	// NextIndex returns the first as-yet unassigned index.
+	//
+	// In a quiescent log, this will be the same as the checkpoint size. In a log with entries actively
+	// being added, this number will be higher since it will take sequenced but not-yet-integrated
+	// entries into account.
```
mhutchinson, line 71: Not-yet-integrated, and also not-yet-published.

mhutchinson (approved):
Approved. Note that the README needs to be updated to reference this metric when freezing the log.

PR #605, opened 2025-04-16
Adjust antispam defaults, and use them in conformance bins

AlCutter:
This PR adjusts defaults for the antispam implementations.

The new numbers come from running the GCP implementation at scale for a while. AWS has been _optimistically_ bumped to match as implementation is similar, but should be revisited once more experience is gained with it.


mhutchinson (approved):
Can you add a line to the PR description to describe how these numbers were picked?

PR #618, opened 2025-04-22
Use BigEndian in badger

AlCutter:
This PR changes the encoding of the hash used to key the Badger-based POSIX antispam DB to use BigEndian.

This is really just to keep it the same as other encodings we do in here (e.g. all the C2SP encodings are BigEndian), it doesn't _really_ matter as it's entirely internal, but consistency is nice.


patflynn (approved):
why the switch

AlCutter:
> why the switch

Just to keep it the same as other encodings we do here (e.g. all the C2SP encodings are BigEndian), it doesn't _really_ matter as it's entirely internal, but consistency is nice.

PR #622, opened 2025-04-25
NewProofBuilder operates in terms of tree size

AlCutter:
This PR makes `NewProofBuilder` require only a tree size, rather than a full `checkpoint`.

This is safe as generated proofs must be verified against checkpoint root hashes by the consumer anyway; any inconsistency will be detected at that point, and brings it into line with most other proof building implementations (Trillian, GoSumDB, tlog, etc.)

The change provides a means to construct consistency proofs without having to hold both checkpoints (e.g. if you're trying to convince someone else of consistency between a checkpoint they hold, and a newer checkpoint you hold - this is exactly what happens during the tlog-witness protocol).

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -188,11 +188,6 @@ type ProofBuilder struct {
 // The returned ProofBuilder can be re-used for proofs related to a given tree size, but
 // it is not thread-safe and should not be accessed concurrently.
 func NewProofBuilder(ctx context.Context, treeSize uint64, f TileFetcherFunc) (*ProofBuilder, error) {
-	ctx, span := tracer.Start(ctx, "tessera.client.NewProofBuilder")
-	defer span.End()
-
-	span.SetAttributes(logSizeKey.Int64(otel.Clamp64(treeSize)))
-
 	pb := &ProofBuilder{
 		treeSize:  treeSize,
 		nodeCache: newNodeCache(f, treeSize),
--- a/client/client.go
+++ b/client/client.go
@@ -212,17 +212,19 @@ func (pb *ProofBuilder) InclusionProof(ctx context.Context, index uint64) ([][]b
 	return pb.fetchNodes(ctx, nodes)
 }
 
-// ConsistencyProof constructs a consistency proof between the smaller passed-in treesize, and
-// the larger size the ProofBuilder was constructed for.
+// ConsistencyProof constructs a consistency proof between the provided tree sizes.
 // This function uses the passed-in function to retrieve tiles containing any log tree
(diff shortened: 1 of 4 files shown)


client/client.go
```diff
@@ -238,23 +205,24 @@ func (pb *ProofBuilder) InclusionProof(ctx context.Context, index uint64) ([][]b
...
-	nodes, err := proof.Inclusion(index, pb.cp.Size)
+	nodes, err := proof.Inclusion(index, pb.treeSize)
 	if err != nil {
 		return nil, fmt.Errorf("failed to calculate inclusion proof node list: %v", err)
 	}
 	return pb.fetchNodes(ctx, nodes)
 }
 
-// ConsistencyProof constructs a consistency proof between the two passed in tree sizes.
+// ConsistencyProof constructs a consistency proof between the smaller passed-in treesize, and
```
roger2hk, line 215:
```suggestion
// ConsistencyProof constructs a consistency proof between the smaller passed-in tree size, and
```

client/client.go
```diff
@@ -181,49 +180,17 @@ func GetEntryBundle(ctx context.Context, f EntryBundleFetcherFunc, i, logSize ui
...
 // NewProofBuilder creates a new ProofBuilder object for a given tree size.
 // The returned ProofBuilder can be re-used for proofs related to a given tree size, but
 // it is not thread-safe and should not be accessed concurrently.
-func NewProofBuilder(ctx context.Context, cp log.Checkpoint, f TileFetcherFunc) (*ProofBuilder, error) {
-	ctx, span := tracer.Start(ctx, "tessera.client.NewProofBuilder")
-	defer span.End()
-
-	span.SetAttributes(logSizeKey.Int64(otel.Clamp64(cp.Size)))
-
+func NewProofBuilder(ctx context.Context, treeSize uint64, f TileFetcherFunc) (*ProofBuilder, error) {
```
roger2hk, line 190: Do we need to mention "no correctness checking of the root hash for the given tree size is performed." in this function doc comment?

roger2hk (approved):

client/client.go
AlCutter: I don't think so, it's probably more surprising that a call to this function would go off and start making network requests before calling one of the `.*Proof` functions below.

client/client.go
AlCutter: (I decided to leave the params here as they were, and pass in both `smaller` and `larger` sizes explicitly for now)

roger2hk: Gotcha. Shall we use `treeSize` for param?

PR #634, opened 2025-05-09
Allow witness signature collection to fail open

AlCutter:
This PR provides what's intended to be a temporary mechanism to enable logs to interact with the witness network in a _fail open_ fashion, whereby they will publish checkpoints & whatever valid witness signatures were obtained for it, regardless of whether the obtained signatures satisfy the specified witness policy.

With this option enabled logs and their clients can start to interact with the witness network in a safe "fail open" fashion as it develops and matures.

### DIFF
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -511,7 +511,7 @@ func (o AppendOptions) valid() error {
 
 // CheckpointPublisher returns a function which should be used to create, sign, and potentially witness a new checkpoint.
 func (o AppendOptions) CheckpointPublisher(lr LogReader, httpClient *http.Client) func(context.Context, uint64, []byte) ([]byte, error) {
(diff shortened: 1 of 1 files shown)


append_lifecycle.go
```diff
@@ -524,7 +525,10 @@ func (o AppendOptions) CheckpointPublisher(lr LogReader, httpClient *http.Client
 
 		cp, err = wg.Witness(ctx, cp)
 		if err != nil {
-			return nil, err
+			if !o.witnessOpts.FailOpen {
```
mhutchinson, line 528: This should have metrics so that operators can safely determine when they are able to switch modes.

mhutchinson (approved):

append_lifecycle.go
AlCutter: Good idea, done

PR #639, opened 2025-05-12
Fix bug in `EntriesPathForLogIndex`

phbnf:
I couldn't see any use of this function in `trillian-tessera` nor `static-ct`, so it's unlikely someone has run into it. The hammer doesn't use this function, because it uses `EntriesPath` directly.


phbnf:
(I thought about deleting it altogether, but I'm not sure it's a good idea: removing it would require clients to re-implement the logic on their side, like the hammer does, increasing the likelihood to reproduce this bug)

AlCutter (approved):
Nice catch!

PR #644, opened 2025-05-14
Add a simple tlog-files fsck tool

AlCutter:
This PR adds a simple `fsck` tool for `tlog-tile` compatible logs.

It recalculates the root hash in a `checkpoint` by reading all entry bundles implied by the checkpoint, verifies that all tiles implied by the checkpoint are present and contain the hashes the should, and that the checkpoint's root hash matches the calculated value.

### DIFF
--- a/internal/fsck/fsck.go
+++ b/internal/fsck/fsck.go
@@ -17,15 +17,17 @@ package fsck
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"fmt"
 	"sync/atomic"
 
+	f_log "github.com/transparency-dev/formats/log"
 	"github.com/transparency-dev/merkle/compact"
 	"github.com/transparency-dev/merkle/rfc6962"
 	"github.com/transparency-dev/trillian-tessera/api"
 	"github.com/transparency-dev/trillian-tessera/api/layout"
-	"github.com/transparency-dev/trillian-tessera/internal/parse"
 	"github.com/transparency-dev/trillian-tessera/internal/stream"
+	"golang.org/x/mod/sumdb/note"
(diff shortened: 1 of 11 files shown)


internal/fsck/fsck.go
```diff
@@ -0,0 +1,241 @@
...
+//
+// The checking will use the provided N parameter to control the number of concurrent workers undertaking
+// this process.
+func Check(ctx context.Context, f Fetcher, N uint, bundleHasher func([]byte) ([][]byte, error)) error {
+	cp, err := f.ReadCheckpoint(ctx)
+	if err != nil {
+		klog.Exitf("fetch initial source checkpoint: %v", err)
+	}
+	klog.V(1).Infof("Fsck: checkpoint:\n%s", cp)
+	// TODO(al): Should really open this with the pubK to verify the checkpoint is well formed.
```
mhutchinson, line 55: IMO it's worth doing this. The more we do this unsafe parsing, the more it will become a dangerous norm.

internal/fsck/fsck.go
```diff
@@ -0,0 +1,241 @@
...
+		ri, b, err := next()
+		if err != nil {
+			klog.Warningf("next: %v", err)
+			break
+		}
+		if err := fTree.AppendBundle(ri, b); err != nil {
+			klog.Warningf("AppendBundle(%v): %v", ri, err)
+			break
+		}
+		if fTree.tree.End() == logSize {
```
mhutchinson, line 95: Can you move this into the condition for the `for`?

```
for fTree.tree.End() < logSize {
...
}
```

mhutchinson (approved):

PR #645, opened 2025-05-15
Follow rename to Tessera

AlCutter:
This PR "fixes up" all the imports and references from `trillian-tessera` to `tessera` following the repo rename.

I've left _some_ references alone (primarily in terraform configs - e.g. IAM service account names) where this could cause a problem for deployed instances.


mhutchinson (approved):
YOLO

PR #651, opened 2025-05-19
bubble up abort other spanner errors to enable retries

bobcallaway:
similar to #497 - we should ensure the errors from some calls in a `ReadWriteTransaction` are wrapped to enable retry logic; `ReadOnlyTransactions` never abort so i think only these are needed

@AlCutter fyi


AlCutter (approved):
Ta (again)!

PR #655, opened 2025-05-20
[Benchmarks] Log some CPU diagnostics info

mhutchinson:
We see a fairly clear bimodal performance distribution for our benchmarks. Nothing in the current output indicates any difference between these runs. I have seen other projects include this CPU info in the filenames to compare apples with apples, and oranges with oranges. Before we wire this info into the file name, first dump it to the log so we can confirm that this is different between slower and faster runs.

### DIFF
--- a/.github/workflows/benchmark-go-main.yml
+++ b/.github/workflows/benchmark-go-main.yml
@@ -30,7 +30,7 @@ jobs:
           go-version-file: go.mod
 
       - name: Run benchmark
-        run: set -o pipefail; go test ./... -benchmem -run=^$ -bench . | tee output.txt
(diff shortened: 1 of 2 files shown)


mhutchinson:
Closing this, as it doesn't provide any more info than was already logged:

```
CPU info AMD EPYC 7763 64-Core Processor ####### new line
goos: linux
goarch: amd64
pkg: github.com/transparency-dev/tessera
cpu: AMD EPYC 7763 64-Core Processor ####### existing line
```

I already checked a slow and a fast run and this CPU info was the same between them. The variance may simply come down to load on the virtual machine hosting the runner. Without using a dedicated machine, there's not much improvement to be made here.

PR #668, opened 2025-05-27
[GCP] Garbage collection

### LINKED ISSUE #697: [GCP] Partial tiles/bundles should be garbage collected
**Description**

Tracking implementation of #670 for GCP.

AlCutter:
This PR adds a "garbage collector" to the GCP storage implementation which periodically removes partial tiles & entry bundles which have become obsoleted by corresponding full versions as the log has grown.

See #672 for corresponding changes to the client code.

Towards #670, fixes #697

### DIFF
--- a/storage/gcp/gcp_test.go
+++ b/storage/gcp/gcp_test.go
@@ -502,6 +502,7 @@ func TestPublishTree(t *testing.T) {
 func TestGarbageCollect(t *testing.T) {
 	ctx := t.Context()
 	batchSize := uint64(60000)
+	integrateEvery := uint64(31234)
 
 	closeDB := newSpannerDB(t)
 	defer closeDB()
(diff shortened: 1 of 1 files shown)


mhutchinson:
> My understanding that beta is going to be tagged ~this week. I'm ok to delay merging this until after tagging beta if that is the case, but I'd appreciate a review in case there are changes to make in the meantime.

Sure, sounds good. I'll take a look at it, but we're aligned that we won't merge until beta is tagged.

storage/gcp/gcp.go
```diff
@@ -949,6 +1008,58 @@ func (s *spannerCoordinator) publishTree(ctx context.Context, minAge time.Durati
 	return nil
 }
 
+// garbageCollect is a long running function which will identify unneeded partial tiles/entry bundles, and call the provided function to remove them.
+//
+// Uses the `GCCoord` table to ensure that only one binary is actively garbage collecting at any given time, and to track progress so that we don't
+// needlessly attempt to GC over regions which have already been cleaned.
+//
+// Returns true if we've "caught up" with the current state of the tree.
+func (s *spannerCoordinator) garbageCollect(ctx context.Context, treeSize uint64, maxDeletes uint, deleteWithPrefix func(ctx context.Context, prefix string) error) error {
```
mhutchinson, line 1017: I discuss this in the comments below, but `maxDeletes` required me to update my intuition. In addition to the non-strict nature of it, my expectations about what it counted as a deletion were wrong.

What I _expected_ was that this would control the maximum number of partial tiles that would be deleted. It actually controls the number of calls to `deleteWithPrefix`, each of which could delete up to 255 partial tiles. Nothing wrong with this, but worth having a doc line to make it clear.

storage/gcp/gcp.go
```diff
@@ -372,6 +428,7 @@ func (a *Appender) publishCheckpoint(ctx context.Context, size uint64, root []by
 type objStore interface {
 	getObject(ctx context.Context, obj string) ([]byte, int64, error)
 	setObject(ctx context.Context, obj string, data []byte, cond *gcs.Conditions, contType string, cacheCtl string) error
+	deleteObjectsWithPrefix(ctx context.Context, prefix string) error
```
mhutchinson, line 431: It's an internal interface, so I'm not overly worried, but this looks like a strange cousin next to these other 2 get/set methods.

storage/gcp/gcp.go
```diff
@@ -949,6 +1008,58 @@ func (s *spannerCoordinator) publishTree(ctx context.Context, minAge time.Durati
...
+	done:
+		for l, f, x := uint64(0), fromSize, treeSize; x > 0; l, f, x = l+1, f>>layout.TileHeight, x>>layout.TileHeight {
+			for ri := range layout.Range(f, x-f, x) {
+				if ri.Partial != 0 || d > maxDeletes {
+					break done
+				}
+				if l == 0 {
+					eg.Go(func() error { return deleteWithPrefix(ctx, layout.EntriesPath(ri.Index, 0)+".p/") })
+					d++
+					fromSize += uint64(ri.N)
```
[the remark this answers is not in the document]
AlCutter: I've switched it a "vertical" implementation - I'd originally discounted that approach in order to avoid repeatedly considering higher level tiles, but avoiding that is far easier than I assumed it would be.

storage/gcp/gcp.go
AlCutter: Open to alternative suggestions, but the thinking was that this enables infra to do this specific task in the best way it can (e.g. POSIX could potentially do an `rmdir()`) rather than force all to conform to an LCD like `findObjectsWithPrefix` and `deleteObject`

mhutchinson (approved):

mhutchinson (approved):
Yeah this reads much better, ta.

PR #681, opened 2025-06-05
Bound spanner request context

AlCutter:
This PR puts a defined timeout on the contexts used for interactions with Spanner.

(see https://cloud.google.com/spanner/docs/transaction-timeout)


AlCutter:
This was more of a speculative change for resilience/recovery in the face of slow/hanging RPCs rather than a performance thing.

I'm not sure how easily this will rebase now, though.

PR #704, opened 2025-06-30
Push partial->full fallback down into fetcher contract

AlCutter:
This PR moves responsibility for falling back to a full resource in the event that a partial one is not found down into the fetcher.

This removes the need for having the fall-back logic in multiple places, and makes it more likely that log interactions will work as expected regardless of whether the log is garbage collecting or not.

There are not many reasons why a client would ever want to just fail when fetching a partial resource rather than fall back, so any client which _does_ want this unusual behaviour will have to provide their own fetchers.

### DIFF
--- a/fsck/fsck.go
+++ b/fsck/fsck.go
@@ -17,6 +17,7 @@ package fsck
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"encoding/base64"
 	"fmt"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


internal/fetcher/fallback.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2025 Google LLC. All Rights Reserved.
```
mhutchinson, line 1: nit: the prevailing licence holder in this repo is "The Tessera authors"

PR #711, opened 2025-07-10
Memoize the outer-most AddFn

AlCutter:
This PR ensures that repeated calls to the future returned by `NewAppender` are idempotent.

### DIFF
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -258,9 +258,26 @@ func NewAppender(ctx context.Context, d Driver, opts *AppendOptions) (*Appender,
 		ctx, span := tracer.Start(ctx, "tessera.Appender.Add")
 		defer span.End()
 
-		return t.Add(ctx, entry)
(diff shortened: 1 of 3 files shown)


append_lifecycle.go
```diff
@@ -258,11 +258,29 @@ func NewAppender(ctx context.Context, d Driver, opts *AppendOptions) (*Appender,
 		ctx, span := tracer.Start(ctx, "tessera.Appender.Add")
 		defer span.End()
 
-		return t.Add(ctx, entry)
+		// NOTE: We memoize the returned value here so that repeated calls to the returned
```
phbnf, line 261: Indentation looks a bit off.
```
		// NOTE: We memoize the returned value here so that repeated calls to the returned
		//       future don't result in unexpected side-effects from inner AddFn functions
		//       being called multiple times.
		//		 Currently this is the outermost wrapping of Add so we do the memoization
		//		 here, if this changes, ensure that move the memoization call so that this
		//       remains true.
```

PR #712, opened 2025-07-11
Ensure antispam doesn't push back when caught up

AlCutter:
This PR addresses an issue where the antispam followers could (rarely) get stuck in pushback mode, preventing any further writes to the log until the jobs were restarted.

If the antispam follower fully caught up while it was in a `pushback` condition, _and_ there were no entries either pending or currently being integration, the antispam would become "stuck" in pushback condition and prevent further entries from being added.

### DIFF
--- a/storage/posix/antispam/badger.go
+++ b/storage/posix/antispam/badger.go
@@ -256,11 +256,6 @@ func (f *follower) Follow(ctx context.Context, lr tessera.LogReader) {
 
 				span.SetAttributes(followFromKey.Int64(otel.Clamp64(followFrom)))
 
-				if followFrom > logSize {
-					// Our view of the log is out of date, exit the busy loop and refresh it.
-					workDone = false
-					return nil
-				}
 				var err error
 				if followFrom >= logSize {
 					// Our view of the log is out of date, update it


storage/aws/antispam/aws_test.go
```diff
@@ -108,6 +107,61 @@ func TestAntispam(t *testing.T) {
 	}
 }
 
+func TestAntispamPushback(t *testing.T) {
```
phbnf, line 110: You might want to describe what behavior you're implementing here, such that if we ever change some functions and have to change this test, we know what we want to reproduce exactly.

AlCutter: I've renamed the tests to be clearer about what they're testing, and expanded the comment at the bottom where they check `pushBack`.

storage/posix/antispam/badger.go
```diff
@@ -257,13 +256,34 @@ func (f *follower) Follow(ctx context.Context, lr tessera.LogReader) {
 
 				span.SetAttributes(followFromKey.Int64(otel.Clamp64(followFrom)))
 
-				if followFrom >= size {
+				if followFrom > logSize {
```
phbnf, line 259: Did you mean to keep this?

storage/posix/antispam/badger.go
```diff
@@ -257,13 +256,34 @@ func (f *follower) Follow(ctx context.Context, lr tessera.LogReader) {
...
+				var err error
+				if followFrom >= logSize {
+					// Our view of the log is out of date, update it
+					logSize, err = lr.IntegratedSize(ctx)
+					if err != nil {
+						workDone = true
+						return fmt.Errorf("populate: IntegratedSize(): %v", err)
+					}
+					switch {
+					case followFrom > logSize:
```
phbnf, line 273: If you meant to keep the above, then this block will never be executed. If removing the block above works, it means that this condition is not covered by a test.

phbnf (approved):

PR #717, opened 2025-07-16
Reduce GCS memory footprint

AlCutter:
This PR significantly reduces the amount of memory used when running on GCP.

Previously, whenever a GCS resource was being created (which happens primarily during integration, with potentially multiple 10s of resources being updated concurrently in logs with high throughput) the GCS client library would allocate 16MB of buffer per object, resulting in a couple of hundred MB of wasted allocation.

With the change, the buffer is reduced to the size of the resource being written plus a small surplus as recommended by the docs.


mhutchinson (approved):
nice

PR #718, opened 2025-07-18
Simplify PublicationAwaiter

AlCutter:
This PR makes the `PublicationAwaiter` more concise by using a `sync.Cond` so that calls to `Await()` are notified to check and release themselves when an updated checkpoint is seen, rather than relying on managing many channels.

This happens to also reduce memory usage & allocs by around 30% but isn't a strong driver.

### DIFF
--- a/await.go
+++ b/await.go
@@ -17,12 +17,9 @@ package tessera
 import (
 	"context"
 	"errors"
-	"fmt"
 	"os"
 	"sync"
 	"time"
-
-	"container/list"
 
 	"github.com/transparency-dev/tessera/internal/parse"
 	"k8s.io/klog/v2"
@@ -33,7 +30,7 @@ import (
 // to fetch checkpoints using the `readCheckpoint` function.
 func NewPublicationAwaiter(ctx context.Context, readCheckpoint func(ctx context.Context) ([]byte, error), pollPeriod time.Duration) *PublicationAwaiter {
 	a := &PublicationAwaiter{
-		waiters: list.New(),
+		c: sync.NewCond(&sync.Mutex{}),
 	}
 	go a.pollLoop(ctx, readCheckpoint, pollPeriod)
 	return a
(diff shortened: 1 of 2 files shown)


mhutchinson (approved):

await_test.go
```diff
@@ -226,3 +226,29 @@ func TestAwait_multiClient(t *testing.T) {
...
+func BenchmarkAwait(b *testing.B) {
+	cpFormat := "origin/\n%d\nhash\n\nsig"
+	cpSize := atomic.Uint64{}
+	readCP := func(_ context.Context) ([]byte, error) {
+		return fmt.Appendf(nil, cpFormat, cpSize.Load()), nil
+	}
+	a := NewPublicationAwaiter(b.Context(), readCP, time.Millisecond)
+	go func() {
+		for {
+			t := time.NewTicker(time.Millisecond)
```
mhutchinson, line 239: Shouldn't this be outside of the loop?

await.go
```diff
@@ -100,101 +91,36 @@ func (a *PublicationAwaiter) Await(ctx context.Context, future IndexFuture) (Ind
...
-		_, size, _, err := parse.CheckpointUnsafe(rawCp)
-		if err != nil {
-			a.releaseClientsErr(err)
-			continue
-		}
-		a.releaseClients(size, rawCp)
+		a.checkpoint = cp
+		a.size = cpSize
+		a.err = cpErr
+		a.c.Broadcast()
```
mhutchinson, line 123: This broadcast seems likely to lead to a large CPU spike in high throughput situations, but worth trying it.

await.go
```diff
@@ -100,101 +91,36 @@ func (a *PublicationAwaiter) Await(ctx context.Context, future IndexFuture) (Ind
...
-			// work happens in personalities that aren't performing writes.
-			a.mu.Lock()
-			hasClients := a.waiters.Front() != nil
-			a.mu.Unlock()
-			if !hasClients {
-				continue
+			cp, cpErr = readCheckpoint(ctx)
+			switch {
+			case errors.Is(cpErr, os.ErrNotExist):
+				return
```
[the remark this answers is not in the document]
AlCutter: Good spot, this should've been changed to a `continue` - done.

AlCutter (commented):
I'll merge this now, we can back it out if it causes any problems.

PR #736, opened 2025-08-12
Disambiguation between Integration and Publication

phbnf:
Personalities that need to wait should wait for a checkpoint to be published, rather than simple integration. For ease of writing, I think that it's okay to consider that "entries are published" only and only when  "checkpoint covering these entries" has been published (even if the entries will be made public before then).

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,9 +267,13 @@ Writing to the log follows this flow:
     
 Once an index has been returned, the new data is sequenced, but not necessarily integrated into the log.
 
-As discussed above in [Integration](#integration), sequenced entries will be _asynchronously_ integrated into the log and be made available via the read API.
-Some personalities may need to block until this has been performed, e.g. because they will provide the requester with an inclusion proof, which requires integration.
-Such personalities are recommended to use [Synchronous Publication](#synchronous-publication) to perform this blocking.
+As discussed above in [Integration](#integration) and [Publishing](#publishing),
+sequenced entries will be first _asynchronously_ integrated into the log, and
+then published on the read API by a second _asynchronous_ process. Some
+personalities may need to block until this has been performed, e.g. because they
+will provide the requester with an inclusion proof, which requires integration
+and publication. Such personalities are recommended to use [Synchronous Publication](#synchronous-publication)
+to perform this blocking.
 
 #### Reading from the Log
 


README.md
```diff
@@ -267,9 +267,13 @@ Writing to the log follows this flow:
...
 
-As discussed above in [Integration](#integration), sequenced entries will be _asynchronously_ integrated into the log and be made available via the read API.
-Some personalities may need to block until this has been performed, e.g. because they will provide the requester with an inclusion proof, which requires integration.
-Such personalities are recommended to use [Synchronous Publication](#synchronous-publication) to perform this blocking.
+As discussed above in [Integration](#integration) and [Publishing](#publishing),
+sequenced entries will be first _asynchronously_ integrated into the log, and
+then published on the read API by a second _asynchronous_ process. Some
+personalities may need to block until this has been performed, e.g. because they
+will provide the requester with an inclusion proof, which requires integration
+and publication. Such personalities are recommended to use [Synchronous Publication](#synchronous-publication)
```
AlCutter, line 275:
```suggestion
As discussed above in [Integration](#integration) and [Publishing](#publishing),
sequenced entries will be _asynchronously_ integrated into the log, with a new `checkpoint` committing to these entries being published periodically by another asynchronous process.

Some personalities may need to block until this has been performed, e.g. because they will provide the requester with an inclusion proof, which requires both then integration of the entry and the publication of a new checkpoint to have completed. Such personalities are recommended to use [Synchronous Publication](#synchronous-publication)
```

AlCutter (approved):
This is kinda covered up in the [`#Publishing` section](https://github.com/transparency-dev/tessera#:~:text=An%20entry%20is%20considered%20published%20once%20it%20is%20committed%20to%20by%20a%20published%20Checkpoint%20(i.e.%20a%20published%20Checkpoint%27s%20size%20is%20larger%20than%20the%20entry%27s%20assigned%20index).):

> An entry is considered published once it is committed to by a published Checkpoint (i.e. a published Checkpoint's size is larger than the entry's assigned index).

But it's not a bad idea to stress the point again here.
I've suggested a small tweak in wording to bring it closer into line  with publication being mainly about checkpoints to avoid the potential for any misinterpretation/confusion about static resources appearing _before_ a checkpoint has been successfully published. Have a look and see what you think.

Issue #762, opened 2025-09-11
Expose progress and detailed stats/status in fsck


AlCutter:
**Description**

AlCutter:
This is now done, and re-used in the TesseraCT fsck too, too.

PR #767, opened 2025-09-15
Add a TUI to cmd/fsck

AlCutter:
This PR adds a simple UI to the `fsck` too which allows the user to visibly see the progress.

When running against large logs, `fsck` can take a long time to complete, so it's important to show that the process is ongoing and progress is being made.

This PR adds several combination status/progress bar widgets - one for entry bundles, and then one for each layer of tiles in the target log. The app uses the `Status` call from `fsck.Fsck` to periodically fetch updated information on the state of the check, and updates the bars accordingly, using a combination of colour and glyph to represent each of the possible states a range of static artifacts can be in.

The UI that this PR adds doesn't completely take over the terminal/use the "alt screen", but works in a way which is complementary to the existing `klog` info lines and terminal scroll-back:

progress

Towards #762


phbnf (approved):
This is amazing!

PR #772, opened 2025-09-16
Fsck add key to UI

AlCutter:
This PR adds a reusable `FsckPanel` which acts as a container for all the UI elements required to show information about an ongoing fsck operations, and adds a simple key to the `fsck` UI for the progress bars.

### DIFF
--- a/cmd/fsck/tui/layerbar.go
+++ b/cmd/fsck/tui/layerbar.go
@@ -38,7 +38,7 @@ var (
 	// stateStyle defines how the different fsck.State types show up in the progress bar.
 	stateStyles = []stateStyle{
 		{
-			state:    fsck.Unknown,
+			state:    fsck.Unchecked,
 			style:    lipgloss.NewStyle().Foreground(lipgloss.Color("#313244")),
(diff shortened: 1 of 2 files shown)


cmd/fsck/tui/layerbar.go
```diff
@@ -36,55 +36,84 @@ var (
...
-		// priority maps each possible state to a relative importance of that priority.
-		// This is used when displaying the overall state of a range of resources where multiple
-		// states are present.
-		// The highest priority state represented in the range "wins".
-		priority int
-	}{
-		fsck.Unknown: {
+	stateStyles = []stateStyle{
+		{
+			state:    fsck.Unknown,
```
phbnf, line 41: Should this be "Scheduled" or something like this? Unknown kind of indicates that there might be something going wrong, when really, it's just that the leaf hasn't been fetched yet. Or maybe there should be a "Scheduled" AND an "Unknown" state?

AlCutter: Good call, I've renamed to "Unchecked" which is what it really represents.

PR #785, opened 2025-10-05
fix missing env var in example in readme

warpfork:
Set the "LOG_DIR" var (already used in the example script) to a simple dummy value which is likely to work, so that the example script as a whole now fully runs with no modification needed.

(Every other detail was already fully copy-pasta friendly -- well done and thank you, author of that! -- so I thought I'd put on the finishing touch.)

### DIFF
--- a/cmd/examples/posix-oneshot/README.md
+++ b/cmd/examples/posix-oneshot/README.md
@@ -7,19 +7,16 @@
 The commands below create a new log and add entries to it, and then show a few approaches to inspect the contents of the log.
 
 ```shell
-# Set the keys via environment variables
+# Set the keys via environment variables, and a local directory where our demo log will be built.
 export LOG_PRIVATE_KEY="PRIVATE+KEY+example.com/log/testdata+33d7b496+AeymY/SZAX0jZcJ8enZ5FY1Dz+wTML2yWSkK+9DSF3eg"
 export LOG_PUBLIC_KEY="example.com/log/testdata+33d7b496+AeHTu4Q3hEIMHNqc6fASMsq3rKNx280NI+oO5xCFkkSx"
+export LOG_DIR="/tmp/tlog"
 
 # Create files containing new leaves to add
 mkdir /tmp/stuff
 echo "foo" > /tmp/stuff/foo
 echo "bar" > /tmp/stuff/bar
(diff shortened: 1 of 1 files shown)


cmd/examples/posix-oneshot/README.md
```diff
@@ -17,6 +17,10 @@ echo "foo" > /tmp/stuff/foo
 echo "bar" > /tmp/stuff/bar
 echo "baz" > /tmp/stuff/baz
 
+# Make a blank dir where we'll store the log data
+export LOG_DIR="/tmp/tlog"
+mkdir "$LOG_DIR"
```
roger2hk, line 22: `mkdir` is not required as the `posix-oneshot` binary will initialize the directory.

```sh
➜  tessera git:(main) go run ./cmd/examples/posix-oneshot --storage_dir=${LOG_DIR} --entries="/tmp/stuff/*"
I1005 20:39:03.655774 1441449 files.go:516] Initializing directory for POSIX log at "/tmp/tlog" (this should only happen ONCE per log!)
I1005 20:39:03.857020 1441449 main.go:121] 0: /tmp/stuff/bar
I1005 20:39:03.857050 1441449 main.go:121] 1: /tmp/stuff/baz
I1005 20:39:03.857055 1441449 main.go:121] 2: /tmp/stuff/foo
```

cmd/examples/posix-oneshot/README.md
```diff
@@ -17,6 +17,10 @@ echo "foo" > /tmp/stuff/foo
 echo "bar" > /tmp/stuff/bar
 echo "baz" > /tmp/stuff/baz
 
+# Make a blank dir where we'll store the log data
+export LOG_DIR="/tmp/tlog"
```
mhutchinson, line 21: For a real usage of this tool, the log directory is really important. Burying the setting of it into the middle of the script doesn't do it justice.

I suggest either moving this to the first line of the script, or we can do something like the following to preserve any env variable the user explicitly sets:

```suggestion
LOG_DIR=${LOG_DIR:-/tmp/tlog}
```

mhutchinson (commented):
Thanks for this! Sorry (not sorry) for making it so you have a suggestion on each of your proposed additions :-)

mhutchinson (approved):

Issue #817, opened 2025-11-24
Flaky test-mysql check in pull requests


roger2hk:
**Description**

The recent test-mysql check has been very flaky in the pull requests.

PR #821, opened 2025-12-02
Fix flaky test-mysql check

roger2hk:
Fix https://github.com/transparency-dev/tessera/issues/817.


roger2hk:
```
W1202 17:31:41.942767    5597 mysql.go:112] publishCheckpoint: readTreeState: Error 1205 (HY000): Lock wait timeout exceeded; try restarting transaction
```

roger2hk:
I think I figured it out. The `TestAppend` completed first without waiting for checkpoint publication, and then the `TestGetTile` kicked in to poll the checkpoint every 10ms locking the checkpoint. That's why the checkpoint can never be published.

Issue #855, opened 2026-02-05
Feature: Support tlog-tiles pruning


mhutchinson:
This feature request is a demand tracker for users looking for support in Tessera for tlog-tiles pruning:

https://github.com/C2SP/C2SP/blob/main/tlog-tiles.md#pruning

Give this comment a 👍 if you want support for this feature, and this helps us prioritise.

PR #935, opened 2026-04-15
[Pruning] Added a design doc for tlog-tiles pruning

mhutchinson:
This is a design for #855. If accepted, we can move on to implementation.

### DIFF
--- a/docs/design/pruning.md
+++ b/docs/design/pruning.md
@@ -11,10 +11,9 @@ This document outlines the design for adding pruning support to Tessera, adherin
 The [tlog-tiles] spec (revision `1cd32db`) defines the rules for pruning but leaves operational details as an exercise for the implementation:
 
 ### What the Spec Defines
-*   **Minimum Index**: The log maintains a `minimum index`. Resources (tiles and entry bundles) whose *end index* is below or equal to the `minimum index` may be pruned.
+*   **Minimum Index**: The log maintains a minimum index which represents the inclusive lower bound of entries it makes available. Resources whose entries fall entirely below this bound may be pruned. This applies to entry bundles, and all tiles (level 0 and higher).
 *   **Safety Recommendation**: The spec recommends denying HTTP requests for pruned resources rather than deleting them from storage, allowing for recovery and support for slow witnesses.
 *   **Tree Invariants**: Pruning does not change the tree structure or invalidate existing proofs.
-*   **Higher-Level Tiles**: The rule applies generically to all resources. A tile at level $L$ covering entries up to index $E$ is only prunable when $E \le \text{minimum index}$. Consequently, higher-level tiles are pruned less frequently (e.g., in chunks of 65,536 for Level 1).
 
 ### Ambiguities and Gaps
 *   **Discovery**: The spec has a `TODO` regarding how clients discover the current `minimum index`.
@@ -28,9 +27,9 @@ Tessera adopts a pragmatic, operator-driven approach to pruning, avoiding the ne
 Pruning is not an automated background process in Tessera. Instead, it is an explicit action performed by the operator using a CLI tool. This minimizes complexity in the core library and ensures operators are aware of the impact.
 
 The tool performs the following steps:
-1.  **Validation**: Ensures the target index $N$ is less than the current checkpoint size, greater than or equal to the current `min_index`, and enforces that $N$ is a multiple of 256 (to align with tile boundaries).
-2.  **State Commitment**: Writes the new $N$ to a statically served `min_index` file at the root of the log.
-3.  **Archiving**: Identifies tiles and entry bundles whose end index is $\le N$ and moves them to an archive location.
+1.  **Validation**: Ensures the new minimum index `N` is less than the current checkpoint size, greater than or equal to the current `min_index`, and enforces that `N` is a multiple of 256 (to align with tile boundaries).
+2.  **State Commitment**: Writes the new `N` to a statically served `min_index` file at the root of the log.
+3. **Archiving**: Identifies tiles and entry bundles whose entries fall entirely below `N` (i.e., their exclusive end index is `<= N`) and moves them to an archive location.
 
 The operation is idempotent and can be safely rerun to continue a failed or interrupted operation.
 
@@ -42,9 +41,8 @@ To address the spec's ambiguity on discovery, Tessera introduces a static `min_i
 ### 3. Archiving Strategy
 To fulfill the spec's recommendation of not deleting data, the pruning tool moves files to an archive location while maintaining the exact same directory structure. This allows the archive to be used directly as a fallback log or for manual recovery.
 
-*   **POSIX**: Files are moved to a parallel `archive/` directory that is not exposed by the web server serving the log.
-*   **Cloud (GCS/AWS)**: Files are moved to a separate, private bucket (e.g., `[log-bucket]-archive`). This ensures that direct reads to the public bucket return a native `404` without requiring a load balancer or proxy.
-    *   **GCS "Brave Mode" Alternative**: For GCS deployments, the tool could support a mode that simply deletes the files from the public bucket. By relying on GCS's default **Soft Delete** feature, operators retain a time-limited recovery window (default 7 days) to undo errors, after which the files are permanently deleted. This avoids the cost and overhead of maintaining a perpetual archive bucket.
+*   **POSIX**: Files are moved to an archive directory that is not exposed by the web server serving the log. This directory would be flag-configured.
+*   **Cloud (GCS/AWS)**: Files are moved to a separate, private bucket (e.g., `[log-bucket]-archive`). This ensures that direct reads to the public bucket return a native `404` without requiring a load balancer or proxy. See Alternatives Considered for an approach that archives within the same bucket.
 *   **Ordering**: To maintain consistency with Tessera's existing Garbage Collection mechanism (which cleans up obsolete partial tiles), the tool should process resources in a bottom-up, left-to-right order: first entry bundles, then Level 0 tiles, and finally higher-level tiles. Within each resource type, it should proceed from the lowest index to the highest.
 
 ### 4. Safety Features
@@ -63,13 +61,16 @@ We considered using lightweight compute at the edge (e.g., Cloudflare Workers, L
 
 *   **Why it was not picked**: This approach introduces significant infrastructure complexity. It requires logs to be served behind a Load Balancer or CDN rather than allowing direct access to object storage buckets (like GCS). This violates Tessera's goal of having "no additional services to manage" and maintaining a low TCO for operators.
 
-### GCS Object Versioning or Soft Delete
-We considered relying on cloud-specific features like GCS Object Versioning or Soft Delete. Under this model, the tool would simply delete the files from the public bucket. If Object Versioning is enabled, the deleted files would become noncurrent versions (returning `404` to public requests) and could be restored later.
+### GCS Object Versioning
+We considered relying on cloud-specific features like GCS Object Versioning. Under this model, the tool would simply delete the files from the public bucket. If Object Versioning is enabled, the deleted files would become noncurrent versions (returning `404` to public requests) and could be restored later.
 
 *   **Why it was not picked**: 
-    *   **Cost Inefficiency**: Tessera frequently updates partial tiles and entry bundles during normal operation. If Object Versioning were enabled on the bucket, every such update would create a new billable version, leading to uncontrolled storage cost growth unless complex lifecycle rules were configured to clean them up. Furthermore, if versioning is enabled, Tessera's existing Garbage Collection mechanism (which deletes obsolete partial tiles) becomes ineffective at saving quota, as the deleted files are merely retained as noncurrent versions.
+    *   **Cost Inefficiency**: Object Versioning is not recommended for Tessera. Tessera writes many partial tiles and entry bundles during normal operation, and enabling Object Versioning would cause Tessera's existing Garbage Collection mechanism (which deletes obsolete partial tiles) to become ineffective at saving quota, as the deleted files would be retained as noncurrent versions.
     *   **Portability**: This approach relies on specific cloud provider features that do not translate well to simple POSIX filesystems or other cloud providers with different versioning semantics. The "move to archive" approach is universally applicable across all storage drivers supported by Tessera.
 
-*   **Note**: While we rejected Object Versioning for the reasons above, a variant relying solely on **Soft Delete** (without versioning enabled) is considered as a viable option for GCS deployments (see the **GCS "Brave Mode" Alternative** in the Archiving Strategy section).
+### GCS "Brave Mode" (Delete with Soft Delete)
+We considered a mode for GCS deployments where the tool simply deletes the files from the public bucket instead of moving them to an archive. By relying on GCS's default **Soft Delete** feature, operators retain a time-limited recovery window (default 7 days) to undo errors, after which the files are permanently deleted.
+
+*   **Why it was not picked**: While this avoids the cost and overhead of maintaining a perpetual archive bucket, it violates the spec's recommendation to not delete data from storage. Additionally, files retained by Soft Delete would still incur costs during the retention window, potentially interfering with the cost-saving goals of garbage collecting partial tiles (though only for a limited time).
 
 [tlog-tiles]: https://c2sp.org/tlog-tiles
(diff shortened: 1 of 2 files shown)


docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+*   **Ordering**: To maintain consistency with Tessera's existing Garbage Collection mechanism (which cleans up obsolete partial tiles), the tool should process resources in a bottom-up, left-to-right order: first entry bundles, then Level 0 tiles, and finally higher-level tiles. Within each resource type, it should proceed from the lowest index to the highest.
+
+### 4. Safety Features
+*   **Confirmation Prompt**: The tool scans the range to be pruned and prompts the operator with the count of files (tiles and entry bundles) to be moved before proceeding.
+*   **Undo Command**: The tool provides an `undo` command that moves files back from the archive to the active area and restores the `min_index` file to its previous value (by storing the previous value in the archive during pruning), facilitating recovery from operator error. In the event of manual recovery, the operator can manually update the `min_index` file based on the data they have restored.
```
roger2hk, line 50: Would this `undo` command work on the GCS "Brave Mode" assuming within the recovery window?

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+The [tlog-tiles] specification introduces an optional mechanism to "prune" logs, allowing logs to make a prefix of the log unavailable while maintaining the Merkle tree structure and validity of existing proofs.
+
+This document outlines the design for adding pruning support to Tessera, adhering to the principles of simplicity and low total cost of ownership (TCO).
+
+## Spec Requirements & Ambiguities
+
+The [tlog-tiles] spec (revision `1cd32db`) defines the rules for pruning but leaves operational details as an exercise for the implementation:
+
+### What the Spec Defines
+*   **Minimum Index**: The log maintains a `minimum index`. Resources (tiles and entry bundles) whose *end index* is below or equal to the `minimum index` may be pruned.
```
AlCutter: I think this could do with being clearer - my immediate reaction to reading this is that it's wrong because `minimum index` is required to continue to be present in the log. It _could_ be that the text is correct as-is because `end index` of a resource is assumed to be the open end of a closed-open range, but the text doesn't say.

Maybe something along the lines of:

> The log maintains a `minimum index` which represents the inclusive lower bound of entries it makes available. Resources whose entries fall *entirely* below this bound may be pruned.
?

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+
+Transparency logs are append-only data structures that grow indefinitely. For high-volume logs, such as Certificate Transparency, this can lead to significant storage costs and overhead for both operators and monitors.
+
+The [tlog-tiles] specification introduces an optional mechanism to "prune" logs, allowing logs to make a prefix of the log unavailable while maintaining the Merkle tree structure and validity of existing proofs.
+
+This document outlines the design for adding pruning support to Tessera, adhering to the principles of simplicity and low total cost of ownership (TCO).
+
+## Spec Requirements & Ambiguities
+
+The [tlog-tiles] spec (revision `1cd32db`) defines the rules for pruning but leaves operational details as an exercise for the implementation:
```
AlCutter, line 11: Image

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+
+### What the Spec Defines
+*   **Minimum Index**: The log maintains a `minimum index`. Resources (tiles and entry bundles) whose *end index* is below or equal to the `minimum index` may be pruned.
+*   **Safety Recommendation**: The spec recommends denying HTTP requests for pruned resources rather than deleting them from storage, allowing for recovery and support for slow witnesses.
+*   **Tree Invariants**: Pruning does not change the tree structure or invalidate existing proofs.
+*   **Higher-Level Tiles**: The rule applies generically to all resources. A tile at level $L$ covering entries up to index $E$ is only prunable when $E \le \text{minimum index}$. Consequently, higher-level tiles are pruned less frequently (e.g., in chunks of 65,536 for Level 1).
```
AlCutter: Probably better to just use `<=` rather than embedding `LaTeX` in here for this.

Can also remove `L` as it's written, maybe just say something like "all tiles exclusively covering entries strictly below `minimum_index` are eligible for pruning"?

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Design
+
+Tessera adopts a pragmatic, operator-driven approach to pruning, avoiding the need for complex runtime infrastructure on the read path.
+
+### 1. Explicit Operator Action (CLI Tool)
+Pruning is not an automated background process in Tessera. Instead, it is an explicit action performed by the operator using a CLI tool. This minimizes complexity in the core library and ensures operators are aware of the impact.
+
+The tool performs the following steps:
+1.  **Validation**: Ensures the target index $N$ is less than the current checkpoint size, greater than or equal to the current `min_index`, and enforces that $N$ is a multiple of 256 (to align with tile boundaries).
```
AlCutter: LaTeX -> "`...`" throughout for readability, unless there's some actual hard to represent maths.

AlCutter: target index --> new minimum index?

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+2.  **State Commitment**: Writes the new $N$ to a statically served `min_index` file at the root of the log.
```
AlCutter: This feels like it should be part of the state, next to the other internal coordination state rather than public.

The mechanism for this info to be made available to clients is (currently) almost certainly going to be application specific since there's no spec for how to represent this, so I'd be tempted to leave the decision about whether to make custom off-spec resources and the format they're in up to them for the moment.

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+3.  **Archiving**: Identifies tiles and entry bundles whose end index is $\le N$ and moves them to an archive location.
+
+The operation is idempotent and can be safely rerun to continue a failed or interrupted operation.
```
AlCutter, line 34: With the steps above, you don't have enough state to know whether subsequent updates to `min_index` need to examine just the resources which cover `[old_min, new_min)` or the whole `[0, new_min)` shebang - you've written the new min state before you attempt to archive resources, which could potentially fail.

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+
+### 2. State File (`min_index`)
+To address the spec's ambiguity on discovery, Tessera introduces a static `min_index` file served at the root of the log (e.g., `<prefix>/min_index`), similar to the `checkpoint` file.
+*   **Format**: A simple text file containing the decimal representation of the minimum index.
+*   **Serving**: Clients can fetch this file to understand the current valid range of the log.
+
+### 3. Archiving Strategy
+To fulfill the spec's recommendation of not deleting data, the pruning tool moves files to an archive location while maintaining the exact same directory structure. This allows the archive to be used directly as a fallback log or for manual recovery.
+
+*   **POSIX**: Files are moved to a parallel `archive/` directory that is not exposed by the web server serving the log.
```
AlCutter: This feels a bit cheeky - breaking out of the location the operator intended for the operation of the log. E.g I could imagine oprs having created independent zfs filesystems for each log so as to provide isolation, then mount these as the root of the log. 

We should probably at least allow the location to be set via a flag, but could also consider just defaulting to creating `./archive/` (or even `./.archive/`) to avoid surprising the opr. It should be easy enough to exclude that directory from an nginx or caddy config.

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+*   **Cloud (GCS/AWS)**: Files are moved to a separate, private bucket (e.g., `[log-bucket]-archive`). This ensures that direct reads to the public bucket return a native `404` without requiring a load balancer or proxy.
```
AlCutter: Similar to above, did you consider just renaming under an `archive/` (or `.archive/`) prefix? In GCS at least, renames within a bucket are atomic.

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+    *   **GCS "Brave Mode" Alternative**: For GCS deployments, the tool could support a mode that simply deletes the files from the public bucket. By relying on GCS's default **Soft Delete** feature, operators retain a time-limited recovery window (default 7 days) to undo errors, after which the files are permanently deleted. This avoids the cost and overhead of maintaining a perpetual archive bucket.
```
AlCutter: Surely "Brave Mode" would just hard delete off the bat :)

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+*   **Ordering**: To maintain consistency with Tessera's existing Garbage Collection mechanism (which cleans up obsolete partial tiles), the tool should process resources in a bottom-up, left-to-right order: first entry bundles, then Level 0 tiles, and finally higher-level tiles. Within each resource type, it should proceed from the lowest index to the highest.
+
+### 4. Safety Features
+*   **Confirmation Prompt**: The tool scans the range to be pruned and prompts the operator with the count of files (tiles and entry bundles) to be moved before proceeding.
```
AlCutter, line 49: Is it necessary to actually list the objects in order to count them, or would it suffice to just calculate the lower bound on the number of resources given the range?

Personally, I think that knowing there are x thousand logical tile/bundle addresses to be moved is more useful than xxx thousand partial and full revisions of some unknown number of logical addresses to be moved.

docs/design/pruning.md
AlCutter, line 50: How does this interact with GC?

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Alternatives Considered
+
+### Archive-in-Place (Renaming Files)
+We considered keeping archived files in the same storage location (bucket or directory) but renaming them (e.g., appending `~` or moving to a sub-folder like `archive/` within the same bucket).
+
+*   **Why it was not picked**: While this makes the "move" operation fast and atomic (metadata rename), it creates security risks for cloud storage users. If a GCS bucket is served publicly with uniform bucket-level access, everything in that bucket is public. Hiding an `archive/` folder or renamed files would require complex IAM conditions or putting a proxy/load balancer in front of the bucket. This violates Tessera's goal of zero-infrastructure read paths. Moving to a separate private bucket provides a much stronger security boundary with less operational complexity.
```
AlCutter, line 57: You could also just accept that motivated folks would still be able to access archived resources via the (unspecified) archive path?

I'm not sure I understand why a strong security boundary for archived resources is important.

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+
+### GCS Object Versioning or Soft Delete
+We considered relying on cloud-specific features like GCS Object Versioning or Soft Delete. Under this model, the tool would simply delete the files from the public bucket. If Object Versioning is enabled, the deleted files would become noncurrent versions (returning `404` to public requests) and could be restored later.
+
+*   **Why it was not picked**: 
+    *   **Cost Inefficiency**: Tessera frequently updates partial tiles and entry bundles during normal operation. If Object Versioning were enabled on the bucket, every such update would create a new billable version, leading to uncontrolled storage cost growth unless complex lifecycle rules were configured to clean them up. Furthermore, if versioning is enabled, Tessera's existing Garbage Collection mechanism (which deletes obsolete partial tiles) becomes ineffective at saving quota, as the deleted files are merely retained as noncurrent versions.
```
AlCutter: I'm not arguing that object versioning is the right approach at all, but the assertion here about Tessera frequently updating partial tiles and entry bundles is... not true?

Tessera only ever re-writes a tile/entrybundle resource if there was an unrecoverable error during integration, which is relatively rare compared to the number of resources written overall.

And GC _could_, of course, just explicitly delete the object plus all of its (almost certainly 1) versions.

docs/design/pruning.md
```diff
@@ -0,0 +1,75 @@
...
+    *   **Portability**: This approach relies on specific cloud provider features that do not translate well to simple POSIX filesystems or other cloud providers with different versioning semantics. The "move to archive" approach is universally applicable across all storage drivers supported by Tessera.
+
+*   **Note**: While we rejected Object Versioning for the reasons above, a variant relying solely on **Soft Delete** (without versioning enabled) is considered as a viable option for GCS deployments (see the **GCS "Brave Mode" Alternative** in the Archiving Strategy section).
```
AlCutter: This seems inconsistent with the opening para in this section, would it make sense to remove Soft Delete entirely here and just leave this dismissal for Object Versioning?

docs/design/pruning.md
mhutchinson: Yes, if we choose to implement the brave mode, then we should implement the undo for it. At least, that's the proposal. We could choose to not do undo at all and make it the operator's concern, but that feels... concerning.

docs/design/pruning.md
mhutchinson: I think the previous version was correct as per your observation about end-index is defined. But used your suggestion.

docs/design/pruning.md
mhutchinson: I've rolled this into the top bullet point.

docs/design/pruning.md
mhutchinson: Yeah, fair.

mhutchinson: Done.

docs/design/pruning.md
mhutchinson: That was my initial thought, but the spec explicitly has this TODO:

TODO: Some HTTP endpoint for fetching the minimum index? The semantics would be something like: serving a minimum index equivalent to returning 404 from the tiles that would be deleted by the pruning criteria, including when evaluating a log client's availability policies.

If logs are going to do pruning, it seems like a good idea that they at least advertise what they believe they are running with. Even if this isn't cryptographically committed to. Really we need the spec to clarify this, and so I proposed this as a forcing function.

docs/design/pruning.md
mhutchinson: I chose this ordering because the `min_index` file is public, and it seems better to say "hey everyone, don't expect anything below N" _before_ we start pruning. If we agree that state isn't public, then writing this after the pruning finishes makes more sense.

Even in the current configuration, we can still identify and prune all of the entries from `[0, new_min)`. We can also get the log operator to provide a different `old_min` for performance gains.

docs/design/pruning.md
mhutchinson: Yeah, this wasn't intended to be completely literal. But I enclosed it in literal quotes, so that's on me. Updated to say it would be a flag-configured directory. PTAL?

docs/design/pruning.md
mhutchinson: This is one of the alternatives considered. Pointed at that.

docs/design/pruning.md
mhutchinson: "Brave (but not stupid) mode" just seemed a bit long? :-)

docs/design/pruning.md
mhutchinson: I don't much mind TBH. I guess scanning the range is expensive so this needs to be reconsidered. Leaving this comment open.

docs/design/pruning.md
mhutchinson: The undo feature? I had assumed that by the time one prunes a tree, it would be thoroughly GCd. Is it a supported configuration to have a no-GC deployment though?

docs/design/pruning.md
mhutchinson: It's not a strong security boundary. It's:
 a) avoiding polluting the serving namespace, e.g. your comment about the `min_index` file
 b) a minor concern about any edge caches also caching this stuff if requested.

If we're strongly opinionated that putting this stuff on the public paths is cool, we can consider changing the primary suggestion for these alternatives.

docs/design/pruning.md
mhutchinson: Yeah good catch. Updating is the wrong term. This was going in a few different directions and ended up botched. Tried to clarify what was intended: object versioning is not recommended because GC cleans up a lot of stuff. We don't want to start recommending object versioning for this feature, if that then makes GC harder.

docs/design/pruning.md
mhutchinson: Actually, I moved "brave mode" down to an alternative considered. We shouldn't be messing about with multiple implementations of this. Let's pick one.

mhutchinson (commented):
Thanks for the comments. Addressed many of them.

PR #938, opened 2026-04-24
Update shutdown process to ensure followers catch up to the target index

roger2hk:


append_lifecycle.go
```diff
@@ -525,9 +535,10 @@ func (t *terminator) Add(ctx context.Context, entry *Entry) IndexFuture {
 	}
 }
 
-// Shutdown ensures that all calls to Add that have returned a value will be resolved. Any
-// futures returned by _this appender_ which resolve to an index will be integrated and have
-// a checkpoint that commits to them published if this returns successfully.
+// Shutdown ensures that all calls to Add that have returned a value will be resolved
+// and that the followers are caught up to the largest index.
```
AlCutter: Maybe move the bit about followers down past where it talks about all things having been integrated (and checkpoint published)?

Most (all?) followers operate on entries from the integrated tree, so it'll kinda follow more naturally when reading if we've already made it clear that all the entries successfully assigned indices will have already been integrated.

append_lifecycle.go
```diff
@@ -539,6 +550,12 @@ func (t *terminator) Shutdown(ctx context.Context) error {
 		// special case no work done
 		return nil
 	}
+
+	ctx, cancel := context.WithTimeout(ctx, t.shutdownTimeout)
+	defer cancel()
+
+	checkpointDone := false
+	followersDone := make([]bool, len(t.followers))
```
AlCutter: Would this be more natural with an `ErrorGroup.WithContext`, and a goroutine for each things we're blocking on - perhaps one for checkpoint, and one for each follower?

roger2hk: Updated to have one goroutine for each things (checkpoint and each follower).

append_lifecycle.go
roger2hk: Done.

append_lifecycle.go
```diff
@@ -551,63 +552,63 @@ func (t *terminator) Shutdown(ctx context.Context) error {
...
+			case <-gCtx.Done():
+				return gCtx.Err()
+			default:
+			}
+			cp, err := t.readCheckpoint(gCtx)
 			if err != nil {
 				if !errors.Is(err, os.ErrNotExist) {
 					return err
 				}
+				time.Sleep(pollInterval)
```
AlCutter: Could pull these sleeps up into a `Ticker` in the `select` at the top - if it's important to avoid the initial delay before the first check you can always initialise the Ticker with a low first duration, and then `Reset()` with the actual poll interval each time you successfully read from the `.C` channel.

AlCutter (approved):

append_lifecycle.go
roger2hk: Done. Updated to use ticker with 1ns and then `Reset()` to 100ms.

PR #940, opened 2026-04-29
Improve checkpoint publication scheduling

phbnf:
This PR schedules checkpoint publication operations as early as possible, as per `checkpoint_interval and the last publication timestamp.

### DIFF
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -75,7 +75,7 @@ const (
 	// GCS has a rate limit 1 update per second for individual objects, but we've observed that attempting
 	// to update at exactly that rate still results in the occasional refusal, so bake in a little wiggle
 	// room.
-	minCheckpointInterval = 1200 * time.Millisecond
+	minCheckpointInterval = 1100 * time.Millisecond
 
 	logContType      = "application/octet-stream"
 	ckptContType     = "text/plain; charset=utf-8"
--- a/storage/gcp/gcp.go
+++ b/storage/gcp/gcp.go
@@ -388,12 +388,8 @@ func (a *Appender) publishCheckpointJob(ctx context.Context, pubInterval, republ
 			if err != nil {
 				return fmt.Errorf("publishCheckpoint failed: %v", err)
 			}
-			nextPublication := pubInterval - time.Since(publishedAt)
-			if nextPublication <= 0 {
-				t.Reset(time.Millisecond) // Schedule a checkpoint update immediately.
-			} else {
-				t.Reset(nextPublication)
-			}
+			// Schedule a checkpoint update immediately, if an updated is due.
+			t.Reset(max(time.Millisecond, pubInterval-time.Since(publishedAt)))
 			return nil
 		}, trace.WithAttributes(otel.PeriodicKey.Bool(true))); err != nil {
 			t.Reset(pubInterval)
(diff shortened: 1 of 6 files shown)


AlCutter:
I'm wondering if we can do this with no visible API changes...

Inside each of the publishCheckpoint routines, rather than polling frequently, we _could_ instead do a better job of scheduling attempts so they have more chance to succeed; e.g. when we check when the last CP was published and discover that it's too new, we now know exactly what the interval must be before the next one can be published. 

Would something along the lines of the sketch below work?
```go
const retry := 200*time.Milliseconds // whatever
t := time.NewTicker(retry)

for {
   select {
     case <-ctx.Done(): 
       // whatever
     case <-t.C:
   }
   
   checkpointAge := ...
   if !shouldPublish {
     t.Reset(minInterval-checkpointAge)
     continue
   }
   if err := publishCheckpoint(); err != nil {
     t.Reset(retry)
     continue
   }
   t.Reset(minInterval)
}
```

phbnf:
Ah yes 100%, that was my original design but I had understood you didn't want do to that, sorry. I'll get back to it.

AlCutter:
> Ah yes 100%, that was my original design but I had understood you didn't want do to that, sorry. I'll get back to it.

Ok, well great!
I don't think we talked about it when we were braining on this together yesterday, but either way - great minds - carry on! :)

phbnf:
I've done something along those lines, have a look. Instead of fetching the last checkpoint update time in `publishCheckpointJob` and thus doing extra reads, I've modified the signature of `publishCheckpoint`. Happy to change the implementation if you'd prefer something else. 

I have not implemented a custom retry interval, I think we can just default back to the user provided value, it keeps things simpler.

I also took that as an opportunity to decrease `minCheckpointInterval` for GCP.

storage/posix/files.go
```diff
@@ -179,11 +179,19 @@ func (a *appender) publishCheckpointJob(ctx context.Context, pubInterval, republ
...
+			publishedAt, err := a.publishCheckpoint(ctx, pubInterval, republishInterval)
+			if err != nil {
 				return err
 			}
+			nextPublication := pubInterval - time.Since(publishedAt)
+			if nextPublication <= 0 {
+				t.Reset(time.Millisecond) // Schedule a checkpoint update immediately.
+			} else {
+				t.Reset(nextPublication)
+			}
```
AlCutter: Could the 3 instances of this be simplified to `t.Reset(max(time.Millisecond, pubInterval - time.Since(publishedAt)))`?

storage/gcp/gcp.go
```diff
@@ -1086,20 +1082,23 @@ func (s *spannerCoordinator) nextIndex(ctx context.Context) (uint64, error) {
...
+func (s *spannerCoordinator) publishCheckpoint(ctx context.Context, minStaleActive, minStaleRepub time.Duration, f func(context.Context, uint64, []byte) error) (time.Time, error) {
+	var pubAt time.Time
 	currentSize, rootHash, err := s.currentTree(ctx)
 	if err != nil {
-		return fmt.Errorf("failed to get current tree: %v", err)
+		return time.Time{}, fmt.Errorf("failed to get current tree: %v", err)
 	}
 
-	return otel.TraceErr(ctx, "tessera.storage.gcp.publishCheckpoint", tracer, func(ctx context.Context, span trace.Span) error {
+	if err := otel.TraceErr(ctx, "tessera.storage.gcp.publishCheckpoint", tracer, func(ctx context.Context, span trace.Span) error {
```
AlCutter: You could change `otel.TraceErr` to `otel.Trace2` and return `(time.Time, error)` from the inner func directly to reduce the readability hit from adding the outer `if` here. (same in other implementations)

AlCutter (approved):

AlCutter (approved):

PR #941, opened 2026-04-30
Notify integration when sequencing is done

phbnf:


storage/aws/aws.go
```diff
@@ -235,12 +235,22 @@ func (s *Storage) newAppender(ctx context.Context, o objStore, seq sequencer, op
...
-		sequencer:   seq,
-		queue:       storage.NewQueue(ctx, opts.BatchMaxAge(), opts.BatchMaxSize(), seq.assignEntries),
-		newCP:       opts.CheckpointPublisher(logStore, s.cfg.HTTPClient),
-		treeUpdated: make(chan struct{}),
-	}
+		logStore:        logStore,
+		sequencer:       seq,
+		newCP:           opts.CheckpointPublisher(logStore, s.cfg.HTTPClient),
+		treeUpdated:     make(chan struct{}),
+		entriesAssigned: make(chan struct{}),
```
AlCutter: _Might_ be worth making `entriesAssigned` have a capacity of `1` - as written this will only trigger an integration run IFF the integration job is blocked in the `select{}`, otherwise it'll fall back to checking after 1s (or whatever).

I'm not entirely sure what fraction of time the job will be sleeping vs working, but this could translate into more "lumpy" apparent integration duration.

Setting the channel size to 1 OTOH would mean that there can be a "work to do" signal waiting for the job, in effect causing it to fast loop. I suspect this will give a more even apparent integration delay, at the cost of potentially a little more busy work. I suspect this is the better trade off, but open to being convinced otherwise!

storage/gcp/gcp.go
```diff
@@ -264,10 +264,20 @@ func (s *Storage) newAppender(ctx context.Context, o objStore, seq *spannerCoord
 			objStore:    o,
 			entriesPath: opts.EntriesPath(),
 		},
-		sequencer: seq,
-		cpUpdated: make(chan struct{}),
+		sequencer:       seq,
+		cpUpdated:       make(chan struct{}),
+		entriesAssigned: make(chan struct{}),
```
AlCutter: (same here)

AlCutter (approved):

storage/aws/aws.go
phbnf: Done. I'm also OOO tomorrow, so I'll leave the PR as it is, and you can merge it if you'd like. I have not looked at the feasibility to add tests.

AlCutter (approved):

phbnf:
Tests failed because of an unrelated race condition, which I'm fixing in https://github.com/transparency-dev/tessera/pull/944.

Issue #945, opened 2026-05-06
Add support for MTC logs & mirrors


AlCutter:
Merkle Tree Certificates represents the continued evolution of Certificate Transparency logs. It's based on the `tlog-.*` C2SP specs, and Tessera should support the creation of Merkle Tree Certficate logs and mirrors.

PR #946, opened 2026-05-06
[MTC] Add first cut of mirror HTTP handler

AlCutter:
This PR adds a very rough sketch for an MTC mirror API server, primarily intended to enable work on the lower storage layers, subtree proofs, and test client to proceed.

Towards #945

### DIFF
--- a/cmd/mtc/mirror/internal/handler/handler.go
+++ b/cmd/mtc/mirror/internal/handler/handler.go
@@ -1,0 +1,289 @@
+// Copyright 2026 The Tessera authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package handler
+
+import (
+	"bufio"
+	"context"
+	"encoding/base64"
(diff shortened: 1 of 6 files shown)


cmd/mtc/mirror/internal/mirror/mirror.go
```diff
@@ -0,0 +1,56 @@
...
+type Mirror struct {
+}
+
+// Package represents a single package of entries and its subtree consistency proof.
+type Package struct {
+	Entries [][]byte
+	Proof   [][]byte
+}
+
+func (m *Mirror) AddCheckpoint(ctx context.Context, cp []byte, oldSize uint64, proof [][]byte) error {
```
phbnf: Nit: old size, proof, and then a cp to match with https://github.com/C2SP/C2SP/blob/main/tlog-witness.md#add-checkpoint.

cmd/mtc/mirror/internal/handler/handler.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// parseAddEntriesPreamble consumes the body of the Add-Entries request.
+//
+// This func effectively takes ownership of the provided Reader, it MUST NOT be
+// further used by the caller.
+//
+// Returns a struct which represents the request, and provides streaming access to the entry packages.
+func parseAddEntriesPreamble(r io.Reader) (*addEntriesRequest, error) {
+	//SPEC: The request body MUST have Content-Type of application/octet-stream and contain the
```
phbnf: Should we check for this?

phbnf (approved):

cmd/mtc/mirror/internal/mirror/mirror.go
AlCutter: Good call

cmd/mtc/mirror/internal/handler/handler.go
AlCutter: Added, up at the handler.

PR #948, opened 2026-05-08
Refactor the queue to be more idiomatic

AlCutter:
This PR refactors the internals of the queue to be more idiomatic, swapping mutexes for channels.

Other than being easier to read and reason about, this change also prevents rare but potentially surprising behaviour whereby two batches could be sequenced in an unexpected ordering. This would not have been fatal for a log, but is better avoided.

### DIFF
--- a/storage/internal/queue.go
+++ b/storage/internal/queue.go
@@ -59,6 +59,9 @@ func NewQueue(ctx context.Context, maxAge time.Duration, maxSize uint, f FlushFu
 		defer close(batches)
 
 		var items []queueItem
+
+		// Initialise a timer with an arbitrarily large value, and immediately
+		// call Stop() to avoid a spurious trigger on the first iteration.
 		timer := time.NewTimer(time.Hour)
 		timer.Stop()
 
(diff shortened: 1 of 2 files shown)


storage/internal/queue.go
```diff
@@ -57,82 +49,91 @@ type FlushFunc func(ctx context.Context, entries []*tessera.Entry) error
...
+
+	// Spin off a goroutine which accumulates added items into batches and sends them to the flush worker
+	// via the batches channel.
+	go func() {
+		defer close(batches)
 
-	// Spin off a worker thread to write the queue flushes to storage.
-	go func(ctx context.Context) {
+		var items []queueItem
+		timer := time.NewTimer(time.Hour)
```
phbnf, line 65: Is this to initialize the timer variable without making it trigger anything? Maybe leave a comment to explain this?

phbnf:
Interesting, we actually spoke about this yesterday in the RB and quickly glancing at the code, I came to the conclusion that out of order flushing could not happen specifically because of the lock... and I was wrong: so +1 for increased readability!

phbnf (approved):

storage/internal/queue.go
AlCutter: Yup, and good call - done.

PR #949, opened 2026-05-08
Fix terminator bug

AlCutter:
This PR fixes an issue with the shutdown terminator in the case where a single first entry has just been added to the log.

Previously, it didn't distinguish between nothing added, and a zeroth entry added.
With this change it now waits for a minimum tree size which covers all locally added entries, which may be zero if no entries were added.

### DIFF
--- a/append_lifecycle_test.go
+++ b/append_lifecycle_test.go
@@ -16,6 +16,8 @@ package tessera
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


append_lifecycle_test.go
```diff
@@ -150,22 +152,70 @@ func mustCreateSigner(t *testing.T, k string) note.Signer {
...
 		},
 	}
-	term.largestIssued.Store(1)
-	term.shutdownTimeout = 10 * time.Millisecond
 
-	err := term.Shutdown(context.Background())
-	if err == nil {
-		t.Fatal("Expected timeout error, got nil")
+	f := term.Add(t.Context(), nil)
+  if	_, err := f(); err != nil {
```
phbnf: I think indentation is off.

phbnf (approved):

PR #958, opened 2026-05-13
Refactor signer name consistency enforcement in AppendOptions validation

roger2hk:
This pull request defers the signer validation in `WithCheckpointSigner` to avoid `os.Exit` and `context.Background`.

**Problem**
The WithCheckpointSigner method was performing immediate validation on the provided signers. However, this method is a configuration setter and does not take a `context.Context`. Only `context.Background()` can be used to log errors when validation failed. Furthermore, it called `os.Exit(1)` directly upon failure, which is anti-idiomatic for library code as it prevents the calling application from handling the failure gracefully.

**Solution**
This change defers the validation of additional signers to the `valid()` method of `AppendOptions`. If validation fails, `valid()` returns an error instead of crashing the process. This approach does not cause any breaking changes.

**Why better?**
- **Idiomatic Go Error Handling**: Libraries should return errors rather than calling `os.Exit`. This allows the application using Tessera to decide how to handle invalid configurations.
- **Eliminates dummy contexts**: By moving the check to `valid()` (which is called by `NewAppender` in a context-aware environment), we no longer need to use `context.Background()` for logging errors.

### DIFF
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -548,7 +548,7 @@ func (t *terminator) Shutdown(ctx context.Context) error {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	t.stopped = true
-	
+
 	wantSize := t.wantTreeSize.Load()
 	if wantSize == 0 {
 		// special case no work done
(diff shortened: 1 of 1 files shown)


AlCutter:
Could you add a description which explains _why_ this change is necessary/better?

roger2hk:
> Could you add a description which explains _why_ this change is necessary/better?

Sure, I've just updated the pull request description.

mhutchinson (approved):

PR #959, opened 2026-05-13
Prevent out-of-range slices in MigrationStorage

roger2hk:
**Problem**
If the bundle fetched from storage is corrupted, incomplete, or simply shorter than what `layout.Range` expects for that index, `bh` will have fewer elements than `ri.First + ri.N`. This will trigger a slice bounds out-of-range panic.

**Solution**
This pull request adds a bounds check in `fetchLeafHashes` across AWS, GCP, and POSIX backends to prevent a potential slice out-of-bounds panic when processing truncated entry bundles during log migration.

### DIFF
--- a/storage/aws/aws.go
+++ b/storage/aws/aws.go
@@ -629,6 +629,9 @@ func (m *MigrationStorage) fetchLeafHashes(ctx context.Context, from, to, source
 			if err != nil {
 				return fmt.Errorf("bundleHasherFunc for bundle index %d: %v", ri.Index, err)
 			}
+			if len(bh) < int(ri.First+ri.N) {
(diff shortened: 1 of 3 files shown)


storage/aws/aws.go
```diff
@@ -629,6 +629,9 @@ func (m *MigrationStorage) fetchLeafHashes(ctx context.Context, from, to, source
 			if err != nil {
 				return fmt.Errorf("bundleHasherFunc for bundle index %d: %v", ri.Index, err)
 			}
+			if len(bh) < int(ri.First+ri.N) {
```
AlCutter:
```
if l := len(bh); l < ... {
   return fmt.Errorf(..., ..., l, ...)
}
```

AlCutter (approved):

PR #964, opened 2026-05-21
Add WIP experimental tlog-mirror client

roger2hk:
This pull request is an experimental/WIP client implementation of the [tlog-mirror](https://c2sp.org/tlog-mirror) spec.

### DIFF
--- a/client/mirror/client.go
+++ b/client/mirror/client.go
@@ -156,6 +156,7 @@ func (e ErrConflict) Error() string {
 //   - The tree size of a valid pending checkpoint, in decimal
 //   - The next entry, in decimal
 //   - An opaque, possibly zero length, ticket value, encoded in base64
+// nolint:unused
 func parseConflict(r io.Reader) error {
 	// TODO(roger2hk): Implement this.
 
--- a/client/mirror/client.go
+++ b/client/mirror/client.go
@@ -155,12 +155,10 @@ func (e ErrConflict) Error() string {
 //   - The tree size of a valid pending checkpoint, in decimal
 //   - The next entry, in decimal
 //   - An opaque, possibly zero length, ticket value, encoded in base64
-//
-// nolint:unused
-func parseConflict(r io.Reader) (ErrConflict, error) {
+func parseConflict(r io.Reader) error {
 	// TODO(roger2hk): Implement this.
 
-	return ErrConflict{}, errors.New("TODO")
+	return errors.New("TODO")
 }
 
 // pushEntries streams entry packages and their proofs to the mirror's /add-entries endpoint.
@@ -191,11 +189,7 @@ func (c *Client) pushEntries(ctx context.Context, uploadStart, uploadEnd uint64,
 	}()
 
 	if resp.StatusCode == http.StatusConflict {
-		conflict, err := parseConflict(resp.Body)
-		if err != nil {
-			return fmt.Errorf("conflict returned, but failed to parse response body: %w", err)
-		}
-		return conflict
+		return parseConflict(resp.Body)
 	}
 
 	if resp.StatusCode != http.StatusOK {
--- a/client/mirror/client.go
+++ b/client/mirror/client.go
@@ -30,13 +30,9 @@ import (
 	"github.com/transparency-dev/tessera/client"
 )
 
-// PackageProver is an interface for generating proof hashes required for
-// the entry package covering the index interval [start, end).
-type PackageProver interface {
-	// PackageProof computes and returns the proof hashes required by
-	// the mirror for the specified entry package.
-	PackageProof(ctx context.Context, start, end uint64) ([][]byte, error)
-}
+// PackageProverFunc computes and returns the proof hashes required by the mirror
+// for the specified entry package covering the index interval [start, end).
+type PackageProverFunc func(ctx context.Context, start, end uint64) ([][]byte, error)
 
 // Options holds the configuration for a tlog-mirror Client.
 type Options struct {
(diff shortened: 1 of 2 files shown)


client/mirror/client.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// parseConflict parses the text/x.tlog.mirror-info body of a 409 Conflict response.
+// Format:
+//   - The tree size of a valid pending checkpoint, in decimal
+//   - The next entry, in decimal
+//   - An opaque, possibly zero length, ticket value, encoded in base64
+//
+// nolint:unused
+func parseConflict(r io.Reader) (ErrConflict, error) {
```
AlCutter: It's quite unusual to return two errors... especially since precisely one of them should always be nil and the other non-nil. 

What's the thinking for it being like this over just returning a single `error` which is either an `ErrConflict` if it successfully parses, or some other type of error otherwise?

client/mirror/client.go
```diff
@@ -0,0 +1,256 @@
...
+	pr, _ := io.Pipe()
+
+	// TODO(roger2hk): Implement streaming entries.
+
+	u, err := c.opts.mirrorURL.Parse("add-entries")
+	if err != nil {
+		return fmt.Errorf("failed to parse add-entries URL: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), pr)
```
AlCutter, line 186: I don't understand how this can work, this is the only reference to `pr`, and `pr` is never closed anywhere. I suspect this method will just block forever if called.

client/mirror/client.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create HTTP request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/octet-stream")
+	req.Header.Set("Content-Encoding", "gzip")
```
AlCutter, line 191: Out of curiosity, does setting this header "magically" make the body read from `pr` be gzipped behind the scenes, or does the streaming body need to be compressed before it's fed to `pr`?

client/mirror/client.go
```diff
@@ -0,0 +1,256 @@
...
+	return fmt.Sprintf("mirror sync conflict: pending size %d, next entry %d, ticket length %d", e.PendingSize, e.NextEntry, len(e.Ticket))
+}
+
+// parseConflict parses the text/x.tlog.mirror-info body of a 409 Conflict response.
+// Format:
+//   - The tree size of a valid pending checkpoint, in decimal
+//   - The next entry, in decimal
+//   - An opaque, possibly zero length, ticket value, encoded in base64
+//
+// nolint:unused
```
AlCutter, line 156: This `nolint:unused` seems to be untrue? (this is called from L194)

client/mirror/client.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+
+	if resp.StatusCode == http.StatusConflict {
+		conflict, err := parseConflict(resp.Body)
+		if err != nil {
+			return fmt.Errorf("conflict returned, but failed to parse response body: %w", err)
+		}
```
AlCutter: Related to the comment above, this extra info could possibly be moved up into `parseConflict` so this just becomes `return parseConflict(resp.Body)`

client/mirror/client.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	req.Header.Set("Content-Type", "application/octet-stream")
+	req.Header.Set("Content-Encoding", "gzip")
+
+	resp, err := c.opts.httpClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("POST %s failed: %w", u, err)
+	}
+	defer func() {
+		_ = resp.Body.Close()
```
AlCutter, line 200: I think this should possibly also do a defensive `_, _ = io.Copy(io.Discard, resp.Body)` or similar so that the HTTP connection can be reused with `keepalive` even when there's some unexpected situation. (Currently, I think all cases are covered via `parseConflict` and `!= http.StatusOK`, but it's easy to imagine this being broken in the future with further changes.

client/mirror/client.go
roger2hk: For some reason, the lint is still complaining. I guess this is because the parent function is not called anywhere yet.

```sh
> golangci-lint run ./...
client/mirror/client.go:159:6: func parseConflict is unused (unused)
     ^
1 issues:
* unused: 1
```

client/mirror/client.go
roger2hk: Added the missing `gzip.NewWriter(pw)`.

client/mirror/client.go
roger2hk: Updated to `return parseConflict(resp.Body)`.

client/mirror/client.go
roger2hk: Added the defensive `_, _ = io.Copy(io.Discard, resp.Body)`.

client/mirror/client.go
roger2hk: Added the `pr.Close()`.

client/mirror/client.go
roger2hk: Updated to return a single `error`.

client/mirror/client.go
```diff
@@ -0,0 +1,265 @@
...
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/transparency-dev/tessera/client"
+)
+
+// PackageProver is an interface for generating proof hashes required for
+// the entry package covering the index interval [start, end).
+type PackageProver interface {
```
AlCutter: Could this be a `type PackageProver func(...)` or are you expecting to widen the interface?

roger2hk: Updated to a function type. I couldn't think of any new function for the interface.

AlCutter (approved):

PR #966, opened 2026-05-21
publishCheckpoint returns the next publication time instead of the previous + nanoseconds for AWS

### LINKED ISSUE #965: Too many checkpoint updates attempts when a tree isn't growing
When a tree doesn't grow, Tessera tries to update the checkpoint too often.

Introduced in #940.

phbnf:
Fixes #965.

This PR makes `publishCheckpoint` return the timestamp at which Tessera should try again to publish a checkpoint, instead of the last time that it successfully updated it. This allows better scheduling, especially when the tree doesn't grow.

While we're here, also make the AWS implementation store the last publication timestamp in nanoseconds rather than seconds, like other implementation. This allows for sub-second scheduling.

This PR adds some test (contrary to #940). It checks that:
 - the returned timestamps is always far enough in the future, but not _too_ far away. Implementing tighter bounds is challenging.
 - adds a specific test for when `minStaleRepub` is set to 0.

### DIFF
--- a/storage/aws/aws.go
+++ b/storage/aws/aws.go
@@ -305,7 +305,7 @@ func (a *Appender) integrateEntriesJob(ctx context.Context) {
 			defer func() {
 				opsHistogram.Record(ctx, time.Since(start).Milliseconds(), metric.WithAttributes(opNameKey.String("integrateEntries")))
 			}()
-			
+
 			ctx, cancel := context.WithTimeout(ctx, defaultIntegrationTimeout)
 			defer cancel() // Note: ok because we're in a func passed to TraceErr here!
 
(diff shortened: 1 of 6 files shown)


storage/aws/aws.go
```diff
@@ -1294,16 +1294,25 @@ func (s *mySQLSequencer) publishCheckpoint(ctx context.Context, minStaleActive,
...
 	var lastSize sql.NullInt64
-	if err := pRow.Scan(&pubAt, &lastSize); err != nil {
+	if err := pRow.Scan(&publishedAt, &lastSize); err != nil {
 		return time.Time{}, fmt.Errorf("failed to parse PubCoord: %v", err)
 	}
-	cpAge := time.Since(time.Unix(pubAt, 0))
+	var pubAt time.Time
+	// Tessera used to store publishedAt in seconds, and now stores it in nanoseconds.
+	// This handles the transition.
+	const nanosecondThreshold = 1_310_324_790_000
```
AlCutter, line 1305: hehe, nice :)

AlCutter (approved):

PR #973, opened 2026-06-02
[POSIX] Use flock instead of fcntl

mhutchinson:
BSD flock locks are associated with the open file description rather than the PID, which prevents different file descriptors in the same process from concurrently acquiring the lock. This avoids a silent-release vulnerability where closing any file descriptor to the lock file in the process immediately drops the lock for all other descriptors.

This doesn't fix any immediate known issues. However, it does make things less brittle. Should we change which files are written from different goroutines in the future, this will protect against multiple workers ignoring each other's locks.

It also avoids a risk that a library user could create 2 instances of a POSIX storage backed by the same directory. Previously these would have ignored each other's locks, but now they will respect them.

### DIFF
--- a/storage/posix/files.go
+++ b/storage/posix/files.go
@@ -217,8 +217,8 @@ func (s *Storage) lockFile(ctx context.Context, p string) (func() error, error) 
 		for {
 			if err := syscall.Flock(int(f.Fd()), syscall.LOCK_EX); err != syscall.EINTR {
 				if err != nil {
-					_ = f.Close()
-					return nil, err
+					errClose := f.Close()
+					return nil, errors.Join(err, errClose)
 				}
 				span.AddEvent("Lock taken")
 				posixOpsHistogram.Record(ctx, time.Since(now).Milliseconds(), metric.WithAttributes(opNameKey.String(fmt.Sprintf("lock-%s", p))))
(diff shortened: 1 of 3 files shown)


storage/posix/files.go
```diff
@@ -208,30 +207,28 @@ func (s *Storage) lockFile(ctx context.Context, p string) (func() error, error)
 
 		span.AddEvent("Open file")
 		p = filepath.Join(s.cfg.Path, stateDir, p)
-		f, err := os.OpenFile(p, syscall.O_CREAT|syscall.O_RDWR|syscall.O_CLOEXEC, filePerm)
+		f, err := os.OpenFile(p, os.O_CREATE|os.O_RDWR, filePerm)
```
AlCutter, line 210: What's the thinking behind removing the close-on-exec bit?
With that flag present we're explicitly saying "this is _our_ FD, and we don't want it being passed on to any child processes".
Without it, we could leak this FD to a child process if, say, the application uses `os/exec`. That probably isn't what we want - this is strictly internal to Tessera.

AlCutter (approved):

storage/posix/files.go
AlCutter: Chat IRL: It's because Go's runtime already _always_ sets this flag inside the implementation of `OpenFile`.

AlCutter (approved):

PR #981, opened 2026-06-03
Implement tlog-mirror client `Sync` function

roger2hk:

This pull request not only implements the tlog-mirror client `Sync` function but also moves the mirror cosignatures response from `add-checkpoint` to `add-entries`.

If I understand correctly, the witness signatures from `add-checkpoint` endpoints are not required to return to the applications using the `tlog-mirror` client library. Only mirror's own signatures are required.

### DIFF
--- a/client/mirror/client.go
+++ b/client/mirror/client.go
@@ -178,19 +178,19 @@ func (c *Client) pushEntries(ctx context.Context, uploadStart, uploadEnd uint64,
 
 	u, err := c.opts.mirrorURL.Parse("add-entries")
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse add-entries URL: %w", err)
+		return nil, fmt.Errorf("failed to parse add-entries URL: %v", err)
 	}
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), pr)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create HTTP request: %w", err)
+		return nil, fmt.Errorf("failed to create HTTP request: %v", err)
 	}
 	req.Header.Set("Content-Type", "application/octet-stream")
 	req.Header.Set("Content-Encoding", "gzip")
 
 	resp, err := c.opts.httpClient.Do(req)
 	if err != nil {
-		return nil, fmt.Errorf("POST %s failed: %w", u, err)
+		return nil, fmt.Errorf("POST %s failed: %v", u, err)
 	}
 	defer func() {
 		// Drain any remaining response body to enable HTTP keep-alive/connection reuse.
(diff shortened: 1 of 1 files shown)


client/mirror/client.go
```diff
@@ -147,22 +147,19 @@ func (e ErrConflict) Error() string {
...
 func parseConflict(r io.Reader) error {
 	// TODO(roger2hk): Implement this.
 
 	return errors.New("TODO")
 }
 
 // pushEntries streams entry packages and their proofs to the mirror's /add-entries endpoint.
-// nolint:unused
-func (c *Client) pushEntries(ctx context.Context, uploadStart, uploadEnd uint64, ticket []byte) error {
+func (c *Client) pushEntries(ctx context.Context, uploadStart, uploadEnd uint64, ticket []byte) ([]byte, error) {
```
AlCutter, line 163: nit: it would be nice to update the comment to add `// Returns ...`, especially since the returned type is a `[]byte`

client/mirror/client.go
```diff
@@ -180,83 +177,133 @@ func (c *Client) pushEntries(ctx context.Context, uploadStart, uploadEnd uint64,
...
-	// 3. Push entries up to target size in packages of 256, handling concurrent conflicts and retries.
+	// Get the mirror's current state by querying it with upload_start=0, upload_end=0 (guaranteed to conflict).
+	_, err := c.pushEntries(ctx, 0, 0, nil)
+	if err == nil {
+		return nil, errors.New("unexpected success when querying mirror status with size 0")
+	}
+
+	var conflict ErrConflict
+	if !errors.As(err, &conflict) {
+		return nil, fmt.Errorf("failed to retrieve mirror status: %w", err)
```
AlCutter: Careful with the `%w` - do we explicitly want whatever errors retuned by `pushEntries` to become part of our API?

client/mirror/client.go
```diff
@@ -180,83 +177,133 @@ func (c *Client) pushEntries(ctx context.Context, uploadStart, uploadEnd uint64,
...
+	for {
+		var err error
+		cosigs, err = c.pushEntries(ctx, nextEntry, uploadEnd, ticket)
+		if err == nil {
+			break
+		}
 
-	return nil, errors.New("WIP")
+		if !errors.As(err, &conflict) {
+			return nil, fmt.Errorf("sync failed during entry upload: %w", err)
```
AlCutter: Same `%w` question/comment here.

AlCutter (approved):

client/mirror/client.go
roger2hk: Updated all `%w` to `%v`.

client/mirror/client.go
roger2hk: Updated all `%w` to `%v`.

client/mirror/client.go
roger2hk: Added the return comment.

AlCutter (approved):

PR #982, opened 2026-06-03
Implement tlog-mirror client `parseConflict` function

roger2hk:

### DIFF
--- a/client/mirror/client.go
+++ b/client/mirror/client.go
@@ -19,7 +19,7 @@
 package mirror
 
 import (
-	"bufio"
+	"bytes"
 	"compress/gzip"
 	"context"
 	"encoding/base64"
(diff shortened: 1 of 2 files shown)


client/mirror/client.go
```diff
@@ -155,9 +158,50 @@ func (e ErrConflict) Error() string {
 //
 // nolint:unused
 func parseConflict(r io.Reader) error {
-	// TODO(roger2hk): Implement this.
+	scanner := bufio.NewScanner(r)
```
AlCutter: Since this is expected to be a small amount of data, it might be more concise to do something like:

```go
	all, err := io.ReadAll(&io.LimitedReader{N: 10 << 10, R: r})
	if err != nil {
		return err
	}
	bits := bytes.Split(all, []byte("\n"))
	if len(bits) != 3 {
		return fmt.Errorf("got %d fields, want 3", len(bits))
	}
	return nil
    
    pendingSize, err := strconv(bits[0])
    ...
    nextEntry, err := strconv(bits[1])
    ...
    ticket, err := base64.St...
    ...
}

Comment (roger2hk) [reply client/mirror/client.go]: According to the specification, there will be 3 newlines. The `len(bits)` should be `4` instead of `3`.

```
When sending a "409 Conflict" or "202 Accepted" response, the response body MUST have a Content-Type of text/x.tlog.mirror-info and consist of three lines, each followed by a newline (U+000A):

- The tree size of a valid pending checkpoint, in decimal
- The next entry, in decimal
- An opaque, possibly zero length, ticket value, encoded in base64
```

roger2hk: Updated to use `io.LimitedReader`.

AlCutter (approved):

PR #987, opened 2026-06-05
Add witness core

AlCutter:
This PR adds the logic for a witness to the MTC mirror code.

Towards #945

### DIFF
--- a/cmd/mtc/mirror/internal/handler/handler.go
+++ b/cmd/mtc/mirror/internal/handler/handler.go
@@ -63,11 +63,10 @@ func addCheckpoint(m Mirror) http.HandlerFunc {
 			return
 		}
 
-		sc, body, contentType, err := handleUpdate(r.Context(), m, origin, oldSize, cp, proof)
+		sc, body, contentType, err := handleCheckpointUpdate(r.Context(), m, origin, oldSize, cp, proof)
 		if err != nil {
-			status := http.StatusInternalServerError
 			slog.InfoContext(r.Context(), "Witness update failed", slog.Any("error", err.Error()))
-			w.WriteHeader(status)
+			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
 
@@ -83,10 +82,10 @@ func addCheckpoint(m Mirror) http.HandlerFunc {
 	}
 }
 
-// handleUpdate submits the provided checkpoint to the witness and interprets any errors which may result.
+// handleCheckpointUpdate submits the provided checkpoint to the witness and interprets any errors which may result.
 //
 // Returns an appropriate HTTP status code, response body, and Content Type representing the outcome.
-func handleUpdate(ctx context.Context, m Mirror, origin string, oldSize uint64, cp []byte, proof [][]byte) (int, []byte, string, error) {
+func handleCheckpointUpdate(ctx context.Context, m Mirror, origin string, oldSize uint64, cp []byte, proof [][]byte) (int, []byte, string, error) {
 	sigs, trustedSize, updateErr := m.AddCheckpoint(ctx, origin, oldSize, proof, cp)
 	// Finally, handle any "soft" error from the update:
 	if updateErr != nil {
@@ -117,27 +116,28 @@ func handleUpdate(ctx context.Context, m Mirror, origin string, oldSize uint64, 
 // - a previous size line,
 // - zero or more consistency proof lines,
 // - and an empty line,
-// - followed by a [checkpoint][].
+// - followed by a checkpoint.
 func parseBody(r io.Reader) (string, uint64, [][]byte, []byte, error) {
 	b := bufio.NewReader(r)
-	sizeLine, _, err := b.ReadLine()
+	sizeLine, err := b.ReadString('\n')
 	if err != nil {
 		return "", 0, nil, nil, err
 	}
 	var size uint64
-	if n, err := fmt.Sscanf(string(sizeLine), "old %d", &size); err != nil || n != 1 {
+	if n, err := fmt.Sscanf(strings.TrimSuffix(sizeLine, "\n"), "old %d", &size); err != nil || n != 1 {
 		return "", 0, nil, nil, err
 	}
 	proof := [][]byte{}
 	for {
-		l, _, err := b.ReadLine()
+		l, err := b.ReadString('\n')
 		if err != nil {
 			return "", 0, nil, nil, err
 		}
+		l = strings.TrimSuffix(l, "\n")
 		if len(l) == 0 {
 			break
 		}
-		hash, err := base64.StdEncoding.DecodeString(string(l))
+		hash, err := base64.StdEncoding.DecodeString(l)
 		if err != nil {
 			return "", 0, nil, nil, err
 		}
@@ -149,7 +149,7 @@ func parseBody(r io.Reader) (string, uint64, [][]byte, []byte, error) {
 	}
 	s := strings.SplitN(string(cp), "\n", 2)
 	if len(s) != 2 {
-		return "", 0, nil, nil, err
+		return "", 0, nil, nil, errors.New("invalid checkpoint")
 	}
 
 	origin := s[0]
--- a/cmd/mtc/mirror/internal/handler/handler.go
+++ b/cmd/mtc/mirror/internal/handler/handler.go
@@ -22,14 +22,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
-	"strconv"
 	"strings"
 
-	"log/slog"
-
 	"github.com/transparency-dev/tessera"
-	"github.com/transparency-dev/tessera/internal/parse"
+	"github.com/transparency-dev/tessera/internal/witness"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


cmd/mtc/mirror/internal/handler/handler.go
```diff
@@ -55,75 +53,107 @@ func New(m Mirror) http.Handler {
...
+		}
+		proof = append(proof, hash)
+	}
+	cp, err := io.ReadAll(b)
+	if err != nil {
+		return "", 0, nil, nil, err
 	}
+	s := strings.SplitN(string(cp), "\n", 2)
+	if len(s) != 2 {
+		return "", 0, nil, nil, err
```
roger2hk: The `err` is `nil` here as `strings.SplitN` doesn't return any `err`.

AlCutter: Good spot, ta!

internal/witness/witness.go
```diff
@@ -11,32 +11,253 @@
...
+		// SPEC:  If the old size matches the checkpoint size, the witness MUST check that the root hashes are
+		//        also identical.
+		if next.Size == prev.Size {
+			if !bytes.Equal(next.Hash, prev.Hash) {
+				slog.ErrorContext(ctx, "INCONSISTENT CHECKPOINTS", slog.String("origin", w.logVerifier.Name()), slog.Any("prev", prev), slog.Any("next", next))
+				retSize, retSigs = 0, nil
+				return nil, ErrRootMismatch
+			}
+			// This used to short-circuit here to save work.
+			// However, having the most recently witnessed timestamp available is beneficial to demonstrate freshness.
```
roger2hk: If the sizes are equal and the hashes match, we can sign and return the new checkpoint immediately without verifying the consistency proof on L183.

cmd/mtc/mirror/internal/handler/handler.go
```diff
@@ -55,75 +53,107 @@ func New(m Mirror) http.Handler {
...
+// parseBody reads the incoming request and parses into constituent parts.
+//
+// The request body MUST be a sequence of
+// - a previous size line,
+// - zero or more consistency proof lines,
+// - and an empty line,
+// - followed by a [checkpoint][].
+func parseBody(r io.Reader) (string, uint64, [][]byte, []byte, error) {
+	b := bufio.NewReader(r)
+	sizeLine, _, err := b.ReadLine()
```
roger2hk: `ReadLine()` from `bufio.ReadLine` has a default read buffer size of 4096 bytes. If the line has more than 4096 bytes, it will return unexpected values.

internal/witness/witness.go
AlCutter: While this is true, I think this would be a poor trade-off: we would be adding cognitive load/complexity within a security critical function (currently there is a single location where the act of creating a signature can occur) for an extremely minimal, if any, gain in performance (verifying a consistency proof of length zero is essentially free).

cmd/mtc/mirror/internal/handler/handler.go
AlCutter: I'm not sure what the attack would be, but just in case I've changed to `ReadString` anyway.

internal/witness/witness.go
AlCutter: Actually, that "single location" claim wasn't true, but thanks to your comment I've made it true!

internal/witness/witness.go
```diff
@@ -11,32 +11,253 @@
...
+// Update updates the latest checkpoint if nextRaw is consistent with the current
+// latest one for this log.
+//
+// The values returned depend on whether or not the new checkpoint is accepted, and
+// if not, the reason it was rejected. This can be determined through the error:
+//
+// - no error: The checkpoint was accepted, and a serialised note-signature is returned.
+// - ErrCheckpointStale or ErrOldSizeInvalid: the presented checkpoint is out of date, the size of the current checkpoint is returned.
+// - Any other error, no supporting values are returned.
 func (w *Witness) Update(ctx context.Context, oldSize uint64, nextRaw []byte, cProof [][]byte) ([]byte, uint64, error) {
```
roger2hk, line 107: The origin of the new checkpoint was not checked against the old checkpoint in this function. I think there is a chance to mess up the witness if the client submits a checkpoint with a different origin signed with the correct key(s).

cmd/mtc/mirror/internal/handler/handler.go
roger2hk: Agreed, since we cap the HTTP request body size at 64 KiB, there is no actual memory attack vector here.

The main issue with `ReadLine` is its API fragility although it's not likely to hit the 4096 bytes limit.

cmd/mtc/mirror/internal/handler/handler.go
```diff
@@ -55,75 +53,107 @@ func New(m Mirror) http.Handler {
...
+	if err != nil {
+		return "", 0, nil, nil, err
+	}
+	var size uint64
+	if n, err := fmt.Sscanf(strings.TrimSuffix(sizeLine, "\n"), "old %d", &size); err != nil || n != 1 {
+		return "", 0, nil, nil, err
+	}
+	proof := [][]byte{}
+	for {
+		l, _, err := b.ReadLine()
```
roger2hk: Use `ReadString` here as well?

internal/witness/witness.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 The Tessera authors. All Rights Reserved.
+// Copyright 2021 Google LLC. All Rights Reserved.
```
roger2hk, line 1: Back to 2021? 🤔

cmd/mtc/mirror/internal/handler/handler.go
```diff
@@ -55,75 +53,107 @@ func New(m Mirror) http.Handler {
...
+			slog.InfoContext(r.Context(), "Invalid witness request", slog.Any("error", err.Error()))
+			w.WriteHeader(http.StatusBadRequest)
 			return
 		}
-		oldSize, err := strconv.ParseUint(strings.TrimPrefix(oldLine, "old "), 10, 64)
+
+		sc, body, contentType, err := handleUpdate(r.Context(), m, origin, oldSize, cp, proof)
 		if err != nil {
-			http.Error(w, "invalid old size", http.StatusBadRequest)
+			status := http.StatusInternalServerError
```
roger2hk: nit: Move `status` into `w.WriteHeader` to consistent with L62?

cmd/mtc/mirror/internal/handler/handler.go
```diff
@@ -55,75 +53,107 @@ func New(m Mirror) http.Handler {
...
 
-		// 3. Remaining data is the checkpoint.
-		cp, err := io.ReadAll(reader)
-		if err != nil {
-			http.Error(w, "failed to read checkpoint", http.StatusBadRequest)
-			return
+// handleUpdate submits the provided checkpoint to the witness and interprets any errors which may result.
+//
+// Returns an appropriate HTTP status code, response body, and Content Type representing the outcome.
+func handleUpdate(ctx context.Context, m Mirror, origin string, oldSize uint64, cp []byte, proof [][]byte) (int, []byte, string, error) {
```
roger2hk: The function name `handleUpdate` is a bit ambiguous in this `handler.go`. Maybe it's better to rename this to `handleCheckpointUpdate` when there are also entries to handle.

cmd/mtc/mirror/internal/handler/handler.go
```diff
@@ -55,75 +53,107 @@ func New(m Mirror) http.Handler {
...
 
-		// 4. Extract the origin from the checkpoint and pass the request to the backend to take action.
-		origin, _, _, err := parse.CheckpointUnsafe(cp)
+// parseBody reads the incoming request and parses into constituent parts.
+//
+// The request body MUST be a sequence of
+// - a previous size line,
+// - zero or more consistency proof lines,
+// - and an empty line,
+// - followed by a [checkpoint][].
```
roger2hk: The `[checkpoint]` link doesn't sound valid.

internal/witness/witness.go
```diff
@@ -11,32 +11,253 @@
...
+	signers     []note.Signer
+	logVerifier note.Verifier
 }
 
-func New(s Persistence) *Witness {
-	return &Witness{storage: s}
+// New creates a new witness for the configured log.
+func New(ctx context.Context, wo Opts) (*Witness, error) {
+	return &Witness{
+		lsp:         wo.Persistence,
```
roger2hk: nit: Rename `lsp` or add comment? It took me a while to understand `lsp` stands for Log State Persistence.

internal/witness/witness.go
```diff
@@ -11,32 +11,253 @@
...
+	"unicode"
+	"unicode/utf8"
+
+	"github.com/transparency-dev/formats/log"
+	"github.com/transparency-dev/merkle/proof"
+	"github.com/transparency-dev/merkle/rfc6962"
+	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/mod/sumdb/note"
+
+	i_otel "github.com/transparency-dev/tessera/internal/otel"
```
roger2hk: Avoid using `_` for imported packages.

https://google.github.io/styleguide/go/decisions#import-renaming

roger2hk (approved):

internal/witness/witness.go
AlCutter: This code is copied from `transparency-dev/witness`, so I left the copyright header intact.

internal/witness/witness.go
AlCutter: The origin is checked in `log.ParseCheckpoint()` (via `verifyCheckpoint()`), into which we pass the name from the logVerifier as the expected origin.

PR #989, opened 2026-06-08
Add POSIX mirror lifecycle outline & witness support

AlCutter:
This PR adds an skeleton POSIX storage implementation for the mirror lifecycle, and connects it up to the POSIX mirror binary.

Towards #945

### DIFF
--- a/cmd/mtc/mirror/posix/main.go
+++ b/cmd/mtc/mirror/posix/main.go
@@ -19,14 +19,17 @@ import (
 	"encoding/json"
 	"errors"
 	"flag"
+	"fmt"
 	"log/slog"
 	"net/http"
 	"os"
 	"path/filepath"
+	"net/url"
 
 	fnote "github.com/transparency-dev/formats/note"
 	"github.com/transparency-dev/tessera"
 	"github.com/transparency-dev/tessera/cmd/mtc/mirror/internal/config"
+	"github.com/transparency-dev/tessera/storage/posix"
 	"github.com/transparency-dev/tessera/cmd/mtc/mirror/internal/handler"
 	"github.com/transparency-dev/witness/omniwitness"
 	"github.com/transparency-dev/witness/witness"
@@ -42,6 +45,7 @@ var (
 	listenAddr        = flag.String("listen_addr", ":8080", "The address to listen on for HTTP requests.")
 	storageDir        = flag.String("storage_dir", "", "Directory to store mirror data.")
 	witnessSignerPath = flag.String("witness_signer_path", "", "The path to the note-formatted witness signer secret key.")
+	mirrorSignerPath  = flag.String("mirror_signer_path", "", "The path to the note-formatted mirror signer secret key.")
 	mirrorConfigPath  = flag.String("config_path", "", "The path to the mirror configuration file.")
 )
 
@@ -57,6 +61,8 @@ func main() {
 		}
 	}()
 
+	mirrorSigner := mustCreateSigner(ctx, *mirrorSignerPath)
+
 	mux := handler.NewMirrorMux()
 	cfg := mirrorConfigFromFlags(ctx)
 	for _, l := range cfg.Logs {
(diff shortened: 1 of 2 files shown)


storage/posix/files.go
```diff
@@ -1110,3 +1111,82 @@ func (m *MigrationStorage) fetchLeafHashes(ctx context.Context, from, to, source
...
+	unlock, err := m.s.lockFile(ctx, treeStateLock)
+	if err != nil {
+		return fmt.Errorf("failed to lock tree state: %v", err)
+	}
+	defer func() {
+		if err := unlock(); err != nil {
+			// Oh dear, panic Mr Mainwaring!
+			panic(fmt.Errorf("failed to unlock tree state: %v", err))
+		}
+		m.s.mu.Unlock()
```
roger2hk: There is a chance of mutex deadlock. The mutex unlock will not trigger when the `err` is not `nil` on L1151.

Defer the mutex unlock immediately after `m.s.mu.Lock()`.

cmd/mtc/mirror/posix/main.go
```diff
@@ -147,30 +186,30 @@ func witnessFromFlags(ctx context.Context) (*witness.Witness, *sqlite.Persistenc
...
+
+func mustCreateSigner(ctx context.Context, path string) note.Signer {
+	if path == "" {
+		slog.ErrorContext(ctx, "Signer path not specified")
+		os.Exit(1)
+	}
+	r, err := os.ReadFile(path)
 	if err != nil {
-		slog.ErrorContext(ctx, "Failed to read witness cosigner file", slog.String("path", *witnessSignerPath), slog.Any("error", err))
+		slog.ErrorContext(ctx, "Failed to read cosigner file", slog.String("path", path), slog.Any("error", err))
```
roger2hk: I'm not too familiar with the terminology. Should we say `signer` instead of `cosigner` in the error log in this function?

cmd/mtc/mirror/posix/main.go
```diff
@@ -87,6 +103,29 @@ func main() {
 	}
 }
 
+// newMirrorTarget creates a new POSIX driver and MirrorTarget for the given origin.
+//
+// The target directory for the driver is derived from the storage directory and the origin in accordance
+// with the `tlog-mirror` spec, allowing the root of the storage directory to be exported directly to read-only clients.
+func newMirrorTarget(ctx context.Context, w *witness.Witness, origin string, mirrorSigner note.Signer) (*tessera.MirrorTarget, error) {	
+		targetDir := filepath.Join(*storageDir, url.PathEscape(origin))
+		if err := os.MkdirAll(targetDir, 0o700); err != nil && !errors.Is(err, os.ErrExist) {
```
roger2hk: It should be safe to remove `!errors.Is(err, os.ErrExist)` as no error will be returned if the directory is already created.

```
If path is already a directory, MkdirAll does nothing and returns nil.
```

https://pkg.go.dev/os#MkdirAll

roger2hk: `0o700` seems to be contradicted to the godoc comment "allowing the root of the storage directory to be exported directly to read-only clients".

roger2hk (approved):

cmd/mtc/mirror/posix/main.go
AlCutter: ==> `0o755`

cmd/mtc/mirror/posix/main.go
AlCutter: It's a bit dependent on context, sadly - these are cosigners from the PoV of `tlog-*` and MTC spec PoV, and the _type_ of key these are expected to be are `Cosignature/v1`, but from the `note` library PoV they're `note.Signer`s.

I've renamed anything to do with them as `...Cosigner...` in this file for consistency, as the operator is more likely to be familiar with.

storage/posix/files.go
AlCutter: Good catch!

I've done two things:
1) changed all "double lock" locations to defer the unlocks separately (because that's a more correct pattern)
2) changed _this_ location to `panic()` on failure to lock to match all the others - this also "fixes" the deadlock issue, but makes behaviour consistent across lifecycles (we can later decide if we want to modify the behaviour).

PR #994, opened 2026-06-09
[Dependabot] Don't bump go version

mhutchinson:
We don't want our upstream deps being in control of this. We can change it, but on our own cadence. Obviously major security issues that require a bump can change this cadence, but in general, we do this when the next version of Go is released.


mhutchinson:
Motivation is #993 right now, but I've been fighting this for a week or more across the repos.

phbnf (approved):

PR #995, opened 2026-06-09
Implement tlog-mirror client `pushEntries`

roger2hk:


client/mirror/client.go
```diff
@@ -195,15 +196,7 @@ func (c *Client) pushEntries(ctx context.Context, uploadStart, uploadEnd uint64,
...
-	go func() {
-		gw := gzip.NewWriter(pw)
-		defer func() {
-			_ = gw.Close()
-			_ = pw.Close()
-		}()
-
-		// TODO(roger2hk): Implement streaming entries.
-	}()
+	go c.streamEntries(ctx, uploadStart, uploadEnd, ticket, pw)
```
AlCutter, line 202: What happens to this goroutine if the HTTP request breaks? Will it gracefully stop and exit, or will it hang?

Do you need to create a cancelable context from `ctx`, use that here and defer the cancel()?

roger2hk: Good catch! The goroutine would hang. I've added the cancelable context.

AlCutter (approved):

PR #1002, opened 2026-06-12
Use transparency-dev/witness

AlCutter:
This PR switches over to directly using the witness implementation from [`transparency-dev/witness`](https://github.com/transparency-dev/witness).

Towards #945


mirror_lifecycle.go
```diff
@@ -22,14 +22,15 @@ import (
...
 	"github.com/transparency-dev/tessera/api/layout"
-	"github.com/transparency-dev/tessera/internal/witness"
+	"github.com/transparency-dev/witness/witness"
 	"golang.org/x/mod/sumdb/note"
 )
 
 // MirrorOptions holds mirror lifecycle settings for all storage implementations.
 type MirrorOptions struct {
 	logVerifier note.Verifier
 	signer      note.Signer
```
roger2hk: Are the `logVerified` and `signer` still required? It seems like they're now embedded in the `witness` struct.

cmd/mtc/mirror/internal/handler/handler_test.go
```diff
@@ -110,7 +58,11 @@ func TestAddEntries(t *testing.T) {
 			return []byte("— test-cosig\n"), nil
 		},
 	}
-	h := New(mock)
+	mux := NewMirrorMux()
+	if err := mux.AddTarget(testOrigin, mock); err != nil {
+		t.Fatalf("AddTarget() failed: %v", err)
+	}
+	h := New(mux, nil)
```
roger2hk, line 65: Passing the `nil` here would cause a panic if any test tries to call `/add-checkpoint` endpoint later.

roger2hk (approved):

mirror_lifecycle.go
AlCutter: `signer` will still be needed for responses to `add-entries`, I'll remove `logVerifier` for the moment...

cmd/mtc/mirror/internal/handler/handler_test.go
AlCutter: Yup, I'll update this as the implementation of `add-checkpoint` progresses.

PR #1003, opened 2026-06-15
Add tlog-mirror client `Sync` function unit tests

roger2hk:

Non-happy path unit tests are added.

### DIFF
--- a/client/mirror/client_test.go
+++ b/client/mirror/client_test.go
@@ -410,15 +410,28 @@ func TestNewClientValidation(t *testing.T) {
 
 // fakeMirror is a mock implementation of a tlog-mirror compliant server.
 type fakeMirror struct {
-	t                   *testing.T
-	origin              string
-	ticket              []byte
-	targetCheckpoint    []byte
-	expectedCosigs      []byte
-	numEntries          int
-	proofHashes         [][]byte
-	addEntriesStatus    int
-	addCheckpointStatus int
+	t                      *testing.T
+	origin                 string
+	ticket                 []byte
+	targetCheckpoint       []byte
+	expectedCosigs         []byte
+	numEntries             int
+	proofHashes            [][]byte
+	addEntriesStatus       int
+	addCheckpointStatus    int
+	initialPendingSize     uint64
+	initialNextEntry       uint64
+	initialStatus          int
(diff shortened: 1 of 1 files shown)


client/mirror/client_test.go
```diff
@@ -570,9 +616,34 @@ func (fm *fakeMirror) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		if len(lines) != 2 {
+			fm.t.Errorf("invalid checkpoint request body format")
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		header := lines[0]
+		bodyCheckpoint := lines[1]
+
+		headerLines := bytes.Split(header, []byte("\n"))
+		if len(headerLines) < 1 {
```
AlCutter: would `== 0` be clearer?

client/mirror/client_test.go
```diff
@@ -570,9 +616,34 @@ func (fm *fakeMirror) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		if oldSize != fm.initialPendingSize {
+			fm.t.Errorf("oldSize = %d, want %d", oldSize, fm.initialPendingSize)
+		}
+		if oldSize > 0 && len(headerLines) <= 1 {
+			fm.t.Errorf("expected consistency proof lines in header")
+		}
+
+		if !bytes.Equal(bodyCheckpoint, fm.targetCheckpoint) {
+			fm.t.Errorf("body checkpoint = %q, want %q", string(bodyCheckpoint), string(fm.targetCheckpoint))
 		}
```
AlCutter, line 665: These cases log an error, but allow the method to get to the bottom and return `StatusOK`, although the check immediately above explicitly returns a bad status and bails.

Is this difference intended?

AlCutter (approved):

client/mirror/client_test.go
roger2hk: Oops, the difference was not intended. The header will now write `http.StatusBadRequest` when there is a failure.

client/mirror/client_test.go
roger2hk: Updated to `== 0`.

PR #1004, opened 2026-06-17
Use transparency-dev/witness sqlite persistence

AlCutter:
This PR uses the newly-exposed sqlite witness persistence layer for the POSIX MTC-mirror binary.

Towards #945


cmd/mtc/mirror/posix/main.go
```diff
@@ -58,11 +67,40 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	wPath := filepath.Join(*storageDir, witnessDir)
+	if err := os.MkdirAll(wPath, 0o700); err != nil && !errors.Is(err, os.ErrExist) {
+		slog.ErrorContext(ctx, "Failed to create witness directory", slog.String("path", wPath))
+		os.Exit(1)
+	}
+
+	db, err := sql.Open("sqlite", filepath.Join(wPath, "witness.db"))
```
roger2hk: Close the `db` before shutdown?

cmd/mtc/mirror/posix/main.go
```diff
@@ -16,21 +16,30 @@ package main
 
 import (
```
roger2hk, line 17: Do you need to import "modernc.org/sqlite" package for the sqlite3 driver or rely on the transitive dependency "github.com/transparency-dev/witness/persistence/sqlite"?

roger2hk (approved):

cmd/mtc/mirror/posix/main.go
```diff
@@ -109,3 +118,22 @@ func (f fakeTarget) AddEntries(ctx context.Context, uploadStart, uploadEnd uint6
 	return nil, nil
 }
 
+func witnessSignerFromFlags(ctx context.Context) []note.Signer {
```
roger2hk, line 113: Should the function name be `witnessSignersFromFlags` assuming there could be more than one signer?

cmd/mtc/mirror/posix/main.go
mhutchinson: Discussed in person, but you're missing some "magic" flags here to get good performance: https://github.com/transparency-dev/witness/pull/562

I think we should have a factory method for the Sqlite persistence layer that does this stuff in Witness to avoid anyone needing to know about it outside that package.

AlCutter: This is all handled inside the witness' sqlite persistence layer now 👍

cmd/mtc/mirror/posix/main.go
AlCutter: It's only returning a `[]` so that I don't have to then special case when there are _no_ signers.

cmd/mtc/mirror/posix/main.go
AlCutter: Only the transitive dep; that imports it directly now - this is important because we explicitly don't want the caller deciding to use `mattn/sqlite3` while the implementation is expecting `modernc.org/sqlite`.

mhutchinson (approved):

cmd/mtc/mirror/posix/main.go
AlCutter: This is a good call, but will need a round-trip via the witness to add a `Close` method - I'll do that in another PR.

PR #1006, opened 2026-06-18
Add simple outline mirror config

AlCutter:
This PR adds a simple outline of a means to configure logs in the MTC mirror.

Towards #945

### DIFF
--- a/cmd/mtc/mirror/posix/main.go
+++ b/cmd/mtc/mirror/posix/main.go
@@ -16,16 +16,19 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"flag"
+	"log/slog"
 	"net/http"
 	"os"
 	"path/filepath"
-	"log/slog"
 
 	fnote "github.com/transparency-dev/formats/note"
 	"github.com/transparency-dev/tessera"
+	"github.com/transparency-dev/tessera/cmd/mtc/mirror/internal/config"
 	"github.com/transparency-dev/tessera/cmd/mtc/mirror/internal/handler"
+	"github.com/transparency-dev/witness/omniwitness"
 	"github.com/transparency-dev/witness/witness"
 	"github.com/transparency-dev/witness/persistence/sqlite"
 	"golang.org/x/mod/sumdb/note"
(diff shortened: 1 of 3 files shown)


cmd/mtc/mirror/posix/main.go
```diff
@@ -36,27 +40,38 @@ const (
...
+	cfg := mirrorConfigFromFlags(ctx)
+	for _, l := range cfg.Logs {
+		origin := strings.Split(l.VKey, "+")[0]
+		if err := mux.AddTarget(origin, &fakeTarget{}); err != nil {
+			slog.ErrorContext(ctx, "Failed to add target", slog.String("origin", origin), slog.Any("error", err))
+			os.Exit(1)
+		}
+		if err := wp.AddLogs(ctx, []omniwitness.Log{
+			{Origin: origin, VKey: l.VKey},
+		}); err != nil {
```
roger2hk, line 75: With `wp.AddLogs` function, you can add all the logs in one go instead of multiple write transactions.

cmd/mtc/mirror/posix/main.go
```diff
@@ -36,27 +40,38 @@ const (
...
 		}
 	}()
 
 	mux := handler.NewMirrorMux()
-	if err := mux.AddTarget("example.com/log", &fakeTarget{}); err != nil {
-		slog.ErrorContext(ctx, "Failed to add target", slog.Any("error", err))
-		os.Exit(1)
+	cfg := mirrorConfigFromFlags(ctx)
+	for _, l := range cfg.Logs {
+		origin := strings.Split(l.VKey, "+")[0]
```
roger2hk: Use `fnote.NewVerifier` here to validate the verification key and get the origin from it?

cmd/mtc/mirror/posix/main.go
```diff
@@ -68,49 +83,64 @@ func main() {
...
 
-	w, err := witness.New(ctx, witness.Opts{
-		Persistence: p,
-		Signers:     witnessSignerFromFlags(ctx),
+ 	w, err := witness.New(ctx, witness.Opts{
+ 		Persistence: p,
+ 		Signers:     witnessSignerFromFlags(ctx),
 		VerifierForLog: func(ctx context.Context, origin string) (note.Verifier, bool, error) {
-			return v, true, nil
+			log, ok, err := p.Log(ctx, origin)
```
roger2hk, line 136: The `VerifierForLog` func is called every time hitting the SQLite database when the latest checkpoint is updated. Maybe it's better to put those note verifiers in an in-memory cache for performance.

roger2hk (approved):

cmd/mtc/mirror/posix/main.go
AlCutter: Yeah, I just it this way to avoid having to loop and append to a var, and then add all. I expect all of this to change quite a lot in the near future.

cmd/mtc/mirror/posix/main.go
AlCutter: This is by design, we can always add a cache later if this actually becomes a bottleneck.

roger2hk (approved):

PR #1007, opened 2026-06-18
Support gzip encoding in add-entries requests

AlCutter:
This PR adds support for reading `add-entries` requests with `gzip` encoded bodies.

### DIFF
--- a/cmd/mtc/mirror/internal/handler/handler.go
+++ b/cmd/mtc/mirror/internal/handler/handler.go
@@ -15,6 +15,7 @@
 package handler
 
 import (
+	"compress/gzip"
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


cmd/mtc/mirror/internal/handler/handler.go
```diff
@@ -50,7 +51,32 @@ func addEntries(m *MirrorMux) http.HandlerFunc {
...
+		defer func() {
+			_ = r.Body.Close()
+		}()
+		// SPEC: Mirrors MUST support receiving Content-Encoding: gzip in add-entries requests
+		reader := r.Body
+		switch ce := r.Header.Get("Content-Encoding"); ce {
+		case "":
+			// No Content-Encoding header, nothing to do.
+			break
+		case "gzip":
```
roger2hk, line 66: The header value is case insensitive. `GZIP` or `GZip` are valid values in HTTP/1.1.

https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.2.1

cmd/mtc/mirror/internal/handler/handler.go
```diff
@@ -50,7 +51,32 @@ func addEntries(m *MirrorMux) http.HandlerFunc {
...
+				http.Error(w, fmt.Sprintf("invalid gzip request body: %v", err), http.StatusBadRequest)
+				return
+			}
+			defer func() {
+				_ = reader.Close()
+			}()
+		default:
+			// Log unknown encodings and treat as malformed request
+			slog.WarnContext(r.Context(), "Unknown Content-Encoding", slog.String("encoding", ce))
+			http.Error(w, fmt.Sprintf("unsupported content encoding %q", ce), http.StatusBadRequest)
```
roger2hk, line 79: Here is what the tlog-mirror said:

> When sending responses, mirrors SHOULD send an Accept-Encoding header that includes gzip and any other supported compression algorithms.

I think we should return the supported content encoding when returning this bad request error. Maybe move L94 to the top of the function?

AlCutter: Sure

cmd/mtc/mirror/internal/handler/handler.go
```diff
@@ -64,6 +92,7 @@ func addEntries(m *MirrorMux) http.HandlerFunc {
 			return
 		}
 
+		// SPEC: When sending responses, mirrors SHOULD send an Accept-Encoding header that includes gzip
```
roger2hk: Move this comment as well?

roger2hk (approved):

PR #1013, opened 2026-06-19
Update the MirrorTarget API to return info we'll need at the handler level.

AlCutter:
Update the MirrorTarget API to return info we'll need at the handler level.

Towards #945

### DIFF
--- a/cmd/mtc/mirror/internal/handler/handler.go
+++ b/cmd/mtc/mirror/internal/handler/handler.go
@@ -76,20 +76,23 @@ func addEntries(m *MirrorMux) http.HandlerFunc {
 		default:
 			// Log unknown encodings and treat as malformed request
 			slog.WarnContext(r.Context(), "Unknown Content-Encoding", slog.String("encoding", ce))
-			http.Error(w, fmt.Sprintf("unsupported content encoding %q", ce), http.StatusBadRequest)
+			http.Error(w, "Unsupported content encoding", http.StatusBadRequest)
 			return
 		}
 		req, err := parseAddEntriesPreamble(reader)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


cmd/mtc/mirror/internal/handler/handler.go
```diff
@@ -85,11 +85,14 @@ func addEntries(m *MirrorMux) http.HandlerFunc {
...
-			slog.ErrorContext(r.Context(), "AddEntries failed", slog.Any("error", err))
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+		_, _, _, cosigs, err := m.AddEntries(r.Context(), req.logOrigin, req.uploadStart, req.uploadEnd, req.ticket, req.NextPackage)
+		switch {
+		case errors.Is(err, ErrUnknownLog):
+			http.Error(w, err.Error(), http.StatusNotFound)
+			return
+		case err != nil:
+			slog.ErrorContext(r.Context(), "Failed to add entries", slog.String("origin", req.logOrigin), slog.Any("err", err))
+			http.Error(w, fmt.Sprintf("failed to add entries: %v", err), http.StatusInternalServerError)
```
roger2hk: Sounds risky to return the error message in the HTTP response when we've already logged the error.

roger2hk (approved):

cmd/mtc/mirror/internal/handler/handler.go
AlCutter: Because it may leak details to the user? I've changed to just send a generic error string.

roger2hk: Yes, not sure what the error message will contain.

roger2hk (approved):

PR #1015, opened 2026-06-19
Implement MirrorTarget AddEntries

AlCutter:
This PR implements more of the `MirrorTarget` lifecycle.

Towards #945

### DIFF
--- a/mirror_lifecycle_test.go
+++ b/mirror_lifecycle_test.go
@@ -175,7 +175,7 @@ func TestMirrorTarget_AddEntries_CompleteUpload(t *testing.T) {
 	}
 
 	nextEntry, pendingSize, _, cosigs, err := mt.AddEntries(ctx, testUploadStart, testUploadEnd, validTicket, func() (*MirrorPackage, error) {
-		return &MirrorPackage{Entries: [][]byte{[]byte("e1")}}, io.EOF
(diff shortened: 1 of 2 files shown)


mirror_lifecycle_test.go
```diff
@@ -0,0 +1,192 @@
...
+		cpSource: func(ctx context.Context) ([]byte, error) { return []byte(testPendingCP), nil },
+	}
+
+	validTicket, err := mt.sealTicket(ctx, &ticket{PendingCP: []byte(testPendingCP)})
+	if err != nil {
+		t.Fatalf("sealTicket failed: %v", err)
+	}
+
+	nextEntry, pendingSize, _, cosigs, err := mt.AddEntries(ctx, testUploadStart, testUploadEnd, validTicket, func() (*MirrorPackage, error) {
+		return &MirrorPackage{Entries: [][]byte{[]byte("e1")}}, io.EOF
```
roger2hk: Is it possible to return both package and EOF error at the same time?

roger2hk (approved):

mirror_lifecycle_test.go
AlCutter: Good spot, not currently... removed the returned package since it's not used anyway.

Issue #1010, opened 2026-06-19
`tlog-mirror` Hammer Load Testing Tool


roger2hk:
**Description**

This hammer sets up write traffic to a mirror to test correctness and performance under load.
The write traffic is sent according to the [tlog-mirror](https://c2sp.org/tlog-mirror) spec, and thus could be used to load test any tlog-mirror server.

The target log must support `POST` requests to both `/add-checkpoint` and `add-entries` paths.

roger2hk:
The `tlog-mirror` hammer tool is ready. Read this [README](https://github.com/transparency-dev/tessera/blob/main/internal/mirror/hammer/README.md) to get started.

PR #1016, opened 2026-06-19
Integrate tlog-mirror client into mirror hammer

roger2hk:
Towards https://github.com/transparency-dev/tessera/issues/945 and https://github.com/transparency-dev/tessera/issues/1010.

Not sure if it is a good idea to create the `LatestRaw()` function to get the raw checkpoint.

### DIFF
--- a/internal/mirror/hammer/hammer.go
+++ b/internal/mirror/hammer/hammer.go
@@ -41,7 +41,7 @@ import (
 )
 
 func init() {
-	flag.Var(&mirrorURL, "mirror_url", "Root URL for writing to a mirror (can be specified multiple times), e.g. https://log.server/and/path/ (optional, defaults to log_url)")
+	flag.Var(&mirrorURL, "mirror_url", "Root URL for writing to a mirror (can be specified multiple times), e.g. https://log.server/and/path/")
 }
 
 var (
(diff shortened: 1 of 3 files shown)


internal/mirror/hammer/hammer.go
```diff
@@ -31,29 +32,24 @@ import (
...
 	"github.com/transparency-dev/tessera/storage/posix"
 	"golang.org/x/net/http2"
 
 	"log/slog"
 )
 
 func init() {
-	flag.Var(&logURL, "log_url", "Log storage root URL (can be specified multiple times), e.g. https://log.server/and/path/")
-	flag.Var(&writeLogURL, "write_log_url", "Root URL for writing to a log (can be specified multiple times), e.g. https://log.server/and/path/ (optional, defaults to log_url)")
+	flag.Var(&mirrorURL, "mirror_url", "Root URL for writing to a mirror (can be specified multiple times), e.g. https://log.server/and/path/ (optional, defaults to log_url)")
```
AlCutter: Description needs an update, I think - it references `log_url`, which was removed.

AlCutter (approved):

internal/mirror/hammer/hammer.go
roger2hk: Good catch. Removed the last bit in the description.

PR #1017, opened 2026-06-23
Add basic POSIX mirror tree support

AlCutter:
This PR adds a first cut at fleshing out the POSIX storage support for MTC mirrors.

Towards #945

### DIFF
--- a/storage/posix/files.go
+++ b/storage/posix/files.go
@@ -16,6 +16,7 @@ package posix
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -24,6 +25,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"sync"
 	"syscall"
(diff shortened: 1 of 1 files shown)


storage/posix/files.go
```diff
@@ -1181,13 +1183,130 @@ func (m *MirrorWriter) initialise(ctx context.Context) error {
...
+	m.s.mu.Lock()
+	unlock, err := m.s.lockFile(ctx, treeStateLock)
+	if err != nil {
+		panic(err)
+	}
+	defer func() {
+		if err := unlock(); err != nil {
+			panic(err)
+		}
+		m.s.mu.Unlock()
```
roger2hk: Same comment as the previous pull request. Immediately `defer` the mutex unlock after it is locked.

storage/posix/files.go
```diff
@@ -1181,13 +1183,130 @@ func (m *MirrorWriter) initialise(ctx context.Context) error {
...
+
+	return nil
+}
+
+func (m *MirrorWriter) fetchLeafHashes(ctx context.Context, from, to, sourceSize uint64) ([][]byte, error) {
+	const maxBundles = 300
+
+	lh := make([][]byte, 0, maxBundles)
+	n := 0
+	for ri := range layout.Range(from, to, sourceSize) {
```
roger2hk:
```suggestion
	for ri := range layout.Range(from, to - from, sourceSize) {
```

https://github.com/transparency-dev/tessera/blob/20980b05a3a12ff5a0fb1bb95c549e8967bf2599/api/layout/paths.go#L45-L49

storage/posix/files.go
```diff
@@ -1181,13 +1183,130 @@ func (m *MirrorWriter) initialise(ctx context.Context) error {
...
+	return slices.Clip(data), nil
+}
+
 // IntegrateBundles integrates a sequence of entry bundles into the tree, starting at the provided bundle index bundleIdx.
 // Returns the new size and root hash of the tree if successful.
 func (m *MirrorWriter) IntegrateBundles(ctx context.Context, bundleIdx uint64, bundles iter.Seq[api.EntryBundle]) (uint64, []byte, error) {
-	return 0, nil, errors.New("unimplemented")
+	targetSize := bundleIdx * layout.EntryBundleWidth
+	for b := range bundles {
+		p := len(b.Entries)
```
roger2hk, line 1215: If the number of entries is `0`, it will be treated as a full tile writing empty bundle data to the full tile file.

storage/posix/files.go
```diff
@@ -1181,13 +1183,130 @@ func (m *MirrorWriter) initialise(ctx context.Context) error {
 	return nil
 }
 
+// marshalTlogEntryBundle returns a tlog-tiles compatible serialization of the provided entrybundle.
+func marshalTlogEntryBundle(b api.EntryBundle) ([]byte, error) {
+	// Max size we could possibly write out.
+	data := make([]byte, 0, len(b.Entries)*(2+1<<16))
```
roger2hk, line 1192: Note that this preallocation would consume around 16 MiB for a full bundle.

storage/posix/files.go
```diff
@@ -1181,13 +1183,130 @@ func (m *MirrorWriter) initialise(ctx context.Context) error {
...
+		slog.DebugContext(ctx, "Integrated to", slog.Uint64("size", m.curSize))
+	}
+
+	return nil
+}
+
+func (m *MirrorWriter) fetchLeafHashes(ctx context.Context, from, to, sourceSize uint64) ([][]byte, error) {
+	const maxBundles = 300
+
+	lh := make([][]byte, 0, maxBundles)
```
roger2hk: There are 256 hashes in one bundle. Should this be `maxBundles * layout.EntryBundleWidth` instead of `maxBundles`?

storage/posix/files.go
AlCutter: d'oh, this was grandfathered in. Fixed.

storage/posix/files.go
AlCutter: Good spot, fixed, and factored this func out to be shared with `MigrationTarget` too (which was also affected).

storage/posix/files.go
AlCutter: Not sure how this could happen in a valid request, but added a check to be defensive.
Also added a check that we can't get partial sized bundles mid-stream.

storage/posix/files.go
AlCutter: That's correct - I went back and forth on this for a bit, in the end I went with this since it's "only" 16MB, and there should only really ever ~one "live" alloc of this per mirror.

Left a comment about it in case we want to go back in the future and change to a more conservative default and let Go's slice reallocator sort it out.

storage/posix/files.go
AlCutter: Good catch, updated.

storage/posix/files.go
```diff
@@ -1178,13 +1184,120 @@ func (m *MirrorWriter) initialise(ctx context.Context) error {
...
+
+func (m *MirrorWriter) buildTree(ctx context.Context, targetSize uint64) error {
+	// Double locking:
+	// - The mutex `Lock()` ensures that multiple concurrent calls to this function within a task are serialised.
+	// - The POSIX `lockFile()` ensures that distinct tasks are serialised.
+	m.s.mu.Lock()
+	defer m.s.mu.Unlock()
+	unlock, err := m.s.lockFile(ctx, treeStateLock)
+	if err != nil {
+		panic(err)
```
roger2hk, line 1262: If the `ctx` is cancelled or timed out, this will trigger the panic.

storage/posix/files.go
roger2hk: This function is public, so just in case.

roger2hk (approved):

storage/posix/files.go
AlCutter: As in the comment on the other PR, I'll leave these consistent as panics for now, and we can decide what to do about them in a separate PR.

PR #1020, opened 2026-06-25
[POSIX] Clean up temp files in error cases

mhutchinson:


storage/posix/file_ops.go
```diff
@@ -186,17 +195,31 @@ func createTemp(prefix string, d []byte) (name string, err error) {
...
 	defer func() {
-		if errC := f.Close(); errC != nil && err == nil {
-			err = errC
+		if errC := f.Close(); errC != nil {
+			if err == nil {
+				err = errC
+			}
+			success = false
 		}
 	}()
```
AlCutter: This sorta looks like it's going to defeat the point of the `createTemp` function by deleting the file it just created?

Note that the comment on the func says it's on the _caller_ to remove the temporary file once they've finished with it.

Maybe you could have `createTemp` return a "cleanup" func which the caller can use?

storage/posix/file_ops.go
```diff
@@ -151,10 +151,19 @@ func overwrite(name string, d []byte) error {
 		if err != nil {
 			return fmt.Errorf("failed to create temp file: %w", err)
 		}
+		success := false
```
AlCutter: I wonder if this would be cleaner if you named the `error` in the func signature and just check for `err != nil` in the deferred func - that's a pattern I've seen used for this sort of thing.

storage/posix/file_ops.go
mhutchinson: This only deletes files that would otherwise get leaked when we return an error. I agree it's hard to read though. I think dropping the success variable and using a named error would help. But would it completely mitigate your concern?

e.g. before we tried to create a temp file and write to it. The create worked, but the write didn't work (perhaps threw an error, or the len check didn't work out). We don't give the filename to the user, but before this change the created file stays around. Whatever the problem was on disk is probably now getting worse.

AlCutter: Gotcha, ok - yeah, I think named returns vars would help, I think you could eliminate `tmpName` too that way?

mhutchinson: Latest commits use named error returns, and `tmpName` has gone. PTAL?

AlCutter (approved):

storage/posix/file_ops.go
mhutchinson: This is done now.

PR #1026, opened 2026-06-29
Add `Checkpoint()` to return the checkpoint bytes in fsck tool

roger2hk:


fsck/fsck.go
```diff
@@ -166,6 +170,11 @@ func (f *Fsck) Check(ctx context.Context) error {
 	return nil
 }
 
+// VerifiedCheckpointRaw returns the raw checkpoint as bytes after the checkpoint was fetched and verified.
+func (f *Fsck) VerifiedCheckpointRaw() []byte {
```
AlCutter: Maybe just call it `VerifiedCheckpoint`?

AlCutter (approved):

fsck/fsck.go
```diff
@@ -166,6 +170,11 @@ func (f *Fsck) Check(ctx context.Context) error {
 	return nil
 }
 
+// VerifiedCheckpointRaw returns the raw checkpoint as bytes after the checkpoint was fetched and verified.
```
AlCutter: Actually, maybe just call the field and this func `checkpoint`, without the `verified` bit?

This checkpoint can be accessed even if the fsck fails, which is good, but that could be a bit confusing as to what "verified" means... Since we're returning a serialized checkpoint, the caller will have to `note.Open` it themselves, so if verification just means that fsck used the `note.Verifier` to verify the signature, then the caller is about to do the same verification themselves to get access to the checkpoint's fields.

roger2hk: Done. Removed the `verified` bit.

fsck/fsck.go
roger2hk: Done. Removed the `raw` bit.

AlCutter (approved):

PR #1028, opened 2026-07-01
Multi sig witness

AlCutter:
This PR improves the way the Tessera witness client handles policies which require multiple cosignatures from the same URL (e.g. a witness which is returning both `ed25519` and `MLDSA cosignatures`).

Previously, such a policy would have caused multiple `add-checkpoint` requests to be sent to the witness, one for each `vkey` in the policy, and all but the first of those would be guaranteed to receive a `409` in response causing more requests to be sent.

With the changes in this PR, Tessera should make at most a single call to each witness URL present in the policy (conflicts not withstanding).

I've added a new `WitnessEndpoint` func on the witness policy interface so as to avoid breaking semver by changing the func signature. In practice, I doubt anyone uses the old `Endpoints` func directly (as opposed to just creating the policy structure and passing it directly to `WithWitnesses`), but...

### DIFF
--- a/internal/witness/client.go
+++ b/internal/witness/client.go
@@ -23,8 +23,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
 	"net/http"
 	"os"
+	"slices"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -287,7 +290,7 @@ func (pf *sharedConsistencyProofFetcher) ConsistencyProof(ctx context.Context, s
 // It also has the size of the checkpoint that the log thinks that the witness last signed.
 // This is important for sending update proofs.
 // This is defaulted to zero on startup and calibrated after the first request, which is expected by the spec:
-// `If a client doesn't have information on the latest cosigned checkpoint, it MAY initially make a request with a old size of zero to obtain it`
+// `If a client doesn't have information on the latest cosigned checkpoint, it MAY initially make a request with an old size of zero to obtain it`
 type witnessClient struct {
 	client    *http.Client
 	url       string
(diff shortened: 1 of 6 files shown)


internal/witness/client.go
```diff
@@ -132,6 +135,26 @@ func NewWitnessGateway(group WitnessGroup, client *http.Client, oldSize uint64,
...
+func dedupVerifiers(vs []note.Verifier) []note.Verifier {
+	type verifierKey struct {
+		name string
+		hash uint32
+	}
+
+	seen := make(map[verifierKey]bool)
+	var dedup []note.Verifier
+	for _, v := range vs {
+		k := verifierKey{name: v.Name(), hash: v.KeyHash()}
```
roger2hk, line 152: Could `v` be `nil` here? If so, add a defensive nil check.

```go
    if v == nil {
        continue
    }
```

internal/witness/client.go
```diff
@@ -266,10 +289,10 @@ func (pf *sharedConsistencyProofFetcher) ConsistencyProof(ctx context.Context, s
 // This is defaulted to zero on startup and calibrated after the first request, which is expected by the spec:
 // `If a client doesn't have information on the latest cosigned checkpoint, it MAY initially make a request with a old size of zero to obtain it`
```
roger2hk:
```suggestion
// `If a client doesn't have information on the latest cosigned checkpoint, it MAY initially make a request with an old size of zero to obtain it`
```

internal/witness/client.go
```diff
@@ -279,7 +302,7 @@ func (w *witnessClient) update(ctx context.Context, cp []byte, size uint64, fetc
 			recursed = v.(uint)
 		}
 		if recursed >= maxUpdateRecursion {
-			return nil, fmt.Errorf("too many consecutive requests to witness %s", w.verifier.Name())
+			return nil, fmt.Errorf("too many consecutive requests to witness %s", w.verifiers[0].Name())
```
roger2hk: Would putting the first verifier in the error log be confusing later?

internal/witness/client.go
```diff
@@ -324,11 +347,15 @@ func (w *witnessClient) update(ctx context.Context, cp []byte, size uint64, fetc
 			signed := make([]byte, len(cp)+len(rb))
 			copy(signed, cp)
 			copy(signed[len(cp):], rb)
-			if n, err := note.Open(signed, note.VerifierList(w.verifier)); err != nil {
-				return nil, fmt.Errorf("witness %q at %q replied with invalid signature: %q\nconstructed note: %q\nerror: %v", w.verifier.Name(), w.url, rb, string(signed), err)
+			if n, err := note.Open(signed, note.VerifierList(w.verifiers...)); err != nil {
+				return nil, fmt.Errorf("witness %q at %q replied with invalid signature: %q\nconstructed note: %q\nerror: %v", w.verifiers[0].Name(), w.url, rb, string(signed), err)
```
roger2hk: Would putting the first verifier in the error log be confusing later?

roger2hk (approved):

internal/witness/client.go
AlCutter: In theory no, because the only use case for this currently is for a witness which returns multiple signatures, so all names should be identical. I've enforced this this is the case when parsing from policy at least.

That said, I could imagine this also supporting some sort of external witness gateway thing in the future, so I've changed to collate & uniq the names into a string here, and used that.

internal/witness/client.go
AlCutter: (same)

PR #1031, opened 2026-07-02
Handle unaligned uploadStart

AlCutter:
This PR adds support for mirror uploads which don't start on an entrybundle boundary.

Towards #945

### DIFF
--- a/mirror_lifecycle.go
+++ b/mirror_lifecycle.go
@@ -296,7 +296,7 @@ func (mt *MirrorTarget) AddEntries(ctx context.Context, uploadStart, uploadEnd u
 	}
 }
 
-// bundleIterator returns an interator which yields entry bundles after verifying their subtree consistency with the provided pending checkpoint.
+// bundleIterator returns an iterator which yields entry bundles after verifying their subtree consistency with the provided pending checkpoint.
 //
 // Yielded entry bundles are always aligned to bundle boundaries. Specifically, this means that if the provided start is _not_ bundle aligned, then we will
 // fetch entries from the bundle at start/256 and use those entries to left-pad the first yielded bundle.
@@ -316,6 +316,10 @@ func (mt *MirrorTarget) bundleIterator(ctx context.Context, next func() (*Mirror
 			b := &api.EntryBundle{}
 			if err := b.UnmarshalText(br); err != nil {
 				yield(nil, fmt.Errorf("failed to unmarshal bundle containing uploadStart (%d): %v", start, err))
+				return
+			}
+			if l := len(b.Entries); l < int(p) {
+				yield(nil, fmt.Errorf("POTENTIAL CORRUPTION: partial bundle at index %d.%d has only %d entries", start/layout.EntryBundleWidth, p, l))
 				return
 			}
 			padEntries = b.Entries[:p]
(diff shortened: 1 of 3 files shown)


storage/posix/files.go
```diff
@@ -1233,7 +1234,8 @@ func (m *MirrorWriter) IntegrateBundles(ctx context.Context, bundleIdx uint64, b
 		if err != nil {
 			return 0, nil, fmt.Errorf("failed to marshal bundle at index %d: %v", bundleIdx, err)
 		}
-		// ...and then write it out.
+
+		// Now write the bundle out out.
```
roger2hk: I guess you only want one `out`.

mirror_lifecycle.go
```diff
@@ -298,29 +296,59 @@ func (mt *MirrorTarget) AddEntries(ctx context.Context, uploadStart, uploadEnd u
 	}
 }
 
-// bundleIterator returns an interator which yields MirrorPackages after verifying their subtree consistency with the provided pending checkpoint.
+// bundleIterator returns an interator which yields entry bundles after verifying their subtree consistency with the provided pending checkpoint.
```
roger2hk:
```suggestion
// bundleIterator returns an iterator which yields entry bundles after verifying their subtree consistency with the provided pending checkpoint.
```

mirror_lifecycle.go
```diff
@@ -298,29 +296,59 @@ func (mt *MirrorTarget) AddEntries(ctx context.Context, uploadStart, uploadEnd u
...
+				yield(nil, fmt.Errorf("failed to read bundle containing uploadStart (%d): %v", start, err))
+				return
+			}
+			// Parse and clip, just in case we were returned data from a full tile.
+			b := &api.EntryBundle{}
+			if err := b.UnmarshalText(br); err != nil {
+				yield(nil, fmt.Errorf("failed to unmarshal bundle containing uploadStart (%d): %v", start, err))
+				return
+			}
+			padEntries = b.Entries[:p]
```
roger2hk, line 325: Can the client side trigger the mirror server runtime panic (i.e. slice bounds out of range) by providing entries length less than `p`?

storage/posix/files.go
AlCutter: Weeeell, we _do_ really _really_ want to be writing it out. Like, _out_ out. :)

mirror_lifecycle.go
AlCutter: I don't think so, not unless there's been something gone quite wrong before - we read a partial bundle which matches the offset of `uploadStart` into that bundle. If there aren't enough entries to trim to `p` from that bundle, then the log is probably corrupt. OTOH, with garbage collection, we could (at least theoretically) end up being returned a full tile with more than `p` entries so need to trim it down.

I've added a defensive check to avoid the panic in this situation, though, as unlikely as it is, it would mean that one corrupt mirror tree would potentially make all the others served by the same process unavailable too.

roger2hk: Gotcha, the partial bundle comes from the mirror's log storage. It's all good if the client couldn't trigger panic in the mirror server.

roger2hk (approved):

PR #1035, opened 2026-07-03
Enforce that same key can't be used for witness and mirror cosigs

AlCutter:
Towards #945

### DIFF
--- a/cmd/mtc/mirror/posix/main.go
+++ b/cmd/mtc/mirror/posix/main.go
@@ -65,6 +65,9 @@ func main() {
 
 	mirrorCosigner := mustCreateCosigner(ctx, *mirrorCosignerPath)
 
+	// Mirror cosigner MUST NOT use the same key as add-checkpoint cosigner.
+	assertDistinctSigners(ctx, w.Signers, []note.Signer{mirrorCosigner})
+
 	mux := handler.NewMirrorMux()
 	cfg := mirrorConfigFromFlags(ctx)
 	for _, l := range cfg.Logs {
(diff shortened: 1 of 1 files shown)


cmd/mtc/mirror/posix/main.go
```diff
@@ -216,3 +219,21 @@ func mustCreateCosigner(ctx context.Context, path string) note.Signer {
 	}
 	return s
 }
+
+// assertDistinctSigners asserts that the two provided lists of cosigners have no signers in common.
+func assertDistinctSigners(ctx context.Context, w, m []note.Signer) {
+	type nhKey struct {
+		name string
+		hash uint32
+	}
```
roger2hk, line 228: What if there are keys with different names but the same key hash?

cmd/mtc/mirror/posix/main.go
```diff
@@ -216,3 +219,21 @@ func mustCreateCosigner(ctx context.Context, path string) note.Signer {
...
+	aMap := make(map[nhKey]struct{}, len(w))
+	for _, s := range w {
+		aMap[nhKey{name: s.Name(), hash: s.KeyHash()}] = struct{}{}
+	}
+	for _, s := range m {
+		if _, ok := aMap[nhKey{name: s.Name(), hash: s.KeyHash()}]; ok {
+			slog.ErrorContext(ctx, "Cannot use same signing key for witness and mirror", slog.String("name", s.Name()), slog.String("hash", fmt.Sprintf("%0x", s.KeyHash())))
```
roger2hk: Should we use `%08x` for the key hash to align with the `note` package?

roger2hk (approved):

cmd/mtc/mirror/posix/main.go
AlCutter: That's _very_ unlikely (1 in 2^31) to have two non-identical keys with the same hash. 

Not impossible, but I'd argue that it's better to have a tiny chance of having to regenerate one of your new keys, but in doing so gain 100% certainty of avoiding a catastrophic outcome by accidentally passing the same key file to both flags.

cmd/mtc/mirror/posix/main.go
AlCutter: Sure, pretty sure we discovered the other day this will make no difference to the output, but may as well make the code match.

PR #1036, opened 2026-07-03
Refactor ticket handling

AlCutter:
Extracts the ticket handling from the AddEntries method.

This has a couple of benefits:
1. Reduction in cognitive load - these methods are smaller and more tightly scoped.
2. In the case where no ticket was provided, but `uploadEnd` matches the current pending checkpoint from the witness, we now continue to process the request. Previously, we returned an `ErrConflict` causing an unnecessary round-trip.

Towards #945

### DIFF
--- a/mirror_lifecycle.go
+++ b/mirror_lifecycle.go
@@ -208,47 +208,11 @@ type MirrorPackage struct {
 // ticket for future invocations, and, optionally, a cosignature over a pending checkpoint
 // whose size matches uploadEnd if one exists.
 func (mt *MirrorTarget) AddEntries(ctx context.Context, uploadStart, uploadEnd uint64, ticketBytes []byte, next func() (*MirrorPackage, error)) (nextEntry uint64, pendingSize uint64, newTicket []byte, cosigs []byte, err error) {
-	curIntegratedSize, err := mt.reader.IntegratedSize(ctx)
+	nextEntry, pendingSize, ticket, pendingCP, pendingNote, err := mt.openOrCreateTicket(ctx, ticketBytes, uploadEnd)
 	if err != nil {
-		return 0, 0, nil, nil, fmt.Errorf("failed to read integrated size: %w", err)
-	}
-	ticketCP, err := mt.openTicket(ticketBytes)
-	if err != nil {
-		slog.DebugContext(ctx, "Invalid Ticket received, returning new one", slog.Any("error", err), slog.Uint64("uploadStart", uploadStart), slog.Uint64("uploadEnd", uploadEnd))
-
-		// If the client didn't provide a [valid] ticket, then we don't have a pending
-		// checkpoint to validate against, so we return a new ticket with the
-		// current checkpoint.
-		pendingCPRaw, err := mt.cpSource(ctx)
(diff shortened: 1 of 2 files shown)


mirror_lifecycle.go
```diff
@@ -426,26 +389,85 @@ func (mt *MirrorTarget) IntegratedSize(ctx context.Context) (uint64, error) {
...
+
+	ticketCP, err := mt.open(ticketBytes)
+	if err != nil {
+		slog.DebugContext(ctx, "Failed to open ticket", slog.Any("error", err))
+	} else {
+		pendingCP, _, pendingNote, err = log.ParseCheckpoint(ticketCP, mt.logVerifier.Name(), mt.logVerifier)
+		if err != nil {
+			slog.DebugContext(ctx, "Failed to parse ticket", slog.Any("error", err))
+		}
+		slog.DebugContext(ctx, "Valid ticket", slog.Uint64("nextEntry", nextEntry), slog.Uint64("pendingSize", pendingCP.Size))
```
roger2hk: If `log.ParseCheckpoint` returns any `err`, the `pendingCP` would be `nil`. It would trigger a nil pointer deference panic on L412.

mirror_lifecycle.go
```diff
@@ -283,16 +247,15 @@ func (mt *MirrorTarget) AddEntries(ctx context.Context, uploadStart, uploadEnd u
...
+		return nextEntry, pendingSize, nil, sigs, nil
+	case nextEntry > pendingSize:
+		slog.WarnContext(ctx, "nextEntry > pendingSize", slog.Uint64("nextEntry", nextEntry), slog.Uint64("pendingSize", pendingSize))
+		return nextEntry, pendingSize, nil, nil, nil
 	default:
-		slog.WarnContext(ctx, "Incomplete upload", slog.Uint64("nextEntry", nextEntry), slog.Uint64("pendingSize", pendingCP.Size))
+		slog.WarnContext(ctx, "Incomplete upload", slog.Uint64("nextEntry", nextEntry), slog.Uint64("pendingSize", pendingSize))
 		// Incomplete upload, return an updated ticket with the current checkpoint.
-		return nextEntry, pendingCP.Size, ticketBytes, nil, nil
+		return nextEntry, pendingSize, ticketBytes, nil, nil
```
roger2hk, line 260: The comment said `an updated ticket`. Should that be `ticket` instead of `ticketBytes`?

AlCutter: No really - the comment is talking in the abstract about what we're doing, the ticket we're returning happens to be stored in `ticketBytes`.

roger2hk: The `ticketBytes` is never changed when passing into the `AddEntries` function. Do you mean the `ticketBytes` was already updated out of `AddEntries` function?

roger2hk (approved):

mirror_lifecycle.go
AlCutter: Sorry, I see what you're saying now. I've removed the named return params as it was causing confusion, and, as you pointed out, a bug!

PR #1040, opened 2026-07-03
storage/gcp: add optional SpannerTablePrefix to Config

eperrine-ant:
## What

Adds an optional `SpannerTablePrefix` field to the GCP storage `Config`, which is
prepended to the names of all Spanner tables used for log coordination state
(`Tessera`, `SeqCoord`, `Seq`, `IntCoord`, `PubCoord`, `GCCoord`). The same option is
added to the GCP antispam implementation as `AntispamOpts.SpannerTablePrefix` (for its
`IDSeq` and `FollowCoord` tables).

## Why

This allows multiple logs to share a single Spanner database, with each log's
coordination state in its own set of tables — mirroring the existing `BucketPrefix`
option for GCS objects. Useful for multi-tenant deployments where provisioning a
Spanner database per log is wasteful.

## Notes

- The prefix defaults to empty, which preserves the existing table names and
  behaviour unchanged for current deployments.
- Non-empty prefixes are validated (must start with a letter; letters, digits, or
  underscores only; at most 64 characters) so that prefixed table names remain valid
  Spanner identifiers, and so the value is safe to interpolate into the
  `CREATE TABLE` DDL.
- The prefix namespaces only the Spanner state; logs sharing a database must still
  use a distinct `Bucket`/`BucketPrefix` each, since GCS object paths are unaffected.

## Testing

- New `TestSpannerTablePrefix` exercises a prefixed coordinator end-to-end against
  `spannertest` and asserts state lands in the prefixed tables only.
- New `TestSpannerTablePrefixValidation` covers accepted/rejected prefixes,
  including DDL metacharacters.
- Existing tests pass unchanged with the default empty prefix.

### DIFF
--- a/storage/gcp/antispam/gcp.go
+++ b/storage/gcp/antispam/gcp.go
@@ -28,6 +28,7 @@ import (
 	"io"
 	"iter"
 	"os"
+	"regexp"
 	"sync/atomic"
 	"time"
 
@@ -72,7 +73,18 @@ type AntispamOpts struct {
 	// When the antispam follower is at least this many entries behind the size of the locally integrated tree,
 	// the antispam decorator will return a wrapped tessera.ErrPushback for every Add request.
 	PushbackThreshold uint
+
+	// SpannerTablePrefix is an optional prefix to prepend to the names of all Spanner tables.
+	// This can be used e.g. to store the antispam state for multiple logs in the same Spanner database.
+	// If set, it must start with a letter, contain only letters, digits, or underscores
+	// (e.g. "log1_"), and be at most 64 characters long, so that prefixed table names
+	// remain valid Spanner identifiers.
+	SpannerTablePrefix string
 }
+
+// tablePrefixRE matches valid values for a Spanner table prefix: empty, or a leading
+// letter followed by up to 63 letters, digits, or underscores.
+var tablePrefixRE = regexp.MustCompile(`^([A-Za-z][A-Za-z0-9_]{0,63})?$`)
 
 // NewAntispam returns an antispam driver which uses Spanner to maintain a mapping of
 // previously seen entries and their assigned indices.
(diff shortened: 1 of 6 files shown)


AlCutter:
Hi @eperrine-ant,

Thanks for the request, this is _doable_, however I'd love to hear a bit more about your use-case (if possible, feel free to get in touch directly if that's easier), particularly around tenant log lifecycle as there are trade-offs to this approach.

One of the things we learned from operating Trillian logs (which were multi-tenant from the outset) is that deleting old logs is more painful & potentially risky than it needs to be.
For Tessera, we decided to avoid this scenario by having each log have its own Spanner Database. For the turn-up/turn-down scenario this translates into creation or removal of a Spanner Database resource.

It's entirely up to you whether you create whole new Spanner instances for each Spanner DB, or choose to create each Spanner DB in the same Spanner instance. 

Your PR is _kinda_ re-implementing this strategy of having multiple Spanner DBs inside a single Spanner Instance by creating new tables for each log. I think it's _probably_ cleaner/more idiomatic and safer to just create the new DB instances (e.g. you can add IAM controls by database), but I'd love to know if/why this might not work for you.

storage/gcp/antispam/gcp.go
```diff
@@ -72,8 +73,19 @@ type AntispamOpts struct {
 	// When the antispam follower is at least this many entries behind the size of the locally integrated tree,
 	// the antispam decorator will return a wrapped tessera.ErrPushback for every Add request.
 	PushbackThreshold uint
+
+	// SpannerTablePrefix is an optional prefix to prepend to the names of all Spanner tables.
+	// This can be used e.g. to store the antispam state for multiple logs in the same Spanner database.
+	// If set, it must start with a letter, contain only letters, digits, or underscores
+	// (e.g. "log1_"), and be at most 64 characters long, so that prefixed table names
+	// remain valid Spanner identifiers.
```
AlCutter, line 81: We should probably recommend that these prefixes are derived from the log origin strings to make it easy for the operator to associate tables with specific logs.

(same for the similar option field in ../gcp.go)

storage/gcp/gcp.go
```diff
@@ -237,11 +254,11 @@ func (s *Storage) Appender(ctx context.Context, opts *tessera.AppendOptions) (*t
 			return nil, nil, fmt.Errorf("failed to connect to Spanner: %v", err)
 		}
 	}
-	if err := initDB(ctx, s.cfg.Spanner); err != nil {
+	if err := initDB(ctx, s.cfg.Spanner, s.cfg.SpannerTablePrefix); err != nil {
 		return nil, nil, fmt.Errorf("failed to verify/init Spanner schema: %v", err)
 	}
 
-	seq, err := newSpannerCoordinator(ctx, s.cfg.SpannerClient, uint64(opts.PushbackMaxOutstanding()))
+	seq, err := newSpannerCoordinator(ctx, s.cfg.SpannerClient, s.cfg.SpannerTablePrefix, uint64(opts.PushbackMaxOutstanding()))
```
AlCutter: You could do something like this to avoid having to add `tbl()` funcs but also to make the approach in `initDB` a more consistent with how it works in the various coordinator methods:

```suggestion
    table := func(t string) string {
      return s.cfg.SpannerTablePrefix + t
    }
	if err := initDB(ctx, s.cfg.Spanner, table); err != nil {
return nil, nil, fmt.Errorf("failed to verify/init Spanner schema: %v", err)
}

	seq, err := newSpannerCoordinator(ctx, s.cfg.SpannerClient, table, uint64(opts.PushbackMaxOutstanding()))
```

storage/gcp/gcp.go
```diff
@@ -157,10 +158,26 @@ type Config struct {
 	BucketPrefix string
 	// Spanner is the GCP resource URI of the spanner database instance to use.
 	Spanner string
+	// SpannerTablePrefix is an optional prefix to prepend to the names of all Spanner tables.
```
AlCutter, line 161: I'm wondering whether we should (probably later) have a mechanism to set both `BucketPrefix` and `SpannerTablePrefix` at the same time from the same string, might avoid some foot-guns.

Maybe just pop in a `TODO` to think about this/decide based on how folks use the opts?

eperrine-ant: added TODO

storage/gcp/gcp.go
eperrine-ant: done

storage/gcp/antispam/gcp.go
eperrine-ant: agreed, done

AlCutter (approved):

AlCutter:
Many thanks for the contribution @eperrine-ant!

PR #1041, opened 2026-07-06
Handle no pending checkpoint in MTC Mirror

roger2hk:
This pull request removes `ErrNoPendingCheckpoint` and returns `409 Conflict` instead of `400 Bad Request` when there is no pending checkpoint.

Towards #945.

### DIFF
--- a/mirror_lifecycle.go
+++ b/mirror_lifecycle.go
@@ -47,6 +47,9 @@ var (
 	// ErrConflict is returned when the requested upload range conflicts with the
 	// current state of the log.
 	ErrConflict = errors.New("tree size conflict")
+	// ErrNoPendingCheckpoint is returned when a pending checkpoint cannot be
+	// determined.
+	ErrNoPendingCheckpoint = errors.New("no pending checkpoint")
 	// ErrInvalidProof is returned when a proof fails to verify.
 	ErrInvalidProof = errors.New("invalid proof")
 )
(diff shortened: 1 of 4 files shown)


mirror_lifecycle.go
```diff
@@ -224,7 +221,10 @@ func (mt *MirrorTarget) AddEntries(ctx context.Context, uploadStart, uploadEnd u
 			return 0, 0, nil, nil, fmt.Errorf("failed to get pending checkpoint: %v", err)
 		}
 		if len(pendingCPRaw) == 0 {
-			return 0, 0, nil, nil, ErrNoPendingCheckpoint
+			if uploadEnd == curIntegratedSize {
+				return curIntegratedSize, curIntegratedSize, ticketBytes, nil, nil
+			}
+			return curIntegratedSize, curIntegratedSize, nil, nil, ErrConflict
```
AlCutter: I think perhaps this should be a `422` - that way we can leave `409` for the cases where we return that "conflict body" response, and, arguably, this request is unprocessable because while the body was well formed, what semantics does your proof convey if there _is_ no pending checkpoint to use it with?

I'd maybe leave `ErrNoPendingCheckpoint` here (so we can tell what's happened at the layer above, if necessary - e.g. metrics), but then change the http handler to return `http.StatusUnprocessableEntity` when it sees that error?

roger2hk: I've reverted to add `ErrNoPendingCheckpoint` back and return "422 Unprocessable Entity" in HTTP handler.

AlCutter (approved):

PR #1049, opened 2026-07-08
Handle reuploads

AlCutter:
This PR updates the mirror to permit a small number of bundle to be re-uploaded by a client, as permitted by the tlog-mirror spec, and updates the POSIX mirror storage to write bundles in an idempotent fashion.

Towards #945


storage/posix/file_ops.go
roger2hk:
```suggestion
			// Wrap the error here because we need to know if it's os.ErrExist at higher levels.
```

roger2hk (approved):

PR #1054, opened 2026-07-10
storage/gcp/antispam: add optional SpannerClient to AntispamOpts

eperrine-ant:
## What

Adds an optional `SpannerClient` field to the GCP antispam `AntispamOpts`. If set, `NewAntispam` uses the provided client instead of creating its own, mirroring the existing `Config.SpannerClient` on the GCP log storage.

## Why

Multi-tenant deployments that store several logs' state in one Spanner database (via `SpannerTablePrefix`, #1040) currently get a separate Spanner client — with its own session pool — for every antispam instance, on top of the log storage's client. Allowing a shared client keeps session usage bounded no matter how many logs share the database.

## Notes

- An injected client must be connected to the database named by the `spannerDB` parameter. This is validated at construction, since a mismatched client would otherwise silently disable deduplication (lookups treat NotFound as a cache miss).
- An injected client is never closed by Tessera; its lifecycle stays with the caller.
- An injected client is also used to apply the initial schema seed mutations. When no client is provided, behavior is unchanged: table preparation uses its own temporary client, and the long-lived client is created after table preparation succeeds.

## Testing

- New shared-client case in `TestAntispamStorage` drives the full append → follow → dedup-lookup cycle through an injected client against `spannertest`.
- New `TestAntispamSharedClientWrongDatabase` asserts a client connected to a different database is rejected.
- Existing tests pass unchanged with the field unset.

### DIFF
--- a/storage/gcp/antispam/gcp.go
+++ b/storage/gcp/antispam/gcp.go
@@ -84,6 +84,11 @@ type AntispamOpts struct {
 	// underscores, and prepending a letter if the origin doesn't start with one), to
 	// make it easy to associate tables with specific logs.
 	SpannerTablePrefix string
+
+	// SpannerClient will be used to interact with Spanner. If unset, Tessera will create one.
+	// If set, it must be connected to the database identified by the spannerDB parameter to
+	// NewAntispam, and it will never be closed by Tessera.
+	SpannerClient *spanner.Client
 }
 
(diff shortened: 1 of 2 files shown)


phbnf:
Thanks for the PR Erin, reassigning to @AlCutter for review.

AlCutter (approved):

PR #1055, opened 2026-07-13
Improve conflict handling in mirror client

roger2hk:
Towards https://github.com/transparency-dev/tessera/issues/945 and https://github.com/transparency-dev/tessera/issues/1010.

### DIFF
--- a/client/mirror/client.go
+++ b/client/mirror/client.go
@@ -27,9 +27,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 
 	"github.com/transparency-dev/tessera/api/layout"
 	"github.com/transparency-dev/tessera/client"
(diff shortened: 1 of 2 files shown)


client/mirror/client.go
```diff
@@ -404,47 +446,30 @@ func (c *Client) buildCheckpointRequestBody(oldSize uint64, proof [][]byte, chec
...
-				return nil, fmt.Errorf("failed to create ProofBuilder: %v", err)
-			}
-			proof, err = pb.ConsistencyProof(ctx, oldSize, targetSize)
-			if err != nil {
-				return nil, fmt.Errorf("failed to generate consistency proof: %v", err)
-			}
+	// Push the checkpoint with the old size (0 if not provided).
+	for c.oldSize < targetSize {
+		err := c.pushCheckpoint(ctx, c.oldSize, targetSize, targetCheckpointRaw)
+		if err == nil {
```
AlCutter: Could this be re-written to indent error/less common flows?

```go
if err != nil {
  if !errors.As(...) {
     return ...
  }
  c.oldSize = ...
  break
}
...
```

AlCutter (approved):

client/mirror/client.go
roger2hk: I've re-written this. It's a bit different from the suggested code with the `break` replaced by `continue`.

AlCutter (approved):

PR #1057, opened 2026-07-13
Ensure partial resources implied by published checkpoint exist

AlCutter:
This PR ensures that when a mirror checkpoint is published, all the partial resources implied by its size are present.

Prior to this PR, two parties could be uploading entries for differently sized checkpoints concurrently, it's possible for the integrated tree size to have moved beyond the smaller sized checkpoint but without having yet reached the larger checkpoint size (e.g. if this client sends almost all entries but then stalls). If the client trying to mirror the _smaller_ checkpoint then sends an `add-entries` request with zero entries, this checkpoint will be published, but the static resources in the tree will not meet geometry implied by the checkpoint size.

Towards #945

### DIFF
--- a/storage/posix/files.go
+++ b/storage/posix/files.go
@@ -1385,26 +1385,35 @@ func (m *MirrorWriter) UpdateCheckpoint(ctx context.Context, fn func(old []byte)
 // If an implied partial resource is not already present, this function will attempt to create
 // it from a strictly larger resource whose presence is implied by treeSize.
 func (m *MirrorWriter) ensureGeometry(ctx context.Context, cpSize, treeSize uint64) error {
+	if cpSize == 0 {
+		return nil
+	}
 	if cpSize > treeSize {
 		return fmt.Errorf("new size %d is greater than integrated size %d", cpSize, treeSize)
 	}
-	levels := uint64(1<<bits.Len64(cpSize-1)) / 8
-	idx := (cpSize - 1) / layout.EntryBundleWidth
-	for l := uint64(0); l < levels; l, idx = l+1, idx>>layout.TileHeight {
+	ml := maxLevel(cpSize)
+	idx := (cpSize - 1) >> layout.TileHeight
+	for l := uint64(0); l <= uint64(ml); l, idx = l+1, idx>>layout.TileHeight {
 		treeP := layout.PartialTileSize(l, idx, treeSize)
 		cpP := layout.PartialTileSize(l, idx, cpSize)
-		if cpP != treeP {
-			if l == 0 {
-				if err := m.ensurePartialBundle(ctx, idx, cpP, treeP); err != nil {
-					return err
-				}
-			}
-			if err := m.ensurePartialTile(ctx, l, idx, cpP, treeP); err != nil {
+		if l == 0 {
+			if err := m.ensurePartialBundle(ctx, idx, cpP, treeP); err != nil {
 				return err
 			}
 		}
+		if err := m.ensurePartialTile(ctx, l, idx, cpP, treeP); err != nil {
+			return err
+		}
 	}
 	return nil
+}
+
+// maxLevel returns the maximum tile level for a tree of the given size.
+func maxLevel(sz uint64) int {
+	if sz == 0 {
+		return 0
+	}
+	return (bits.Len64(sz) - 1) / layout.TileHeight
 }
 
 // ensurePartialBundle ensures that the partial entry bundle implied by the new size is
@@ -1413,6 +1422,10 @@ func (m *MirrorWriter) ensureGeometry(ctx context.Context, cpSize, treeSize uint
 // If the implied partial entry bundle is not already present, this function will attempt to create
 // it from the entry bundle implied by treeSize.
 func (m *MirrorWriter) ensurePartialBundle(ctx context.Context, idx uint64, cpP, treeP uint8) error {
+	if cpP == treeP {
+		return nil
+	}
+
 	// Check if the entry bundle already exists.
 	_, err := m.s.stat(m.logStorage.entriesPath(idx, cpP))
 	switch {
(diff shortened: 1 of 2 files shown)


storage/posix/files.go
```diff
@@ -1351,5 +1363,129 @@ func (m *MirrorWriter) UpdateCheckpoint(ctx context.Context, fn func(old []byte)
...
+		return fmt.Errorf("failed to stat entry bundle @%d.%d: %v", idx, cpP, err)
+	default:
+		// Already exists, we're done!
+		return nil
+	}
+
+	// Read bundle implied by the tree size...
+	d, err := m.logStorage.ReadEntryBundle(ctx, idx, treeP)
+	if err != nil {
+		return fmt.Errorf("failed to read entrybundle @%d.%d: %v", idx, treeP, err)
```
roger2hk:
```suggestion
		return fmt.Errorf("failed to read entry bundle @%d.%d: %v", idx, treeP, err)
```

storage/posix/files.go
```diff
@@ -1351,5 +1363,129 @@ func (m *MirrorWriter) UpdateCheckpoint(ctx context.Context, fn func(old []byte)
...
+	}
+	eb := &api.EntryBundle{}
+	if err := eb.UnmarshalText(d); err != nil {
+		return fmt.Errorf("failed to unmarshal entrybundle @%d.%d: %v", idx, treeP, err)
```
roger2hk:
```suggestion
		return fmt.Errorf("failed to unmarshal entry bundle @%d.%d: %v", idx, treeP, err)
```

storage/posix/files.go
```diff
@@ -1351,5 +1363,129 @@ func (m *MirrorWriter) UpdateCheckpoint(ctx context.Context, fn func(old []byte)
...
+	}
+
+	// Then trim it down to the size implied by the checkpoint, and write it out.
+	eb.Entries = eb.Entries[:cpP]
+	d, err = marshalTlogEntryBundle(eb)
+	if err != nil {
+		return fmt.Errorf("failed to marshal entrybundle @%d.%d: %v", idx, cpP, err)
```
roger2hk:
```suggestion
		return fmt.Errorf("failed to marshal entry bundle @%d.%d: %v", idx, cpP, err)
```

storage/posix/files.go
```diff
@@ -1351,5 +1363,129 @@ func (m *MirrorWriter) UpdateCheckpoint(ctx context.Context, fn func(old []byte)
...
+	}
+	if err := m.logStorage.writeBundleIdempotent(ctx, idx, cpP, d); err != nil {
+		return fmt.Errorf("failed to write entrybundle @%d.%d: %v", idx, cpP, err)
```
roger2hk:
```suggestion
		return fmt.Errorf("failed to write entry bundle @%d.%d: %v", idx, cpP, err)
```

storage/posix/files.go
```diff
@@ -1351,5 +1363,129 @@ func (m *MirrorWriter) UpdateCheckpoint(ctx context.Context, fn func(old []byte)
...
+func (m *MirrorWriter) ensureGeometry(ctx context.Context, cpSize, treeSize uint64) error {
+	if cpSize > treeSize {
+		return fmt.Errorf("new size %d is greater than integrated size %d", cpSize, treeSize)
+	}
+	levels := uint64(1)
+	if cpSize > 1 {
+		levels = 1 << bits.Len64(cpSize-1)
+	}
+	idx := (cpSize - 1) / layout.EntryBundleWidth
+	for l := uint64(0); l < levels; l, idx = l+layout.TileHeight, idx>>layout.TileHeight {
```
roger2hk: Is `l` the tree level or the tile level? I'm confused as `layout.PartialTileSize` expects the tile level while the `for` loop seems to be the tree level.

AlCutter: Thanks Roger, good spot - obviously overcooked my brain in the sun!
Fixed, and I've expanded the test to cover multiple tile levels.

storage/posix/files.go
```diff
@@ -1351,5 +1363,126 @@ func (m *MirrorWriter) UpdateCheckpoint(ctx context.Context, fn func(old []byte)
...
+// ensureGeometry checks that the partial tiles and entry bundles implied by the new size are
+// present in the stored resources.
+//
+// If an implied partial resource is not already present, this function will attempt to create
+// it from a strictly larger resource whose presence is implied by treeSize.
+func (m *MirrorWriter) ensureGeometry(ctx context.Context, cpSize, treeSize uint64) error {
+	if cpSize > treeSize {
+		return fmt.Errorf("new size %d is greater than integrated size %d", cpSize, treeSize)
+	}
+	levels := uint64(1 << bits.Len64(cpSize-1))
```
roger2hk: Is `levels` the tree level? I think we want to have the tile level here.

storage/posix/files.go
```diff
@@ -1351,5 +1363,129 @@ func (m *MirrorWriter) UpdateCheckpoint(ctx context.Context, fn func(old []byte)
...
+// If an implied partial resource is not already present, this function will attempt to create
+// it from a strictly larger resource whose presence is implied by treeSize.
+func (m *MirrorWriter) ensureGeometry(ctx context.Context, cpSize, treeSize uint64) error {
+	if cpSize == 0 {
+		return nil
+	}
+	if cpSize > treeSize {
+		return fmt.Errorf("new size %d is greater than integrated size %d", cpSize, treeSize)
+	}
+	maxLevel := uint64(bits.Len64(cpSize-1) / layout.TileHeight)
```
roger2hk: This is still incorrect for edge cases like 129 or 32,769. See https://go.dev/play/p/NbZubNc5jvx.

```suggestion
	maxLevel := uint64((bits.Len64(cpSize) - 1) / layout.TileHeight)
```

AlCutter: This is embarrassing, sorry. PTAL

roger2hk (approved):

PR #1062, opened 2026-07-16
Enable mirror hammer to use an existing log as a source.

AlCutter:
This PR enables an existing log to optionally be used as the source of entries when hammering a mirror.

Towards #945

### DIFF
--- a/internal/mirror/hammer/hammer.go
+++ b/internal/mirror/hammer/hammer.go
@@ -37,6 +37,8 @@ import (
 	"github.com/transparency-dev/tessera/storage/posix"
 	"golang.org/x/net/http2"
 
+	sdbNote "golang.org/x/mod/sumdb/note"
+
 	"log/slog"
 )
 
@@ -49,6 +51,8 @@ var (
 
 	storageDir = flag.String("storage_dir", "", "Root directory to store log data")
 	logPrivKey = flag.String("log_private_key", "", "Location of private key file")
+	sourceLogURL        = flag.String("source_log_url", "", "URL of the log to mirror, or empty if using a locally created log.")
+	sourceLogPubKeyPath = flag.String("source_log_public_key_path", "", "Path to the public key of the log to mirror. Required if --source_log_url is set.")
 
 	maxWriteOpsPerSecond = flag.Int("max_write_ops", 0, "The maximum number of write operations per second")
 	numWriters           = flag.Int("num_writers", 0, "The number of independent write tasks to run")
@@ -70,6 +74,12 @@ var (
 
 	hc *http.Client
 )
+
+type logReader interface {
+	ReadCheckpoint(ctx context.Context) ([]byte, error)
+	ReadEntryBundle(ctx context.Context, index uint64, p uint8) ([]byte, error)
+	ReadTile(ctx context.Context, level, index uint64, p uint8) ([]byte, error)
+}
 
 func main() {
 	flag.Parse()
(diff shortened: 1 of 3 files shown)


internal/mirror/hammer/loadtest/tui.go
```diff
@@ -74,29 +77,31 @@ func (c *tuiController) Run(ctx context.Context) {
...
-			slog.InfoContext(ctx, "Increasing the number of workers")
-			c.hammer.randomReaders.Grow(ctx)
-			c.hammer.fullReaders.Grow(ctx)
-			c.hammer.writers.Grow(ctx)
-		case 'W':
-			slog.InfoContext(ctx, "Decreasing the number of workers")
-			c.hammer.randomReaders.Shrink(ctx)
-			c.hammer.fullReaders.Shrink(ctx)
-			c.hammer.writers.Shrink(ctx)
+		if c.hammer != nil {
```
roger2hk, line 80: Does it mean there will be no way to control the QPS copying entries from the existing log to the mirror server?

internal/mirror/hammer/hammer.go
```diff
@@ -350,3 +328,103 @@ func runMirrorSync(ctx context.Context, tracker *client.LogStateTracker, mClient
...
+		u, err := url.Parse(logURL)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to parse log URL %q: %v", logURL, err)
+		}
+		lr, err = client.NewHTTPFetcher(u, hc)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to create HTTP fetcher: %v", err)
+		}
+	default:
+		lr = client.FileFetcher{Root: logURL}
```
roger2hk: Would the `logURL := strings.ToLower(logURL)` on L392 cause any issue if the `logURL` contains some uppercase value?

internal/mirror/hammer/hammer.go
```diff
@@ -47,6 +49,9 @@ func init() {
 var (
 	mirrorURL multiStringFlag
 
+	sourceLogURL        = flag.String("source_log_url", "", "URL of the log to mirror, or empty if using a locally created log.")
+	sourceLogPubKeyPath = flag.String("source_log_pubkey_path", "", "Path to the public key of the log to mirror. Required if --source_log_url is set.")
```
roger2hk: Shall we align the naming of this flag with the `log_private_key` flag below? Maybe call it `source_log_public_key`?

internal/mirror/hammer/loadtest/tui.go
AlCutter: Correct - AFAIK the mirror client has been implemented to send requests as quickly as possible :)

These rate controls never controlled the client send rate, they only ever controlled the rate at which the hammer adds things to the source log; if the mirroring process pauses for a while, then the moment it resumes it'll go as fast as possible.

internal/mirror/hammer/hammer.go
AlCutter: Good spot, restructured.

internal/mirror/hammer/hammer.go
AlCutter: I've made them consistent, and also renamed the local log flags such that it's a little clearer they relate to this and not some other usage.

internal/mirror/hammer/hammer.go
```diff
@@ -175,18 +159,13 @@ func main() {
...
-		MaxWriteOpsPerSecond: *maxWriteOpsPerSecond,
-		NumReadersRandom:     0,
-		NumReadersFull:       0,
-		NumWriters:           *numWriters,
+	var hammer *loadtest.Hammer
+	var ha *loadtest.HammerAnalyser
+	if appender != nil {
+		hammer, ha = newHammer(ctx, appender, logReader, tracker)
+		ha.Run(ctx)
+		hammer.Run(ctx)
```
roger2hk: Only "Standalone Hammer Log mode" will call `hammer.Run`, so the tracker checkpoint size is never updated in the "Replication mode".

If `--leaf_write_goal` flag is provided in the replication mode, the hammer will never stop.

The TUI tracker information will not be updated in the replication mode. Is that intended? If so, we can add the logic to enforce replication mode to non-TUI.

roger2hk (approved):

internal/mirror/hammer/hammer.go
AlCutter: I've added a workaround to the TUI mode, it can probably be improved, but at least it'll show the checkpoint size now.

PR #1067, opened 2026-07-21
Allow mirrored log origins to be configured

AlCutter:
This PR updates the MTC mirror lifecycle & POSIX implementation to honour configured log origins.

This is required due to the ID and CA cosigner OID-based naming rules in the MTC spec.

Towards #945


cmd/mtc/mirror/posix/main.go
```diff
@@ -141,6 +143,17 @@ func mirrorConfigFromFlags(ctx context.Context) []witnessConfig.Log {
...
 		os.Exit(1)
 	}
+	seenOrigins := make(map[string]struct{})
+	for _, c := range cfg {
+		o := c.Origin
+		if o == "" {
+			o = c.Verifier.Name()
+		}
+		if _, seen := seenOrigins[o]; seen {
+			slog.ErrorContext(ctx, "Duplicate origin in mirror config", slog.String("origin", o))
```
roger2hk, line 153: Should this fail and exit immediately?

mirror_lifecycle.go
```diff
@@ -64,13 +64,21 @@ type MirrorOptions struct {
...
+	origin		string
 }
 
 // NewMirrorOptions creates a new options struct with defaults.
 func NewMirrorOptions() *MirrorOptions {
 	return &MirrorOptions{}
 }
 
+// WithOrigin allows the source log's origin to be specified.
+// If unsed, the name of the log verifier will be used.
```
roger2hk:
```suggestion
// If unset, the name of the log verifier will be used.
```

roger2hk (approved):

cmd/mtc/mirror/posix/main.go
AlCutter: Good spot, yes it probably should - I don't _love_ that a bad config will take the service out, but I don't think there's a safe way to easily continue here.

Updated to exit.

PR #1072, opened 2026-07-23
Move witness API path component into gateway implementation

AlCutter:
Move the witness API path detail out of the policy code and into the gateway implementation.

The witness policy handling code shouldn't care about the individual API path(s) on a witness, just its root URL.

Towards #945

### DIFF
--- a/internal/witness/client.go
+++ b/internal/witness/client.go
@@ -124,8 +124,11 @@ func NewWitnessGateway(group WitnessGroup, client *http.Client, oldSize uint64, 
 		vs := dedupVerifiers(vs)
 		if len(vs) > 0 {
 			witnesses = append(witnesses, &witnessClient{
-				client:    client,
-				url:       u,
+				client: client,
+				// Can't use path.Join here as it'll nobble the double-slash in URLs.
+				// url.JoinPath returns an error, which we can't handle here, but we already know that the URL is valid since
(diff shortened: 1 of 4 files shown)


internal/witness/client.go
```diff
@@ -124,8 +124,11 @@ func NewWitnessGateway(group WitnessGroup, client *http.Client, oldSize uint64,
 		vs := dedupVerifiers(vs)
 		if len(vs) > 0 {
 			witnesses = append(witnesses, &witnessClient{
-				client:    client,
-				url:       u,
+				client: client,
+				// Can't use path.Join here as it'll nobble the double-slash in URLs.
+				// url.JoinPath returns an error, which we can't handle here, but we already know that the URL is valid since
+				// it's coming via WitnessGroup.WitnessEndpoints(). So we'll just do the string manipulation ourselves.
+				url:       strings.TrimRight(u, "/") + "/add-checkpoint",
```
roger2hk, line 131: Could the `/` be trimmed in `WitnessGroup.WitnessEndpoints()`? This would ensure all URLs are aligned at the source.

roger2hk (approved):

internal/witness/client.go
AlCutter: Seems more correct to do it here; if it weren't for the potential to return an error, I'd have just used `url.JoinPath` here which would have done this normalisation internally.

PR #1077, opened 2026-07-24
Add WithMirrors option to append lifecycle

AlCutter:
This PR adds the ability to configure mirroring in the append lifecycle.

Since mirrors are witnesses with special powers, we use the same mechanism as `WithWitnesses` to describe the policy.

The implementation of the client/gateway code will come later.

Towards #945

### DIFF
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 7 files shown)


append_lifecycle.go
```diff
@@ -735,12 +739,31 @@ func (o AppendOptions) CheckpointPublisher(lr LogReader, httpClient *http.Client
 			span.AddEvent("Created CP")
 			appenderSignedSize.Record(ctx, otel.Clamp64(size))
 
-			wSigs, err := witnessCheckpoint(ctx, cp, size, o.witnesses, lr, httpClient, o.witnessOpts)
-			if err != nil {
-				return nil, err
+			eg := errgroup.Group{}
+			sigC := make(chan []byte, 2)
```
roger2hk: nit: Using channel seems to be an overkill here. It can be replaced by:

```go
var ws, ms []byte
...
...
...
cp = append(cp, ws...)
cp = append(cp, ms...)
```

append_lifecycle.go
```diff
@@ -950,6 +985,27 @@ func (o *AppendOptions) WithWitnesses(witnesses WitnessGroup, opts *WitnessOptio
...
 }
 
+// WithMirrors configures the set of tlog-mirror servers that Tessera will contact in order to obtain
+// mirror counter-signatures on a checkpoint before publishing it.
+//
+// Requests will be sent to every mirror referenced by the group using the tlog-mirror API at the configured URL.
+// The checkpoint will be accepted for publishing when a sufficient number of mirrors to satisfy the group
+// have responded.
+//
+// If this method is not called, then no mirror counter signatures will be required to publish.
```
roger2hk:
```suggestion
// WithMirrors configures the set of tlog-mirror servers that Tessera will contact in order to obtain
// mirror countersignatures on a checkpoint before publishing it.
//
// Requests will be sent to every mirror referenced by the group using the tlog-mirror API at the configured URL.
// The checkpoint will be accepted for publishing when a sufficient number of mirrors to satisfy the group
// have responded.
//
// If this method is not called, then no mirror countersignatures will be required to publish.
```

Update to `countersignatures` for spelling consistency (`counter signatures`, `counter-signatures`).

roger2hk (approved):

append_lifecycle.go
AlCutter: Done, but meh :)
The assignments are a little more prone to future bugs (e.g. cut-n-paste assigning to the wrong var), but I don't feel very strongly about it.

PR #1080, opened 2026-07-29
Implement MTCLog AddTBS HTTP handler

phbnf:
Towards #945.

This PR also make `MTCLogEntry` fields private, and introduces a `New` function to create these objects.
We could, in a followup PR move the extension checks to it.. they are not used at the moment, so this this is not super important.

### DIFF
--- a/cmd/mtc/log/mtc.go
+++ b/cmd/mtc/log/mtc.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"github.com/transparency-dev/tessera"
+	"github.com/transparency-dev/tessera/cmd/mtc/log/internal/entry"
 	"golang.org/x/crypto/cryptobyte"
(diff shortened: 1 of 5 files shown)


cmd/mtc/log/mtc.go
```diff
@@ -33,6 +34,13 @@ type MTCLog struct {
 // draft-ietf-plants-merkle-tree-certs section 6.2.
 type MTCProof struct{}
 
+// AddTBSRsp contains enough information from the log
+// to build a standalone certificate.
+type AddTBSRsp struct {
+	Index    uint64   `json:"index"`
+	MTCProof MTCProof `json:"mtcproof"`
```
AlCutter: Maybe `mtcProof` to match camel-case style on the request body JSON?

AlCutter (approved):

PR #1081, opened 2026-07-29
Implement MVP mirror support in append lifecycle

AlCutter:
This PR adds a simple MVP implementation of mirroring support in the append lifecycle.

This approach is almost certainly going to change, but this will let us make progress in the meantime.

Towards #945

### DIFF
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -15,6 +15,7 @@
 package tessera
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -30,6 +31,7 @@ import (
 	f_log "github.com/transparency-dev/formats/log"
 	"github.com/transparency-dev/merkle/rfc6962"
 	"github.com/transparency-dev/tessera/api/layout"
+	m_gateway "github.com/transparency-dev/tessera/internal/mirror/gateway"
 	"github.com/transparency-dev/tessera/internal/otel"
 	"github.com/transparency-dev/tessera/internal/parse"
 	"github.com/transparency-dev/tessera/internal/witness"
@@ -745,7 +747,22 @@ func (o *AppendOptions) WithAntispam(inMemEntries uint, as Antispam) *AppendOpti
 }
 
 // CheckpointPublisher returns a function which should be used to create, sign, and potentially witness a new checkpoint.
+// Deprecated: Use CheckpointPublisherContex.
 func (o AppendOptions) CheckpointPublisher(lr LogReader, httpClient *http.Client) func(context.Context, uint64, []byte) ([]byte, error) {
+	return func(ctx context.Context, size uint64, root []byte) ([]byte, error) {
+		return o.CheckpointPublisherContext(ctx, lr, httpClient)(ctx, size, root)
+	}
+}
+
+// CheckpointPublisherContext returns a function which should be used to create, sign, and potentially witness a new checkpoint.
+func (o AppendOptions) CheckpointPublisherContext(ctx context.Context, lr LogReader, httpClient *http.Client) func(context.Context, uint64, []byte) ([]byte, error) {
+	gw := sync.OnceValue(func() *m_gateway.Gateway {
+		if len(o.mirrors.Components) > 0 {
+			return m_gateway.NewGateway(ctx, httpClient, o.mirrors, lr, o.primarySigner.Name())
+		}
+		return nil
+	})
+
 	return func(ctx context.Context, size uint64, root []byte) ([]byte, error) {
 		return otel.Trace(ctx, "tessera.CheckpointPublisher", tracer, func(ctx context.Context, span trace.Span) ([]byte, error) {
 			cp, err := o.newCP(ctx, size, root)
@@ -763,8 +780,13 @@ func (o AppendOptions) CheckpointPublisher(lr LogReader, httpClient *http.Client
 				return err
 			})
 			eg.Go(func() error {
+				if gw == nil {
+					return nil
+				}
+				mirrorCtx, cancel := context.WithTimeout(ctx, o.mirrorOpts.Timeout)
+				defer cancel()
 				var err error
-				ms, err = mirrorCheckpoint(ctx, cp, size, o.mirrors, lr, httpClient, o.mirrorOpts)
+				ms, err = mirrorCheckpoint(mirrorCtx, gw, &o.mirrors, cp, size, o.mirrorOpts.FailOpen)
 				return err
 			})
 
@@ -827,15 +849,38 @@ func witnessCheckpoint(ctx context.Context, cp []byte, cpSize uint64, witnesses 
 	})
 }
 
-// mirrorCheckpoint takes care of mirroring the given checkpoint with the provided mirror policy.
-// Returns signatures from mirrors, ready to append to the checkpoint, or an error.
-func mirrorCheckpoint(ctx context.Context, cp []byte, cpSize uint64, mirrors WitnessGroup, lr LogReader, httpClient *http.Client, opts MirroringOptions) ([]byte, error) {
-	return otel.Trace(ctx, "tessera.mirrorCheckpoint", tracer, func(ctx context.Context, span trace.Span) ([]byte, error) {
-		if len(mirrors.Components) == 0 {
-			return nil, nil
+func mirrorCheckpoint(ctx context.Context, gw *m_gateway.Gateway, policy *WitnessGroup, cp []byte, size uint64, failOpen bool) ([]byte, error) {
+	// TODO(al): Add metrics
+	checkPolicy := func(sigs []byte) ([]byte, error) {
+		newCP := append(slices.Clone(cp), sigs...)
+		if policy.Satisfied(newCP) {
+			return sigs, nil
 		}
-		span.AddEvent("Starting mirroring")
-		return nil, nil
+		if failOpen {
+			slog.WarnContext(ctx, "MirrorGateway: policy not met, failing-open")
+			return sigs, nil
+		}
+		return sigs, fmt.Errorf("MirrorGateway: policy not met")
+	}
+
+	return otel.Trace(ctx, "tessera.mirrorCheckpoint", tracer, func(ctx context.Context, span trace.Span) ([]byte, error) {
+		sigCh := gw.CosignCheckpoint(ctx, cp, size)
+		var sigBlock bytes.Buffer
+		for {
+			select {
+			case <-ctx.Done():
+				return checkPolicy(sigBlock.Bytes())
+			case sig, ok := <-sigCh:
+				if !ok {
+					return checkPolicy(sigBlock.Bytes())
+				}
+				sigBlock.Write(sig)
+				newCP := append(slices.Clone(cp), sigBlock.Bytes()...)
+				if policy.Satisfied(newCP) {
+					return sigBlock.Bytes(), nil
+				}
+			}
+		}
 	})
 }
 
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -19,7 +19,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"net/http"
+	"net/url"
 	"os"
 	"slices"
 	"sync"
@@ -746,8 +748,21 @@ func (o *AppendOptions) WithAntispam(inMemEntries uint, as Antispam) *AppendOpti
 	return o
 }
 
+// parseURLs converts a list of URL strings to a list of *url.URL, failing if any cannot be parsed.
+func parseURLs(us []string) ([]*url.URL, error) {
+	ret := make([]*url.URL, 0, len(us))
+	for _, s := range us {
+		u, err := url.Parse(s)
+		if err != nil {
+			return nil, err
+		}
+		ret = append(ret, u)
+	}
+	return ret, nil
+}
+
 // CheckpointPublisher returns a function which should be used to create, sign, and potentially witness a new checkpoint.
-// Deprecated: Use CheckpointPublisherContex.
+// Deprecated: Use CheckpointPublisherContext.
 func (o AppendOptions) CheckpointPublisher(lr LogReader, httpClient *http.Client) func(context.Context, uint64, []byte) ([]byte, error) {
 	return func(ctx context.Context, size uint64, root []byte) ([]byte, error) {
 		return o.CheckpointPublisherContext(ctx, lr, httpClient)(ctx, size, root)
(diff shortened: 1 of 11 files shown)


append_lifecycle.go
```diff
@@ -827,15 +883,38 @@ func witnessCheckpoint(ctx context.Context, cp []byte, cpSize uint64, witnesses
 	})
 }
 
-// mirrorCheckpoint takes care of mirroring the given checkpoint with the provided mirror policy.
-// Returns signatures from mirrors, ready to append to the checkpoint, or an error.
-func mirrorCheckpoint(ctx context.Context, cp []byte, cpSize uint64, mirrors WitnessGroup, lr LogReader, httpClient *http.Client, opts MirroringOptions) ([]byte, error) {
+func mirrorCheckpoint(ctx context.Context, gw *m_gateway.Gateway, policy *WitnessGroup, cp []byte, size uint64, failOpen bool) ([]byte, error) {
```
phbnf:
```suggestion
func mirrorCheckpoint(ctx context.Context, gw *m_gateway.Gateway, policy *WitnessGroup, cp []byte, size uint64, failOpen bool) ([]byte, error) {
	// TODO(al): Add metrics
	checkPolicy := func(sigs []byte) bool {
		newCP := append(slices.Clone(cp), sigs...)
		return policy.Satisfied(newCP)
	}
	exit  := func(sigs []byte) (][byte, error) {
	    newCP, ok := checkPolicy(sigs)
	    if ok {
	      return sigs, nil
	    }
		if failOpen {
			slog.WarnContext(ctx, "MirrorGateway: policy not met, failing-open")
			return sigs, nil
		}
		return sigs, fmt.Errorf("MirrorGateway: policy not met")
	}

return otel.Trace(ctx, "tessera.mirrorCheckpoint", tracer, func(ctx context.Context, span trace.Span) ([]byte, error) {
		sigCh := gw.CosignCheckpoint(ctx, cp, size)
		var sigBlock bytes.Buffer
		for {
			select {
			case <-ctx.Done():
				return exit(sigBlock.Bytes())
			case sig, ok := <-sigCh:
				if !ok {
					return exit(sigBlock.Bytes())
				}
				sigBlock.Write(sig)
                return checkPolicy(sig)
			}
		}
	})
}
```

phbnf: Very nit, but I think you can do something along these lines. (but don't apply the suggestion, you'll get into CLA issues)

append_lifecycle.go
```diff
@@ -744,8 +748,55 @@ func (o *AppendOptions) WithAntispam(inMemEntries uint, as Antispam) *AppendOpti
...
+			f, err = o.CheckpointPublisherContext(ctx, lr, httpClient)
+		})
+		if err != nil {
+			return nil, err
+		}
+		return f(ctx, size, root)
+	}
+}
+
+// CheckpointPublisherContext returns a function which should be used to create, sign, and potentially witness a new checkpoint.
```
phbnf: `potentially witness and/or mirror a new checkpoint`

append_lifecycle.go
```diff
@@ -763,8 +814,13 @@ func (o AppendOptions) CheckpointPublisher(lr LogReader, httpClient *http.Client
...
 			})
 			eg.Go(func() error {
+				if gw == nil {
+					return nil
+				}
+				mirrorCtx, cancel := context.WithTimeout(ctx, o.mirrorOpts.Timeout)
+				defer cancel()
 				var err error
-				ms, err = mirrorCheckpoint(ctx, cp, size, o.mirrors, lr, httpClient, o.mirrorOpts)
+				ms, err = mirrorCheckpoint(mirrorCtx, gw, &o.mirrors, cp, size, o.mirrorOpts.FailOpen)
```
phbnf: nit: you could move `cp, size` before `&o.mirrors` to match with `witnessCheckpoint`

append_lifecycle.go
```diff
@@ -744,8 +748,55 @@ func (o *AppendOptions) WithAntispam(inMemEntries uint, as Antispam) *AppendOpti
...
+	}
+}
+
+// CheckpointPublisherContext returns a function which should be used to create, sign, and potentially witness a new checkpoint.
+func (o AppendOptions) CheckpointPublisherContext(ctx context.Context, lr LogReader, httpClient *http.Client) (func(context.Context, uint64, []byte) ([]byte, error), error) {
+	mirrorURLs, err := parseURLs(slices.Collect(maps.Keys(o.mirrors.WitnessEndpoints())))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse mirror URLs: %w", err)
+	}
+	gw, err := m_gateway.NewGateway(ctx, m_gateway.Options{
```
phbnf: Should this be skipped is `len(mirrorURLs)==0`? I don't think it's required, but it could help readability, and avoid "ghost" metrics further down the line like it was was the case with witnessing. You could even check the length of WitnessEndpoint and only parse URLs if need be.

internal/mirror/gateway/mirror_gateway.go
```diff
@@ -0,0 +1,256 @@
...
+
+		mOpts := mirror.NewOptions().
+			WithMirrorURL(u).
+			WithHTTPClient(opts.HTTPClient).
+			WithLogOrigin(opts.LogOrigin).
+			WithTileFetcher(opts.LogReader.ReadTile).
+			WithBundleFetcher(opts.LogReader.ReadEntryBundle).
+			WithMirrorCheckpointFetcher(mirrorFetcher.ReadCheckpoint)
+
+		c, err := mirror.NewClient(ctx, mOpts)
```
phbnf, line 111: Can the context be removed from `NewClient`?

internal/mirror/gateway/mirror_gateway.go
```diff
@@ -0,0 +1,256 @@
...
+		case <-ctx.Done():
+			return
+		case job, ok := <-target.goals:
+			if !ok {
+				// Channel closed: down tools and exit.
+				return
+			}
+
+			// TODO(al): Consider making this timeout configurable. This should be fine for the moment though.
+			goalJob := g.chaseGoal(target, job, 1*time.Minute)
```
phbnf: Naming is hard: we've got `g` for Gateway (not for goal), `goal` (a struct, not a job), `target` (a mirror endpoint, not goal target). `g.chaseGoal(target, job, 1*time.Minute)` (which doesn't chase the goal but returns a function that will let the gateway chase this goal).

What about:

  - `g` --> `gw`
  - `job` --> `goal`
  - `chaseGoal` --> `chaseGoalJob` / `goalChaser` / `goalScorer`

internal/mirror/gateway/mirror_gateway.go
```diff
@@ -0,0 +1,256 @@
...
+			done: func(sig []byte, err error) {
+				// Last one out please turn off the lights.
+				defer func() {
+					if N.Add(1) == uint32(len(g.targets)) {
+						close(out)
+					}
+				}()
+
+				if err != nil {
+					slog.ErrorContext(ctx, "MirrorGateway: Sync failed", slog.String("url", target.url.String()), slog.Any("error", err))
```
phbnf, line 163: Should this be an error if we're replacing a goal, which hasn't even been picked up by any worker yet?

internal/mirror/gateway/mirror_gateway.go
```diff
@@ -0,0 +1,256 @@
...
+			case <-ctx.Done():
+				done = true
+			case target.goals <- newGoal:
+				done = true
+			default:
+				// No space in the goals channel, try to supercede the goal currently in there with the new one.
+				// This replacement is "racy", but the worst that can happen is that the worker has already
+				// picked up the old goal and we end up simply queuing the new goal instead of replacing the old one.
+				select {
+				case oldGoal := <-target.goals:
```
phbnf, line 183: Could this behavior get a slow mirror in a state where it never catches up? Should `target.goals` have a bit of a buffer to have piecemeal uploads?

internal/mirror/gateway/mirror_gateway.go
```diff
@@ -0,0 +1,256 @@
...
+
+		target := &mirrorTarget{
+			url:    u,
+			client: c,
+			goals:  make(chan goal, 1),
+		}
+		g.targets = append(g.targets, target)
+
+		// Start the worker goroutine.
+		go g.runWorker(ctx, target)
```
phbnf: Should this happen in a separate loop? Otherwise, if for some reason `client.NewHTTPFetcher` fails, then workers that have already been started will stay around.

internal/mirror/gateway/mirror_gateway.go
```diff
@@ -0,0 +1,256 @@
...
+
+			// In a func for context defer.
+			func() {
+				cctx, cancel := context.WithTimeout(ctx, timeout)
+				defer cancel()
+
+				slog.DebugContext(cctx, "MirrorGateway: Syncing mirror", slog.String("url", target.url.String()), slog.Uint64("goal", job.cpSize))
+				rSigs, rErr = target.client.Sync(cctx, job.cp, job.cpSize)
+				if rErr != nil {
+					// TODO(al): Update the client so we can tell whether an error is permanent or transient, and abandon jobs which will never succeed.
```
phbnf, line 247: In the meantime, should `goal` include `cosignCheckpoint`'s `ctx` such that `chaseGoal` can return if this context is being cancelled?

append_lifecycle.go
```diff
@@ -744,8 +748,55 @@ func (o *AppendOptions) WithAntispam(inMemEntries uint, as Antispam) *AppendOpti
...
+		if err != nil {
+			return nil, err
+		}
+		ret = append(ret, u)
+	}
+	return ret, nil
+}
+
 // CheckpointPublisher returns a function which should be used to create, sign, and potentially witness a new checkpoint.
+// Deprecated: Use CheckpointPublisherContext.
```
phbnf, line 765: Very nit: should validate throw a warning message, or at least prevent `CheckpointPublisher`, and `CheckpointPublisherContext` from both being called?

append_lifecycle.go
AlCutter: Good call - I've done something slightly different but inspired by this comment, and  I think it's a lot easier to see that the policy is checked everywhere it should be. Ta!

append_lifecycle.go
AlCutter: Yeah, not a bad call - makes the publisher func a bit smaller too. Done.

internal/mirror/gateway/mirror_gateway.go
AlCutter: Potentially in another CL, but it may not be a bad idea to keep it around to avoid changing API?

internal/mirror/gateway/mirror_gateway.go
AlCutter: I think it's ok; it's almost certainly the case that the only way of getting goals into this queue is via `publishCheckpoint` in the same task, so that means that any pending goal in the queue which we're about to overwrite came from the previous `publishCheckpoint` which _must_ have already given up and exited since another one is now mid-run.

If this even happens, then it can only be the case that either the mirror is _way_ behind and needs catching up (in which case it's just not in a position to return cosigs for recent checkpoint until it's caught up), or the mirror service itself is down.

internal/mirror/gateway/mirror_gateway.go
AlCutter: All future goals "include" all the work expected from earlier goals, so the mirror should catch up as quickly as it can, which could mean "never" if it's super janky and simply can't keep up - in that case this worker will just keep trying to bring it along as best it can.

internal/mirror/gateway/mirror_gateway.go
AlCutter: Great spot - yes, done!

internal/mirror/gateway/mirror_gateway.go
AlCutter: No, at the moment I do want it to keep going in case it's able to catch the mirror up (this is why there are long-running workers taking goals over channels rather than workers fanning out and back in within the scope of `cosignCheckpoint`.

append_lifecycle.go
AlCutter: I think it's super unlikely to happen - this API is _technically_ public, but the only place it makes sense to call it from is the internals of Tessera.

internal/mirror/gateway/mirror_gateway.go
phbnf: From https://c2sp.org/tlog-mirror#updating-a-mirror: ` The mirror MAY handle the request by internally updating the pending checkpoint and responding with an empty response body.`

So depending on how the mirror is implemented, it might never pushing a new checkpoint corresponding to the previous pending checkpoint. But I guess mirrors should just handle that properly?

append_lifecycle.go
phbnf: Ok, then maybe change the description of that method to remove "or mirror"? I'm not sure that mirroring would work with this method, since the mirror workers would be created using the initial context passed to `publishCheckpoint`, which will eventually be cancelled.

phbnf (approved):

append_lifecycle.go
AlCutter: On further reflection, I've changed the old method to just `panic` - it's really an internal API, and trying to patch it together is just more likely to cause us weird hard-to-track-down behaviour in the future (like the one you've pointed to here). Better to just make it go bang if we've accidentally used the wrong method.

internal/mirror/gateway/mirror_gateway.go
AlCutter: I'm having trouble parsing the second sentence, but _"depending how the client is implemented, it might never push the entries corresponding to a previous pending checkpoint"_ is definitely possible, and the client and mirror should be fine handling that. E.g. the client may push a newer checkpoint to add-checkpoint and then all the entries under that to `add-checkpoint` and everything should be fine.

phbnf: Indeed, sorry I meant: `So depending on how the mirror is implemented, it might never <publish> a new checkpoint corresponding to the previous pending checkpoint.`

Anyway, not blocking for this PR.

phbnf (approved):

internal/mirror/gateway/mirror_gateway.go
AlCutter: Oh, totally - there's absolutely no guarantee that the client which has previously pushed a checkpoint to `add-checkpoint` will ever come back and send any requests _at all_ to `add-entries` (e.g the client == log, and it's in the middle of publishing a new checkpoint, sends `add-checkpoint` then promptly crashes. At best losing only the previously "in-flight" checkpoint, at worst the whole log goes up in smoke).

PR #1090, opened 2026-08-04
Add WithOrigin option to AppendOptions and wire it up for the MTC Log

phbnf:
Towards #945.

### DIFF
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -893,7 +893,6 @@ func (o *AppendOptions) WithCheckpointSigner(s note.Signer, additionalSigners ..
 	if o.origin == "" {
 		o.origin = s.Name()
 	}
-	origin := o.origin
 	o.primarySigner = s
 	o.additionalSigners = additionalSigners
 	o.newCP = func(ctx context.Context, size uint64, hash []byte) ([]byte, error) {
@@ -905,7 +904,7 @@ func (o *AppendOptions) WithCheckpointSigner(s note.Signer, additionalSigners ..
 				hash = emptyRoot[:]
 			}
 			cpRaw := f_log.Checkpoint{
-				Origin: origin,
+				Origin: o.origin,
 				Size:   size,
 				Hash:   hash,
 			}.Marshal()
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -714,7 +714,10 @@ func (o AppendOptions) valid() error {
 
 // WithOrigin configures the log's origin.
 //
-// If unset, WithCheckpointSigner() configures the origin with the checkpoint signer's name.
+// Most logs SHOULD NOT set this option, but instead rely on the default
+// Tessera behaviour of using the signer key's name as the log origin in
+// WithCheckpointSinger().
+//
 // SPEC: https://c2sp.org/tlog-checkpoint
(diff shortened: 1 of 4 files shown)


append_lifecycle.go
```diff
@@ -873,8 +885,15 @@ func (o AppendOptions) ShutdownTimeout() time.Duration {
...
+//
+// If no origin was specified using WithOrigin(), configures the origin with the signer's name.
+// SPEC: https://c2sp.org/tlog-checkpoint
+// "The log’s key name in its signature line SHOULD match the origin line."
 func (o *AppendOptions) WithCheckpointSigner(s note.Signer, additionalSigners ...note.Signer) *AppendOptions {
-	origin := s.Name()
+	if o.origin == "" {
+		o.origin = s.Name()
+	}
+	origin := o.origin
```
AlCutter: There's an unfortunate thing here that `WithOrigin` will just silently fail if it's set _after_ setting `WithCheckpointSigner`.

I think you can just remove `origin := o.origin` above, and have the `newCP` func just reference `o.origin` directly to fix it?

The alternative would be to just have all these options store their params in `o` and then have a `resolveOptions()` func which does all the processing after all params have been set. That seems like a bigger change though.

cmd/mtc/log/posix/main.go
```diff
@@ -39,6 +39,7 @@ var (
 	writeTimeout      = flag.Duration("write_timeout", 0, "The maximum duration before timing out writes of the response.")
 
 	storageDir                = flag.String("storage_dir", "", "Path to root of log storage.")
+	origin                    = flag.String("origin", "", "The log origin name.")
```
AlCutter: I wonder if this should reference `Log ID` either in the name of the flag or in the description to tie it to the MTC spce?

AlCutter (approved):

cmd/mtc/log/posix/main.go
phbnf: I've gone with `logOrigin` and `log_origin` to stick with the specs: https://ietf-plants-wg.github.io/merkle-tree-certs/draft-ietf-plants-merkle-tree-certs.html#section-5.3.1.

I took this as an opportunity to fix the TODO, and actually check the origin an signer format now.

We could take `log_number` flag instead of taking `log_origin`, but I'm a bit uncomfortable with this for now. I think it is quite nice for the log origin to be explicit rather than dynamically computed since it might be reused in other places.

append_lifecycle.go
```diff
@@ -710,6 +712,16 @@ func (o AppendOptions) valid() error {
 	return nil
 }
 
+// WithOrigin configures the log's origin.
+//
+// If unset, WithCheckpointSigner() configures the origin with the checkpoint signer's name.
+// SPEC: https://c2sp.org/tlog-checkpoint
+// "The log’s key name in its signature line SHOULD match the origin line."
```
AlCutter: Could you bring this a little more "forward"? E.g. something like:

> Most logs SHOULD NOT set this option, but instead rely on the default Tessera behaviour of using the signer key's name as the log origin.

AlCutter (approved):
Love the constructing of the origin/signer name in the latest rev, nice!

PR #1091, opened 2026-08-05
Update formats

AlCutter:
Bump `formats` dep.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	github.com/hashicorp/golang-lru/v2 v2.0.7
 	github.com/muesli/termenv v0.16.0
 	github.com/rivo/tview v0.42.0
-	github.com/transparency-dev/formats v0.1.2-0.20260710124811-af9e607161b6
+	github.com/transparency-dev/formats v0.1.2-0.20260805102052-38e6e69c4152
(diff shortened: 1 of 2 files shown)


cmd/conformance/aws/main.go
```diff
@@ -261,16 +263,16 @@ func antispamMysqlConfig() *mysql.Config {
 	}
 }
 
-func signerFromFlags() (note.Signer, []note.Signer) {
+func signerFromFlags() (note.Signer, []sdbnote.Signer) {
```
phbnf: What's the benefit of returning `note.Singer, []sdbnote.Signer` over `sdbnote.Signer, []sdbnote.Signer`?

`WithCheckpoint` only takes `sdbnote.Signers` as arguments, so we might as well keep things consistent? (and also use a named import in append_lifecycle.go`.

phbnf (approved):

cmd/conformance/aws/main.go
AlCutter: I was trying to move towards standardising on our own `formats/note` package where possible, but this seems like a rabbit hole.

Reverted everything but the `go.{mod,sum}` files.

PR #1097, opened 2026-08-07
[MTC] Add landmark package with Landmarks struct and text marshaling

phbnf:
Towards #945.

### DIFF
--- a/cmd/mtc/log/internal/landmark/landmark.go
+++ b/cmd/mtc/log/internal/landmark/landmark.go
@@ -22,11 +22,12 @@ import (
 	"strings"
 )
 
-// Landmarks represents the active landmarks published by a CA log.
+// ActiveLandmarks represents the active landmarks published by a CA log.
 // as described in draft-ietf-plants-merkle-tree-certs section 6.4.3.
 //
-// Landmarks is not safe for concurrent use without external synchronization.
-type Landmarks struct {
+// ActiveLandmarks is not safe for concurrent use without external synchronization.
+type ActiveLandmarks struct {
+	// The number of non-zero landmark tree sizes the log has ever published.
 	lastLandmark       uint64
 	numActiveLandmarks uint64
 	// treeSizes contains numActiveLandmarks + 1 tree sizes in strictly decreasing order.
@@ -35,10 +36,10 @@ type Landmarks struct {
 	treeSizes []uint64
 }
 
-// newLandmarks creates a new Landmarks struct with the given parameters.
+// newActiveLandmarks creates a new ActiveLandmarks struct with the given parameters.
 // Parameters must represent a valid landmarks file as per specs, otherwise
 // returns an error. Specifically, treeSizes MUST have at least one entry.
-func newLandmarks(lastLandmark, numActive uint64, treeSizes []uint64) (*Landmarks, error) {
+func newActiveLandmarks(lastLandmark, numActive uint64, treeSizes []uint64) (*ActiveLandmarks, error) {
 	// SPEC: draft-ietf-plants-merkle-tree-certs section 6.4.3.
 	// "num_active_landmarks <= last_landmark"
 	if numActive > lastLandmark {
@@ -66,11 +67,10 @@ func newLandmarks(lastLandmark, numActive uint64, treeSizes []uint64) (*Landmark
 			return nil, fmt.Errorf("tree sizes must be strictly decreasing: treeSizes[%d]=%d <= treeSizes[%d]=%d", i-1, treeSizes[i-1], i, treeSizes[i])
 		}
 	}
-	tsCopy := slices.Clone(treeSizes)
-	return &Landmarks{
+	return &ActiveLandmarks{
 		lastLandmark:       lastLandmark,
 		numActiveLandmarks: numActive,
-		treeSizes:          tsCopy,
+		treeSizes:          slices.Clone(treeSizes),
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -0,0 +1,172 @@
...
+	// treeSizes contains numActiveLandmarks + 1 tree sizes in strictly decreasing order.
+	// It must never be empty.
+	// treeSizes[i] corresponds to landmark `lastLandmark - i`.
+	treeSizes []uint64
+}
+
+// New creates a new Landmarks struct with the given parameters.
+// Parameters must represent a valid landmarks file as per specs, otherwise
+// returns an error. Specifically, treeSizes MUST have at least one entry.
+func New(lastLandmark, numActive uint64, treeSizes []uint64) (*Landmarks, error) {
```
AlCutter: I wonder if this c'tor should be unexported?

IIUC, the main flow for using this structure is something like:
1. Unmarshall from the stored file
2. Allocate a new land mark
3. Marshall & overwrite the stored file

The only special case is when there is no file yet, but perhaps that's consistent with the zero-value of a `&Landmarks{}` struct?

cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -0,0 +1,172 @@
...
+// as described in draft-ietf-plants-merkle-tree-certs section 6.4.3.
+//
+// Landmarks is not safe for concurrent use without external synchronization.
+type Landmarks struct {
+	lastLandmark       uint64
+	numActiveLandmarks uint64
+	// treeSizes contains numActiveLandmarks + 1 tree sizes in strictly decreasing order.
+	// It must never be empty.
+	// treeSizes[i] corresponds to landmark `lastLandmark - i`.
+	treeSizes []uint64
```
AlCutter, line 36: This is going to grow without bounds, right?

AlCutter (commented):
Just a thought - do we actually need to _load in_ all of the data from the (ever-growing?) landmarks file, or could we just _read_ it but only track derived/"active" state in this struct, and when writing out just append new landmarks? (Obvs storage impl will need to manage that appending appropriately for the infra, but that's true generally)

cmd/mtc/log/internal/landmark/landmark.go
phbnf: Good point, we don't need to export it. It will make some tests a tiny bit harder to write, but we can just use `UnmarshalText` instead. It's in internal anyways, so it's not too important.

cmd/mtc/log/internal/landmark/landmark.go
phbnf: It will not, `AddLandmark` caps its length `maxActive+1`.

I tried a few things, and tried to make `maxActive` a field of `Landmarks`, but it didn't work very well because `UnmarshalText` isn't really aware of `maxActive`.

The next PR should clarify how it works, it introduces a `LandmarkManager`, which computes `maxActive` based on the update frequency and the maximum certificate lifetime. `LandmarkManager` will then call `Update(treeSize, maxActive)`.

AlCutter: Ah, so maybe this struct might be more accurately called `ActiveLandmarks`?

cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -0,0 +1,172 @@
...
+	"strconv"
+	"strings"
+)
+
+// Landmarks represents the active landmarks published by a CA log.
+// as described in draft-ietf-plants-merkle-tree-certs section 6.4.3.
+//
+// Landmarks is not safe for concurrent use without external synchronization.
+type Landmarks struct {
+	lastLandmark       uint64
```
AlCutter, line 31: I think this just a counter of the number of landmarks ever issued by the log, is that right? (worth a comment on this either way, took a while to figure that out from the spec :))

cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	// SPEC: draft-ietf-plants-merkle-tree-certs section 6.4.1.
+	// "The sequence of tree sizes MUST be append-only and strictly monotonically
+	// increasing."
+	for i := 1; i < len(treeSizes); i++ {
+		if treeSizes[i-1] <= treeSizes[i] {
+			return nil, fmt.Errorf("tree sizes must be strictly decreasing: treeSizes[%d]=%d <= treeSizes[%d]=%d", i-1, treeSizes[i-1], i, treeSizes[i])
+		}
+	}
+	tsCopy := slices.Clone(treeSizes)
```
AlCutter: Can just inline this on L73

cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -0,0 +1,172 @@
...
+	if len(text) == 0 || text[len(text)-1] != '\n' {
+		return errors.New("landmarks text must not be empty and must end with a newline")
+	}
+
+	lines := strings.Split(strings.TrimSuffix(string(text), "\n"), "\n")
+	if len(lines) < 2 {
+		return fmt.Errorf("expected at least header line and one tree size line, got %d lines", len(lines))
+	}
+
+	headerParts := strings.Split(lines[0], " ")
```
AlCutter: Could do something this to simplify the continued handling later on:

```go
headerParts, lines := strings.Split(lines[0], " "), lines[1:]
```

AlCutter (approved):

PR #1099, opened 2026-08-07
Refactor witness gateway

AlCutter:
This PR makes the witness gateway behave more like the mirror gateway: the gateway is concerned only with fetching cosigs from witnesses, and policy decisions are now made at the layer above.

This makes both types of "cosigner gateway" look similar, and so we can now reuse common code to process signatures and apply policy.

The custom witness client implementation in the repo has been removed, and we now use the client from the `transparency-dev/witness` repo instead.

Towards #945


internal/witness/client.go
```diff
@@ -90,170 +87,152 @@ func init() {
...
-	}
+		fetchTile: opts.FetchTiles,
+	}, nil
 }
 
-// dedupVerifiers removes duplicate verifiers (identified by name and keyhash) from the given slice.
-// Returns a new slice containing the unique verifiers, the passed in slice isn't modified.
-func dedupVerifiers(vs []note.Verifier) []note.Verifier {
+// dedup merges witnesses with the same URL, deduplicating verifiers within each witness.
+func dedup(ws []Witness) []Witness {
```
phbnf: What if (?!) two witnesses share the same key behind two different urls?

append_lifecycle.go
```diff
@@ -834,77 +867,86 @@ func (o AppendOptions) mirrorGateway(ctx context.Context, lr LogReader) (*m_gate
...
-	// TODO(al): Add metrics
-	checkPolicy := func(sigs []byte, failOpen bool) ([]byte, error) {
-		newCP := append(slices.Clone(cp), sigs...)
-		if !policy.Satisfied(newCP) {
-			if failOpen {
-				slog.WarnContext(ctx, "MirrorGateway: policy not met, failing-open")
-				return sigs, nil
+// mirrorCheckpoint takes care of mirroring the given checkpoint with the provided mirror policy.
+// Returns signatures from mirrors, ready to append to the checkpoint, or an error.
+func mirrorCheckpoint(ctx context.Context, mirror cosigSource, policy *WitnessGroup, cp []byte, cpSize uint64, opts MirroringOptions) ([]byte, error) {
```
phbnf: nit, rename `mirror` to `sigsFetcher` or `mirroredSigns`: anything that could convey that it carries signatures, and they originate from multiple mirrors, not a single one.

internal/witness/client.go
```diff
@@ -311,137 +288,52 @@ func names(w []note.Verifier) []string {
...
+				return nil, backoff.Permanent(err)
+			}
+			d := time.Since(start)
+			statusAttr := witnessStatusKey.String("success")
+			witnessClientRespsTotal.Add(ctx, 1, metric.WithAttributes(nameAttr, statusAttr))
+			witnessClientReqHistogram.Record(ctx, d.Milliseconds(), metric.WithAttributes(nameAttr, statusAttr))
+			w.oldSize = cpSize
+			return sigs, nil
+		},
+			backoff.WithMaxTries(maxUpdateRetries),
```
phbnf, line 358: For readability, maybe move these above close to `backoff.Retry` or un-inline `fn`?

internal/witness/client.go
```diff
@@ -311,137 +288,52 @@ func names(w []note.Verifier) []string {
...
-		b.Write(dst)
-		b.WriteByte('\n')
-	}
-	b.WriteByte('\n')
-	b.Write(cp)
-	return &b
+				if errors.Is(err, witness.ErrCheckpointStale) {
+					w.oldSize = actualSize
+					// Don't mark as permanent so that the backoff will retry.
+					return nil, err
```
phbnf, line 340: What's the rationale for using exponential backoff here? If I understand correctly, the only error that will actually let a retry through is `ErrCheckpointStale`, and in that scenario I think it is okay to retry immediately?

phbnf: Or maybe, re-introduce the comment below that explains that this could happen forever, and that this is what the exponential backoff is about!

internal/witness/client.go
```diff
@@ -311,137 +288,52 @@ func names(w []note.Verifier) []string {
...
+			var (
+				proof [][]byte
+				err   error
+			)
+			if w.oldSize > 0 {
+				proof, err = fetchProof(ctx, w.oldSize, cpSize)
 				if err != nil {
-					return nil, fmt.Errorf("witness at %q replied with x.tlog.size but body %q could not be parsed as decimal", w.url, bodyStr)
-				}
-				// This should _never_ happen - the witness has a larger tree size than the log knows about!
```
phbnf, line 392: I thin that check is now gone.

internal/witness/client.go
AlCutter: This should never happen, but I think it'd probably also be wrong for us to _assume_ that these are the _same_ witness despite what the config file says.

internal/witness/client.go
AlCutter: I've popped in a few more comments which should help.
FWIW the earlier comment about root hashes not matching is obsolete, and referenced an earlier draft of the spec.

internal/witness/client.go
AlCutter: I've added it back, but I wonder if this should move up into `witness/client`.

append_lifecycle.go
```diff
@@ -769,34 +769,42 @@ func (o AppendOptions) CheckpointPublisher(lr LogReader, httpClient *http.Client
...
-			eg.Go(func() error {
-				var err error
-				ws, err = witnessCheckpoint(ctx, cp, size, o.witnesses, lr, httpClient, o.witnessOpts)
-				return err
-			})
-			if mirrorGW != nil {
+			if witnessGateway != nil {
+				eg.Go(func() error {
+					var err error
+					ws, err = witnessCheckpoint(ctx, witnessGateway.CosignCheckpoint, &o.witnesses, cp, cpSize, o.witnessOpts)
```
phbnf: Do you need a `context.WithTimeout` here as well? `o.witnessOpts.Timeout` isn't applied anymore I think.

internal/witness/client.go
```diff
@@ -90,170 +83,152 @@ func init() {
...
+	if opts.HTTPClient == nil {
+		slog.WarnContext(ctx, "WitnessGateway: No HTTP client configured, using DefaultHTTPClient")
+		opts.HTTPClient = http.DefaultClient
+	}
+	if opts.FetchTiles == nil {
+		return nil, fmt.Errorf("fetch tiles is required")
+	}
+
+	witnesses := make([]*witnessClient, 0, len(opts.Witnesses))
+	for _, w := range dedup(opts.Witnesses) {
```
phbnf: Courtesy of Gemini: these checks happen after dedup is called, and dedup doesn't check if `w.URL` is nil.

internal/witness/client.go
```diff
@@ -90,170 +83,152 @@ func init() {
...
-			sigBlock.Write(r.sig)
-
-			// See whether the group is satisfied now
-			if newCp := append(slices.Clone(cp), sigBlock.Bytes()...); wg.group.Satisfied(newCp) {
-				span.AddEvent("Policy satisfied")
-				return sigBlock.Bytes(), nil
-			}
-		}
+	go func() {
+		_ = otel.TraceErr(ctx, "tessera.witnessgateway.CosignCheckpoint.closer", tracer, func(ctx context.Context, span trace.Span) error {
```
phbnf, line 233: Should the span be closed here instead?

phbnf (approved):

append_lifecycle.go
AlCutter: Good catch, thanks!

internal/witness/client.go
AlCutter: I _think_ the implicit "use-case" behind the question would actually just be answered by the span which is opened by the caller in `append_lifecycle.go`'s `witnessCheckpoint`/`mirrorCheckpoint` and `CheckpointPublisher` funcs - "how long did I spend waiting for witness/mirror cosignatures and publishing a checkpoint?"

phbnf (approved):

PR #1101, opened 2026-08-10
[MTC] Publish Landmarks

phbnf:
Towards #945 

This PR copies over Tessera's POSIX functions, and uses them to publish the Landmarks resource.

### DIFF
--- a/cmd/mtc/log/internal/landmark/landmark.go
+++ b/cmd/mtc/log/internal/landmark/landmark.go
@@ -23,7 +23,6 @@ import (
 	"slices"
 	"strconv"
 	"strings"
-	"sync/atomic"
 	"time"
 )
 
@@ -33,6 +32,19 @@ const (
 
 	// publishRetryOnFailure configures when to schedule a new publication upon failure.
 	publishRetryOnFailure = time.Second * 5
+
+	// MaxActiveLandmarks is the maximum allowed number of active landmarks over
+	// any certificate validity period.
+	//
+	// SPEC: CQRP Policy v0.2.0
+	// "MTC CA Operators MUST NOT issue Subscriber certificates with a validity
+	// period exceeding 47 days."
+	//
+	// SPEC: CQRP Policy v0.2.0
+	// "For CA Cosigners with a maximum permitted certificate validity of up to 47
+	// days, MTC CA landmarks SHOULD be generated approximately every four 4 hours,
+	// and MUST NOT exceed a total of 370 landmarks over any 47-day period."
+	MaxActiveLandmarks = 370
 )
 
 // ReadCheckpointSize returns the current log checkpoint size.
@@ -207,19 +219,12 @@ func (a *ActiveLandmarks) AddLandmark(treeSize, maxActive uint64) error {
 	return nil
 }
 
-type published struct {
-	active *ActiveLandmarks
-	pubAt  time.Time
-}
-
 // Publisher manages publication of the landmarks resource at regular intervals.
 type Publisher struct {
 	storage            LandmarksStorage
 	readCheckpointSize ReadCheckpointSize
 	maxActive          uint64
 	pubInterval        time.Duration
-
-	published atomic.Pointer[published]
 }
 
 // NewPublisher creates a new Publisher instance.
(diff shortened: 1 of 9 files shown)


cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -171,3 +206,156 @@ func (a *ActiveLandmarks) AddLandmark(treeSize, maxActive uint64) error {
...
+		}
+		if err := active.UnmarshalText(old); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal landmarks for update: %w", err)
+		}
+
+		if time.Since(oldModTime) < p.pubInterval {
+			slog.DebugContext(ctx, "landmarks update: skipping landmarks write because last update too recent", slog.Time("lastUpdate", oldModTime))
+			return old, nil
+		}
+		if cpSize <= active.latestTreeSize() {
```
AlCutter: Can `cpSize < latestTreeSize` ever happen without something _Very Bad_ having happened? If not, maybe split this case out and slog.Warning/Error?

cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -171,3 +206,156 @@ func (a *ActiveLandmarks) AddLandmark(treeSize, maxActive uint64) error {
...
+}
+
+// Publisher manages publication of the landmarks resource at regular intervals.
+type Publisher struct {
+	storage            LandmarksStorage
+	readCheckpointSize ReadCheckpointSize
+	maxActive          uint64
+	pubInterval        time.Duration
+
+	published atomic.Pointer[published]
```
AlCutter: Why `atomic.Pointer`?

Also, this never seems to be read other than in tests, but perhaps that's coming later?

cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -171,3 +206,156 @@ func (a *ActiveLandmarks) AddLandmark(treeSize, maxActive uint64) error {
...
+		return nil, errors.New("readCheckpointSize must not be nil")
+	}
+	if maxCertLifetime <= 0 {
+		return nil, errors.New("maxCertLifetime must be strictly positive")
+	}
+	if pubInterval <= 0 {
+		return nil, errors.New("pubInterval must be strictly positive")
+	}
+
+	maxActive := uint64(math.Ceil(float64(maxCertLifetime) / float64(pubInterval)))
```
AlCutter: Should this have a `+ 1` ?

AlCutter: Should we return an error if `maxActive` turns out to be way larger than expected and/or log a warning if `pubInterval > time.Hour`?

cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -171,3 +206,156 @@ func (a *ActiveLandmarks) AddLandmark(treeSize, maxActive uint64) error {
...
+	p.published.Store(&published{
+		active: activeLM,
+		pubAt:  modTime,
+	})
+	return nil
+}
+
+// start runs the background landmark publishing loop until ctx is canceled.
+func (p *Publisher) start(ctx context.Context) {
+	ticker := time.NewTicker(time.Millisecond)
```
AlCutter: `Timer` might be a better fit (one-shot vs. continuous) since you're guaranteed to reset it on every tick.

cmd/mtc/log/internal/landmark/landmark.go
phbnf: Added a maximum value based on the CQRP: https://docs.google.com/document/d/1NYkTvAqSTcXYkkYEpyfN9EkNakPZ1cbmR0U4i8KvgXI/edit?tab=t.0.

I had not realized that there are recommended values in the CQRP, so I've also reflected them for `pubInterval`. If left empty, `NewMTCLog` will  set the right values.

phbnf: Oh yes indeed, I got confused about this with the extra non active landmark at the end of this list... added a reference to specs, +1 and tests.

cmd/mtc/log/internal/landmark/landmark.go
phbnf: Yes, this will be read in a new PR and belongs there. I didn't bother removing it when splitting the large PR into smaller ones. Removed for now, such that we can focus on reviewing this then.

cmd/mtc/log/internal/landmark/landmark.go
phbnf: Done.

AlCutter (approved):

PR #1105, opened 2026-08-11
Add sign-subtree support to mirror

AlCutter:
This PR adds support for the `sign-subtree` request to the MTC mirror implementation.

Towards #945

### DIFF
--- a/cmd/mtc/mirror/internal/handler/mirror_mux.go
+++ b/cmd/mtc/mirror/internal/handler/mirror_mux.go
@@ -22,6 +22,7 @@ import (
 	"sync"
 
 	"github.com/transparency-dev/tessera"
+	"github.com/transparency-dev/tessera/internal/parse"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


cmd/mtc/mirror/internal/handler/mirror_mux.go
```diff
@@ -80,4 +99,7 @@ func (m *MirrorMux) target(origin string) (MirrorTarget, error) {
 type MirrorTarget interface {
 	// AddEntries adds verified consistent entries to the mirror.
 	AddEntries(ctx context.Context, uploadStart, uploadEnd uint64, ticket []byte, next func() (*tessera.MirrorPackage, error)) (nextIdx uint64, curSize uint64, newTicket []byte, cosigs []byte, err error)
+	// SignSubtree should verify that the provided checkpoint originates from a known log and is valid, and that the provided subtree and proof
+	// are valid for the provided checkpoint, and, if so, returns a cosignature for the subtree.
```
phbnf: I think it should also verify that the provided checkpoint contains a cosignature from that very witness. It already does this, but the comment does not reflect this.

phbnf (approved):

cmd/mtc/mirror/internal/handler/mirror_mux.go
AlCutter: Ah, yes, that is quite right (and the impl does) - I'll update the comment.

PR #1106, opened 2026-08-11
[MTC] Add `GetSubtreeFor` methods for upcoming `ProofToLandmark` implementation

phbnf:
Towards #945.

This PR introduces `GetSubtreeFor(index)` methods, which we'll use later on to build Landmark relative inclusion proofs.

It also adds a new `checkpoint` package, which takes care of fetching checkpoints, and caching their size. This package will be expanded later, to cache subtree boundaries, which we'll need later for `AddTBS()`.

### DIFF
--- a/cmd/mtc/log/internal/checkpoint/reader.go
+++ b/cmd/mtc/log/internal/checkpoint/reader.go
@@ -66,8 +66,8 @@ func (r *Reader) Checkpoint(ctx context.Context) ([]byte, error) {
 }
 
 // LatestSize returns the most recently observed checkpoint size.
-func (r *Reader) LatestSize(_ context.Context) (uint64, error) {
+func (r *Reader) LatestSize() uint64 {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
-	return r.latestSize, nil
+	return r.latestSize
 }
--- a/cmd/mtc/log/internal/checkpoint/reader.go
+++ b/cmd/mtc/log/internal/checkpoint/reader.go
@@ -1,0 +1,75 @@
+// Copyright 2026 The Tessera authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package checkpoint provides helpers for reading checkpoint and caching sizes.
+package checkpoint
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+
+	flog "github.com/transparency-dev/formats/log"
+)
+
+// Reader wraps a ReadCheckpoint function to cache the latest checkpoint size observed from storage.
+type Reader struct {
+	readCheckpoint func(context.Context) ([]byte, error)
+
+	mu         sync.RWMutex
+	latestSize uint64
+}
+
+// NewReader creates a new Reader instance wrapping readCheckpoint and reads the initial checkpoint from storage.
+func NewReader(ctx context.Context, readCheckpoint func(context.Context) ([]byte, error)) (*Reader, error) {
+	if readCheckpoint == nil {
+		return nil, errors.New("readCheckpoint must not be nil")
+	}
+	r := &Reader{
+		readCheckpoint: readCheckpoint,
+	}
+	if _, err := r.Checkpoint(ctx); err != nil {
+		return nil, fmt.Errorf("initial checkpoint read: %v", err)
+	}
+	return r, nil
+}
+
+// Checkpoint reads the latest checkpoint from underlying storage and updates the latest observed size.
+func (r *Reader) Checkpoint(ctx context.Context) ([]byte, error) {
+	rawCp, err := r.readCheckpoint(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var cp flog.Checkpoint
+	if _, err := cp.Unmarshal(rawCp); err != nil {
+		return nil, fmt.Errorf("parse checkpoint: %v", err)
+	}
+
+	r.mu.Lock()
+	if cp.Size > r.latestSize {
+		r.latestSize = cp.Size
+	}
+	r.mu.Unlock()
+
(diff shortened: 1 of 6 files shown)


cmd/mtc/log/internal/checkpoint/reader.go
```diff
@@ -0,0 +1,75 @@
...
+func (r *Reader) Checkpoint(ctx context.Context) ([]byte, error) {
+	rawCp, err := r.readCheckpoint(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var cp flog.Checkpoint
+	if _, err := cp.Unmarshal(rawCp); err != nil {
+		return nil, fmt.Errorf("parse checkpoint: %v", err)
+	}
```
AlCutter: I'm _a little_ surprised that this works with signed checkpoints :D

Might be a good idea to use `parse.CheckpointUnsafe` for these cases as it helps flag that we're skipping signature verification.

cmd/mtc/log/internal/checkpoint/reader.go
```diff
@@ -0,0 +1,75 @@
...
+
+	r.mu.Lock()
+	if cp.Size > r.latestSize {
+		r.latestSize = cp.Size
+	}
```
AlCutter:
```suggestion
	r.latestSize = max(r.latestSize, cp.Size)
```

cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -219,12 +230,54 @@ func (a *ActiveLandmarks) AddLandmark(treeSize, maxActive uint64) error {
 	return nil
 }
 
+// GetSubtreeFor returns the subtree range [start, end) of active landmarks covering index.
```
AlCutter:
```suggestion
// GetSubtreeFor returns the subtree range [start, end) of the active landmarks covering index.
```

cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -47,6 +50,14 @@ const (
 	MaxActiveLandmarks = 370
 )
 
+var (
+	// ErrTooOld indicates that an index precedes the earliest available active landmark.
+	ErrTooOld = errors.New("entry is older than earliest active landmark")
+
+	// ErrNotCovered indicates that an index is not yet covered by any published active landmark.
+	ErrNotCovered = errors.New("entry is not covered by active landmarks")
```
AlCutter: How about `ErrNotYetCovered` to be clear this is likely transient?

cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -219,12 +230,54 @@ func (a *ActiveLandmarks) AddLandmark(treeSize, maxActive uint64) error {
 	return nil
 }
 
+// GetSubtreeFor returns the subtree range [start, end) of active landmarks covering index.
+// It returns ErrTooOld if index precedes the earliest available active landmark.
+// It returns ErrNotCovered if index is not yet covered by any published active landmark.
+func (a *ActiveLandmarks) GetSubtreeFor(index uint64) (start, end uint64, err error) {
+	switch {
+	case len(a.treeSizes) == 0:
+		return 0, 0, errors.New("no landmarks available")
```
AlCutter: Maybe return a more alarming warning text - I think this is an invariant-breaking condition?

cmd/mtc/log/internal/landmark/landmark.go
```diff
@@ -219,12 +230,54 @@ func (a *ActiveLandmarks) AddLandmark(treeSize, maxActive uint64) error {
...
+}
+
 // Publisher manages publication of the landmarks resource at regular intervals.
 type Publisher struct {
 	storage            LandmarksStorage
 	readCheckpointSize ReadCheckpointSize
 	maxActive          uint64
 	pubInterval        time.Duration
+
+	published atomic.Pointer[published]
```
AlCutter: I think this is fine, but I'd love to know how you got here!

I do think you might need to be careful in the future with `active`, though, since it's a pointer out to some other struct; currently you always make a new `active{}` whenever you Store in here, but if that ever changed then your `atomic.Pointer` won't give you the semantics you're expecting.

I don't know whether it might be clearer/safer to just have a `mu` and make `active` and `pubAt` just be regular fields?

AlCutter (approved):

cmd/mtc/log/internal/landmark/landmark.go
phbnf: Hm yeah good point. This is actually not consistent with other methods, where the code just blindly assumes that there's at least a single entry in the array.

What do you think is the best approach here? Check it everywhere, or never check it? I find it quite similar to checking whether a pointer is nil before calling a method defined on it.

cmd/mtc/log/internal/landmark/landmark.go
phbnf: Hehe, I was quite looking forward to seeing what you'd say about this one.

This is mostly a stylistic decision. In an earlier implementation, Publisher _also_ kept track of a cached checkpoint size. This structure ended up with two mutexes, and it didn't look very good, so I swapped them both to atomic pointers.

If we just make them regular fields, then one needs to remember to lock / unlock the mutex, so I concluded that the tradeoff was similar to making a new `active{}`: you really need to be careful when using this.

I've reverted to a mutex: let's try and be consistent across the code base.

AlCutter: I'm a fan of `atomic` for primitives, `atomic.Pointer` definitely has a bit more subtlety to it :)

cmd/mtc/log/internal/landmark/landmark.go
AlCutter: I guess this is an exported func after all, so arguably anyone calling it must also have used the c'tor to get the instance to call on - if badness happens then that suggests there's a pretty catastrophic bug somewhere in here and it's not necessarily clear whether it's safe to carry on, so maybe just letting it `panic` is ok?

The other thought is whether there's a way to make this state impossible to represent? E.g. having a `smallestSize uint64` and an `additionalSizes []uint64`.
Maybe chew on it and do a follow-up if you come up with something?

AlCutter (approved):

cmd/mtc/log/internal/landmark/landmark.go
phbnf: I've removed this test case for now, at least it's consistent with the rest of the implementation.


`E.g. having a smallestSize uint64 and an additionalSizes []uint64.`

I did think about that yes. It would actually be very nice since there are actually two types of sizes in `/landmarks`: the oldest non active size, and then the active sizes. I tried to separate it out, but then:
 
 - indexes don't match with indexes in the specs
 - our internal representation doesn't match with the specs

PR #1108, opened 2026-08-13
Use mirror witness for add-entries checkpoint signatures

AlCutter:
This PR builds upon the introduction of a mirror-witness in #1105 by fully delegating to it for _all_ mirror cosignatures.

Towards #945

### DIFF
--- a/mirror_lifecycle.go
+++ b/mirror_lifecycle.go
@@ -375,15 +375,6 @@ func (mt *MirrorTarget) publishCheckpoint(ctx context.Context, newCP []byte, new
 		return nil, 0, fmt.Errorf("failed to create proof builder: %w", err)
 	}
 
-	oldSize := mt.oldSize.Load()
-	var cProof [][]byte
-	if oldSize > 0 {
-		cProof, err = pb.ConsistencyProof(ctx, oldSize, newCPSize)
-		if err != nil {
-			return nil, 0, fmt.Errorf("failed to get consistency proof: %w", err)
-		}
-	}
-
 	// SPEC: Finally, the mirror performs the following steps atomically. Note the mirror
 	// 			checkpoint may have changed since the start of this process.
 	//  			- Check if upload_end is still greater than or equal to the mirror checkpoint's tree size.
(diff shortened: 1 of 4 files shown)


mirror_lifecycle.go
```diff
@@ -450,29 +450,29 @@ func (mt *MirrorTarget) openOrCreateTicket(ctx context.Context, ticketBytes []by
 
 	if pendingCP == nil || pendingCP.Size != expectedSize {
 		slog.DebugContext(ctx, "Invalid or incorrect ticket, returning new ticket", slog.Uint64("next_entry", nextEntry), slog.String("ticketCP", string(ticketCP)), slog.Uint64("expectedSize", expectedSize))
-		ticketBytes, pendingCP, pendingNote, err = mt.createNewTicket(ctx)
+		ticketBytes, pendingCP, ticketCP, err = mt.createNewTicket(ctx)
```
phbnf: s/ticketCP/pendingCPRaw?

mirror_lifecycle.go
```diff
@@ -371,47 +369,50 @@ func (mt *MirrorTarget) bundleIterator(ctx context.Context, next func() (*Mirror
...
+				continue
+			}
+			slog.WarnContext(ctx, "Permanent failure updating checkpoint on mirror witness", slog.Uint64("oldSize", oldSize), slog.Uint64("newSize", wSize), slog.Any("error", err))
+			return nil, wSize, fmt.Errorf("failed to update checkpoint: %w", err)
+		}
+		done = true
 	}
+	mt.oldSize.CompareAndSwap(oldSize, newCPSize)
+
+	slog.WarnContext(ctx, "Completed upload", slog.Uint64("oldSize", oldSize), slog.Uint64("newSize", newCPSize))
```
phbnf: Should this be a warning? I can see argument for both. I noticed something similar earlier when doing the full stack test, some successes (I can't remember which, but I'm sure I'll see then again) appeared as warning.

mirror_lifecycle.go
```diff
@@ -371,47 +369,50 @@ func (mt *MirrorTarget) bundleIterator(ctx context.Context, next func() (*Mirror
...
+	// 			HTTP status code. The response body MUST be formatted as in a witness's successful add-checkpoint
+	// 			response: a sequence of one or more note signature lines.
+	var wSigs []byte
+	for done := false; !done; {
+		var wSize uint64
+		wSigs, wSize, err = mt.mirrorWitness.Update(ctx, oldSize, newCP, cProof)
+		if err != nil {
+			if errors.Is(err, witness.ErrCheckpointStale) {
+				slog.WarnContext(ctx, "Retrying stale checkpoint on mirror witness", slog.Uint64("oldSize", oldSize), slog.Uint64("newSize", wSize))
+				oldSize = wSize
```
phbnf: Does the consistency proof need to be updated when this happens?

mirror_lifecycle.go
AlCutter: No, I bumped this up to Warn while debugging so as to see the slog in test output but forgot to bump it back down again :)

mirror_lifecycle.go
AlCutter: Great spot, yes!

phbnf (approved):

PR #1109, opened 2026-08-13
[MTC] Implement ProofToLandmark endpoint

phbnf:
Towards #945.

### DIFF
--- a/cmd/mtc/log/internal/handler/handlers.go
+++ b/cmd/mtc/log/internal/handler/handlers.go
@@ -35,8 +35,8 @@ const (
 	maxAddTBSRequestBodyBytes = 128 << 10
 )
 
-type addTBS func(context.Context, log.TBSCertificateLogEntry) (*log.AddTBSRsp, error)
-type proofToLandmark func(context.Context, uint64) ([]byte, time.Duration, error)
+type addTBSFn func(context.Context, log.TBSCertificateLogEntry) (*log.AddTBSRsp, error)
+type proofToLandmarkFn func(context.Context, uint64) ([]byte, time.Duration, error)
 
 // New returns a new http.Handler for the mtc-tlog service.
 func New(mtcLog *log.MTCLog) http.Handler {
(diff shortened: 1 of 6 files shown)


cmd/mtc/log/internal/handler/handlers.go
```diff
@@ -32,19 +36,13 @@ const (
 )
 
 type addTBS func(context.Context, log.TBSCertificateLogEntry) (*log.AddTBSRsp, error)
+type proofToLandmark func(context.Context, uint64) ([]byte, time.Duration, error)
```
AlCutter: nit: consider adding a `Fn` or `Func` suffix to these?

cmd/mtc/log/internal/handler/handlers.go
```diff
@@ -60,18 +58,18 @@ func addTBSHandler(add addTBS) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		defer func() {
 			if err := r.Body.Close(); err != nil {
-				slog.ErrorContext(r.Context(), "resp.Body.Close()", slog.Any("error", err))
+				slog.ErrorContext(r.Context(), "r.Body.Close()", slog.Any("error", err))
```
AlCutter: This could probably be a `DebugContext`, it'll almost certainly be because the client closed the socket early or similar.

cmd/mtc/log/mtc.go
```diff
@@ -412,14 +438,61 @@ func (l *MTCLog) AddTBS(ctx context.Context, tbs TBSCertificateLogEntry) (*AddTB
...
+		return nil, 0, fmt.Errorf("get subtree for index %d: %v", index, err)
+	case retryAfter > 0:
+		jitter := time.Duration(rand.Int64N(int64(maxRetryAfterJitter)))
+		return nil, retryAfter + landmarkPublicationBuffer + jitter, nil
+	}
+
+	// Construct the inclusion proof to the active landmark.
+	pb, err := client.NewProofBuilder(ctx, end, l.reader.ReadTile)
+	if err != nil {
+		return nil, 0, fmt.Errorf("cannot create proof builder")
```
AlCutter: `%v` the err?

cmd/mtc/log/mtc.go
```diff
@@ -41,6 +45,22 @@ const (
 	// "MTC CA Operators MUST NOT issue Subscriber certificates with a
 	// validity period exceeding 47 days."
 	DefaultMaxCertLifetime = 47 * 24 * time.Hour
+
+	// landmarkPublicationBuffer is an additional buffer added to RetryAfter in ProofToLandmark
+	// responses to account for landmark publication latency (integrating entries, signing, and storage writes).
+	landmarkPublicationBuffer = 500 * time.Millisecond
```
AlCutter, line 52: I could imagine this might get tricky, e.g. if there's a slower "replication" component involved in exposing the tree, but also let's not worry about that for the mo' (maybe add a TODO?).

AlCutter (approved):

cmd/mtc/log/mtc.go
phbnf: Left a TODO for now. I started without this buffer, then added it with the jitter out of caution. I suspect we'll know more about this once we receive feedback from CAs about usage patterns.

PR #1110, opened 2026-08-13
Enable "greedy" witnessing

### LINKED ISSUE #866: Feature: support a configurable witness-wait-delay
If I understand correctly, Tessera will return a checkpoint *immediately* once it has enough witness cosignatures to achieve quorum.

I'd like to be able to configure Tessera to wait (for a _short_ period) for any other witness cosignatures to come in before returning the signed checkpoint. This would enable a best-effort attempt to gather as many witness cosignatures as possible.

AlCutter:
This PR allows logs to configure witnessing to include any "surplus" cosignatures (i.e. beyond the strict set needed to satisfy the witness quorum policy) which can be fetched within the timeout.

Resolves #866

### DIFF
--- a/append_lifecycle.go
+++ b/append_lifecycle.go
@@ -797,7 +797,7 @@ func (o AppendOptions) CheckpointPublisherContext(ctx context.Context, lr LogRea
 					defer cancel()
 
 					var err error
-					ws, err = witnessCheckpoint(ctx, witnessGateway.CosignCheckpoint, &o.witnesses, cp, cpSize, o.witnessOpts.FailOpen)
+					ws, err = witnessCheckpoint(ctx, witnessGateway.CosignCheckpoint, &o.witnesses, cp, cpSize, o.witnessOpts.FailOpen, o.witnessOpts.Greedy)
 					return err
 				})
 			}
@@ -808,7 +808,7 @@ func (o AppendOptions) CheckpointPublisherContext(ctx context.Context, lr LogRea
 					defer cancel()
 
 					var err error
-					ms, err = mirrorCheckpoint(ctx, mirrorGateway.CosignCheckpoint, &o.mirrors, cp, cpSize, o.mirrorOpts.FailOpen)
+					ms, err = mirrorCheckpoint(ctx, mirrorGateway.CosignCheckpoint, &o.mirrors, cp, cpSize, o.mirrorOpts.FailOpen, false)
 					return err
 				})
 			}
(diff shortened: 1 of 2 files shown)


append_lifecycle.go
```diff
@@ -946,15 +949,19 @@ func gatherCosignatures(ctx context.Context, name string, fetcher cosigSource, p
...
 				}
+				gotResponses++
 				sigBlock.Write(sig)
 				// Don't allow failOpen here, or we'll break out of the collection loop prematurely.
-				if sigs, err := checkPolicy(sigBlock.Bytes(), false); err == nil {
+				sigs, err := checkPolicy(sigBlock.Bytes(), false)
+				// We can return with sigs early if we've met policy and either we're not in greedy mode or we've received a response from all configured witnesses.
+				if err == nil && (!greedy || gotResponses == maxExpectedResponses) {
 					return sigs, nil
 				}
```
roger2hk, line 970: If we check the `greedy` and `gotResponse == maxExpectedResponses` conditions earlier, we can slightly improve the performance by calling less `checkPolicy()`.

```go
if !greedy || gotResponses == maxExpectedResponses {
    if sigs, err := checkPolicy(sigBlock.Bytes(), false); err == nil {
        return sigs, nil
    }
}
```

append_lifecycle_test.go
```diff
@@ -581,3 +893,14 @@ func newMirrorHandler(t *testing.T, mirrorSKey string) http.HandlerFunc {
 		}
 	}
 }
+func mustNewWitness(t *testing.T, skey, urlStr string) Witness {
+	url, err := url.Parse(urlStr)
+	if err != nil {
+		t.Fatalf("Failed to parse URL %s: %v", urlStr, err)
+	}
+	wit, err := NewWitness(skey, url)
```
roger2hk: This should be a verifier key. Updating the naming:

```suggestion
func mustNewWitness(t *testing.T, vkey, urlStr string) Witness {
	url, err := url.Parse(urlStr)
	if err != nil {
		t.Fatalf("Failed to parse URL %s: %v", urlStr, err)
	}
	wit, err := NewWitness(vkey, url)
```

append_lifecycle.go
```diff
@@ -917,7 +917,9 @@ var errFailedOpen = errors.New("failed-open")
 
 // gatherCosignatures gathers signatures from a source, applying a policy to determine if the signatures are sufficient.
 // It returns the first set of signatures which satisfy the policy, or an error if the policy is not met and failOpen is false.
```
roger2hk: Include the `greedy` mode in the godoc comment?

roger2hk (approved):

append_lifecycle_test.go
AlCutter: whoops, you're right.

PR #1112, opened 2026-08-13
[MTC] Define MTCProof structure

phbnf:
Towards #945.

This PR:
 - introduces `MTCProof` and its serialization functions.
 A series of checks are done in `new` to prevent building a proof that wouldn't be specs compliant at serialization time. I've gone quite maximalists on all the checks, but I think it should be future proof. Some of them are required (for instance, making sure we don't have too many signatures), others less because they already happen somewhere else in the codebase (like the length of cosigner IDs)... but it's probably better to have them all and call it done.
 - returns `MTCProof` as `[]byte` in the public API rather than an actual proof.
 I could be convinced otherwise, though this decision is mostly rooted in that:
   - it reduces our API surface
   - this `MTCProof`, really, is a signature, and if we return `[]byte`, we comply with the MTC specs, and the CA can use this signature directly.
   - it _might_ (should, must?!) be a good idea for the CA to verify `MTCProof` before they stick it in a cert. I'm thinking that in a followup PR, we could export a `func Verify(proofBytes []byte, tbs TBSCertificateLogEntry, index uint64, origin string, verifiers []note.SubtreeVerifier) error` function, WDYT?

### DIFF
--- a/cmd/mtc/log/internal/mtcproof/mtcproof.go
+++ b/cmd/mtc/log/internal/mtcproof/mtcproof.go
@@ -1,0 +1,249 @@
+// Copyright 2026 The Tessera authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mtcproof
+
+import (
+	"bytes"
+	"cmp"
+	"crypto/sha256"
+	"crypto/x509"
+	"fmt"
+	"slices"
+	"strings"
+
+	"golang.org/x/crypto/cryptobyte"
+)
+
+const (
+	oidPrefix           = "oid/1.3.6.1.4.1."
+	penPrefixDERLen     = 5
+	maxTrustAnchorIDLen = (1 << 8) - 1
+	maxUint48           = (1 << 48) - 1
+)
+
+// hashValue represents a 32-byte hash as per
+// draft-ietf-plants-merkle-tree-certs section 6.2.
+type hashValue [sha256.Size]byte
+
+// SubtreeSignature represents a cosigner's signature on a subtree root
+// as per draft-ietf-plants-merkle-tree-certs section 6.2:
+type SubtreeSignature struct {
+	// CosignerID is the binary representation of the trust anchor ID (1..255 bytes).
+	CosignerID []byte
+	// Signature is the raw signature bytes over the subtree.
+	Signature []byte
+}
+
+// mtcProof represents an MTC inclusion proof as per
+// draft-ietf-plants-merkle-tree-certs section 6.2.
+//
+// New instances MUST be created with new().
+type mtcProof struct {
+	extensions     []byte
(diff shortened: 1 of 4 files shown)


cmd/mtc/log/internal/mtcproof/mtcproof.go
```diff
@@ -0,0 +1,249 @@
...
+	"fmt"
+	"slices"
+	"strings"
+
+	"golang.org/x/crypto/cryptobyte"
+)
+
+const (
+	oidPrefix           = "oid/1.3.6.1.4.1."
+	penPrefixDERLen     = 5
```
AlCutter: What does the "pen" bit mean?

cmd/mtc/log/internal/mtcproof/mtcproof.go
```diff
@@ -0,0 +1,249 @@
...
+	res := der[penPrefixDERLen:]
+	if l := len(res); l == 0 || l > maxTrustAnchorIDLen {
+		return nil, fmt.Errorf("cosigner ID binary length must be 1..%d bytes, got %d", maxTrustAnchorIDLen, l)
+	}
+	return res, nil
+}
+
+// New validates MTC proof parameters, orders signatures canonically, and returns
+// the TLS-encoded MTCProof binary bytes.
+func New(extensions []byte, start, end uint64, inclusionProof [][]byte, signatures []SubtreeSignature) ([]byte, error) {
```
AlCutter: How about `Serialize` (or similar)?
`New` tripped me up because I thought it was a c'tor.

cmd/mtc/log/internal/mtcproof/mtcproof.go
```diff
@@ -0,0 +1,249 @@
...
+		return nil, err
+	}
+	return p.marshal()
+}
+
+// new constructs a validated mtcProof with signatures sorted canonically.
+func new(extensions []byte, start, end uint64, inclusionProof [][]byte, signatures []SubtreeSignature) (*mtcProof, error) {
+	// SPEC: draft-ietf-plants-merkle-tree-certs section 6.2.
+	// "uint48 start; uint48 end;"
+	if start > maxUint48 {
```
AlCutter: Should this be `start >= maxUint48` since `start` must also be `< end`?

(you could put the `start >= end` check up front if you want the clearer error message for the border case)

cmd/mtc/log/internal/mtcproof/mtcproof.go
```diff
@@ -0,0 +1,249 @@
...
+		return nil, fmt.Errorf("end index %d exceeds uint48 maximum (%d)", end, maxUint48)
+	}
+	if start >= end {
+		return nil, fmt.Errorf("start (%d) must be strictly less than end (%d)", start, end)
+	}
+	// SPEC: draft-ietf-plants-merkle-tree-certs section 5.2.1.
+	// "opaque extension_data<0..2^16-1>; "
+	if l := len(extensions); l >= 1<<16 {
+		return nil, fmt.Errorf("extensions too large (%d bytes, max %d)", l, (1<<16)-1)
+	}
```
AlCutter: Maybe update all these checks to be of the form `if limit, l := 1<<16, len(extensions); l >= limit { ... limit-1 }` to ensure error message consistency with check?

AlCutter (approved):

cmd/mtc/log/internal/mtcproof/mtcproof.go
phbnf: Private Enterprise Number. I've renamed it to `oidPrefixDERLen`. It's a _bit_ confusing because `oidPrefix` contains `oid/` but `oidPrefix*DER*Len` doesn't include the `oid/` prefix, but I think that's fine.

AlCutter (approved):

PR #1118, opened 2026-08-14
Refactor mirror library

AlCutter:
This PR does a small refactor of the MTC mirror binary, adding a simple API for instantiating a mirror service.

This is intended to make it easier to re-use when creating other binaries (e.g. other storage implementations, or tests).

Towards #945

### DIFF
--- a/cmd/mtc/mirror/mirror.go
+++ b/cmd/mtc/mirror/mirror.go
@@ -1,0 +1,95 @@
+// Copyright 2026 The Tessera Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package mirror provides functionality for creating and running tlog-mirror services.
+package mirror
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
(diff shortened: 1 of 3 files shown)


cmd/mtc/mirror/mirror.go
```diff
@@ -0,0 +1,95 @@
...
+	}, nil
+}
+
+// ServeHTTP implements the http.Handler interface, and exposes the tlog-mirror protocol via HTTP.
+func (m *Mirror) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	m.handler.ServeHTTP(w, r)
+}
+
+// assertDistinctSigners asserts that the two provided lists of cosigners have no signers in common.
+func assertDistinctSigners(w, m []note.Signer) error {
```
phbnf, line 80: You _could_ add some tests for this one.

cmd/mtc/mirror/posix/main.go
```diff
@@ -122,35 +106,9 @@ func main() {
...
-		WithOrigin(origin).
-		WithLogVerifier(logVerifier).
-		WithSigner(mirrorSigner)
-	return tessera.NewMirrorTarget(ctx, d, mOpts)
-}
-
 // mirrorConfigFromFlags returns a mirror configuration loaded from the provided flags.
 // Exits if the mirror configuration could not be loaded.
-func mirrorConfigFromFlags(ctx context.Context) []witnessConfig.Log {
+func mirrorConfigFromFlags(ctx context.Context) []mirror.LogConfig {
```
phbnf, line 111: This now does more than returning a mirror configuration, it also initializes mirror storage I think.

phbnf (approved):

cmd/mtc/mirror/mirror.go
AlCutter: Good call.

cmd/mtc/mirror/posix/main.go
AlCutter: I've made the comment a bit more accurate while leaving details out.

PR #1121, opened 2026-08-17
[MTC] Store subtrees boundaries to build `AddTBS()` `MTCProof` structures

phbnf:
Towards #945.

This PR computes and fetches the subtree boundaries aligned with checkpoint sizes. We'll use these boundaries later to build consistency proofs for `AddTBS()` requests.

### DIFF
--- a/cmd/mtc/log/internal/checkpoint/reader.go
+++ b/cmd/mtc/log/internal/checkpoint/reader.go
@@ -139,9 +139,6 @@ func (r *Reader) SubtreeForIndex(index uint64) (start, end uint64, ok bool) {
 		}
 	}
 
-	if index < r.subtrees[0].start {
-		return 0, r.subtrees[0].start, true
-	}
-
-	return 0, 0, false
+	// If not found in retained subtrees, it must precede the earliest retained subtree.
+	return 0, r.subtrees[0].start, true
 }
--- a/cmd/mtc/log/internal/checkpoint/reader.go
+++ b/cmd/mtc/log/internal/checkpoint/reader.go
@@ -21,15 +21,31 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/transparency-dev/merkle/proof"
 	"github.com/transparency-dev/tessera/internal/parse"
 )
 
-// Reader wraps a ReadCheckpoint function to cache the latest checkpoint size observed from storage.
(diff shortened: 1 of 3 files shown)


cmd/mtc/log/internal/checkpoint/reader.go
```diff
@@ -59,15 +76,72 @@ func (r *Reader) Checkpoint(ctx context.Context) ([]byte, error) {
...
+	// Search backwards from the end because index is almost always in the most
+	// recently added subtree (e.g. during AddTBS).
+	for i := len(r.subtrees) - 1; i >= 0; i-- {
+		st := r.subtrees[i]
+		if index >= st.start && index < st.end {
+			return st.start, st.end, true
+		}
+	}
+
+	if index < r.subtrees[0].start {
```
AlCutter: Will `index` always be < `r.subtrees[0].start` at this point?

cmd/mtc/log/internal/checkpoint/reader.go
```diff
@@ -59,15 +76,72 @@ func (r *Reader) Checkpoint(ctx context.Context) ([]byte, error) {
...
+		}
+	}
 
 	return rawCp, nil
 }
 
+// pushSubtreeLocked adds a subtree and potentially removes old ones, capping the number of elements.
+// r.mu MUST be locked when calling this method.
+func (r *Reader) pushSubtreeLocked(st subtree) {
+	if len(r.subtrees) >= maxSubtrees {
```
AlCutter, line 105: Would `container/ring` be a more natural structure for `subtrees`?

AlCutter (approved):

cmd/mtc/log/internal/checkpoint/reader.go
phbnf: Great point, updated.

cmd/mtc/log/internal/checkpoint/reader.go
phbnf: Maybe, or even `container/list`. The implementation looked simple enough that I concluded it did not justify using them, but I'm happy to reconsider this in a follow up PR. Conversations with Gemini comforted me that it was the right decision: either option would likely require manual type assertions and, heap allocations.

PR #1122, opened 2026-08-17
Add checkpoint publication timeout in `AppendOptions`

roger2hk:


append_lifecycle.go
```diff
@@ -713,6 +717,9 @@ func (o AppendOptions) valid() error {
 	if o.checkpointRepublishInterval > 0 && o.checkpointRepublishInterval < o.checkpointInterval {
 		return fmt.Errorf("invalid AppendOptions: WithCheckpointRepublishInterval (%d) is smaller than WithCheckpointInterval (%d)", o.checkpointRepublishInterval, o.checkpointInterval)
 	}
+	if o.checkpointPublicationTimeout <= o.witnessOpts.Timeout {
```
AlCutter: Should we return an error here, or perhaps just log a warning and set the timeout to the largest of `checkpointPublicationTimeout`, `witnessOpts.Timeout` and `mirroringOpts.Timeout`?

AlCutter (approved):

append_lifecycle.go
roger2hk: Switched to use the warning message.

I think we could still consider returning an error here in long run. It may cause some interrupt but it is intended to ensure the values are correctly set up. Most applications should get this blocked in non-production environment if the witness timeout value is larger than 5 seconds.

AlCutter: What about `mirrorOpts.Timeout`?

Maybe worth popping a TODO about the error, could also perhaps change the logged warning to say that it'll become a fatal error at some point?

My worry with doing it at the moment is that someone who previously had a working service could end up with a crash-looping service by just updating the dep on Tessera.

roger2hk: Ah, good catch. I totally missed mirror. Added it back.

It's pretty hard to get the attention without crash-looping the service. (Hopefully that only happens in non-production environment)

AlCutter (approved):

PR #1125, opened 2026-08-18
[MTC] Implement MTCProof generation in AddTBS

phbnf:
Towards #945

### DIFF
--- a/cmd/mtc/log/mtc_test.go
+++ b/cmd/mtc/log/mtc_test.go
@@ -18,15 +18,23 @@ import (
 	"bytes"
 	"context"
 	stdasn1 "encoding/asn1"
+	"errors"
 	"fmt"
 	"os"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/transparency-dev/formats/note"
+	"github.com/transparency-dev/merkle/proof"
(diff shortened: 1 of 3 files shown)


cmd/mtc/log/mtc_test.go
```diff
@@ -673,12 +681,243 @@ func TestNewMTCLog(t *testing.T) {
...
+			CosignerID: append([]byte(nil), cosignerID...),
+			Signature:  append([]byte(nil), sig...),
+		})
+	}
+
+	if !s.Empty() {
+		return nil, errors.New("trailing bytes after MTCProof")
+	}
+	return &p, nil
+}
```
AlCutter, line 770: I wonder if this could/should be in `cmd/mtc/log/internal/mtcproof` next to the corresponding `Serialize`? 
It's an internal package so nobody can depend on it so seems safe enough?

If so, maybe call it `Deserialize` to match the existing `Serialize` func?

AlCutter (approved):

cmd/mtc/log/mtc_test.go
phbnf: SGTM, I'll send another PR to do this, and will spend a bit more time polishing the `Deserialize` function since it would now be out of tests.

PR #1126, opened 2026-08-18
[MTC] Add WithOrigin and WithSubtreeSigner to MTC Log and sign Subtrees

phbnf:
Towards #945

### DIFF
--- a/cmd/mtc/log/mtc.go
+++ b/cmd/mtc/log/mtc.go
@@ -26,6 +26,8 @@ import (
 	"time"
 
 	"github.com/transparency-dev/formats/note"
+	"github.com/transparency-dev/merkle/proof"
+	"github.com/transparency-dev/merkle/rfc6962"
 	"github.com/transparency-dev/tessera"
 	"github.com/transparency-dev/tessera/api/layout"
 	"github.com/transparency-dev/tessera/client"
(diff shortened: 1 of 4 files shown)


cmd/mtc/log/mtc.go
```diff
@@ -445,13 +485,24 @@ func (l *MTCLog) AddTBS(ctx context.Context, tbs TBSCertificateLogEntry) (*AddTB
...
 	}
+	subRoot, err := proof.RootFromInclusionProof(rfc6962.DefaultHasher, idx.Index-start, end-start, tEntry.LeafHash(), proofNodes)
+	if err != nil {
+		return nil, fmt.Errorf("cannot compute subtree root for index %d in subtree [%d, %d): %w", idx.Index, start, end, err)
+	}
+
+	selfSig, err := l.subtreeSigner.SignSubtree(0, l.origin, start, end, subRoot)
+	if err != nil {
+		return nil, fmt.Errorf("cannot sign subtree [%d, %d): %w", start, end, err)
+	}
```
AlCutter, line 496: Should this be being done in something along the lines of a "subtree-aware awaiter"?

As-is, it looks like we'll hit the HSM to sign (frequently the same-) subtree for every single added leaf, which doesn't _seem_ right?

Issue #1123, opened 2026-08-17
Tessera witnessing does not support "optional" witnesses


BenBirt:
Assuming I'm reading the code correctly: `WithWitnesses` (https://github.com/transparency-dev/tessera/blob/74fce42c1afed0a0dab12744d59ed30960fdbe98/append_lifecycle.go#L1133) only allows passing a `WitnessGroup`, i.e. ~something like the quorum of a `tlog-policy`.

However, in some circumstances, I'd like to be able to configure a witness which does _not_ contribute to `quorum`. Unfortunately I don't think this API enables me to do that. (But happy be told I'm wrong.) As such those "optional" witnesses are degraded to "never" witnesses.

Perhaps this is an opportunity to switch over to https://github.com/transparency-dev/formats/blob/main/policy/tlog_policy.go?

PR #1128, opened 2026-08-19
Migrate to tlog policy

BenBirt:
(This is a continuation of @AlCutter's https://github.com/transparency-dev/tessera/pull/1127. PR text copied below.)

This PR migrates Tessera to use the policy package from the formats repo when configuring witness and mirroring.

The old in-repo approach has been marked as deprecated, but an effort has been made to maintain backwards compatibility for folks already using that approach. This has made the diff quite a bit larger than it otherwise would have been.

Towards https://github.com/transparency-dev/tessera/issues/1123


append_lifecycle_test.go
```diff
@@ -358,10 +356,7 @@ func TestWithWitnesses(t *testing.T) {
 		},
 	} {
 		t.Run(test.desc, func(t *testing.T) {
-			opts := NewAppendOptions().WithWitnesses(witnesses, test.witnessOpts)
-			if len(opts.witnesses.Components) != 1 {
```
phbnf, line 362: Should this check len(opts.wintessPolicy.Witnesses)?

witness.go
```diff
@@ -15,315 +15,237 @@
...
-		if c.Satisfied(cp) {
-			satisfaction++
-		}
-		if satisfaction >= wg.N {
-			return true
-		}
+}
+
+// alloc returns a unique name, preferring want.
+func (a *nameAllocator) alloc(want string) string {
```
phbnf, line 174: Should this block empty strings for "want"?Or maybe it's not relevant because no code would ever call that with an empty string?

PR #1129, opened 2026-08-19
[MTC] WithOrigin, WithSubtreeSigner, WithSubtreeWitnesses and WithHTTPClient options.

phbnf:
Towards #945 

These will be used to sign subtrees.
While I'm here:
 - add a `checkOriginSignerName` functions to make sure that the origin and the signer name match for an MTC Log
 - make `TestNewMTCLog` tests tabular

### DIFF
--- a/cmd/mtc/log/mtc.go
+++ b/cmd/mtc/log/mtc.go
@@ -23,6 +23,8 @@ import (
 	"fmt"
 	"log/slog"
 	"math/rand/v2"
+	"net/http"
+	"strings"
 	"time"
 
 	"github.com/transparency-dev/formats/note"
@@ -101,6 +103,10 @@ type Options struct {
 	landmarkStorage  landmark.LandmarksStorage
 	landmarkInterval time.Duration
 	maxCertLifetime  time.Duration
+	origin           string
+	subtreeSigner    note.SubtreeSigner
+	subtreeWitnesses tessera.WitnessGroup
+	httpClient       *http.Client
 }
 
 // NewOptions creates a new options struct for configuring MTCLog instances.
@@ -118,6 +124,12 @@ func (o *Options) valid() error {
 	}
 	if o.landmarkStorage == nil {
 		return errors.New("invalid Options: WithLandmarksStorage must be set")
+	}
+	if o.origin == "" {
+		return errors.New("invalid Options: WithOrigin must be set")
+	}
+	if o.subtreeSigner == nil {
+		return errors.New("invalid Options: WithSubtreeSigner must be set")
 	}
 	if o.landmarkInterval < 0 {
 		return errors.New("invalid Options: WithLandmarkInterval must be >= 0")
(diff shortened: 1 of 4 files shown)


cmd/mtc/log/mtc.go
```diff
@@ -398,13 +440,26 @@ func NewMTCLog(ctx context.Context, a *tessera.Appender, opts *Options) (*MTCLog
 		return nil, fmt.Errorf("failed to initialise landmark publisher: %v", err)
 	}
 
+	logCosignerID, err := mtcproof.ParseCosignerID(opts.subtreeSigner.Name())
+	if err != nil {
+		return nil, fmt.Errorf("invalid subtree signer name %q: %w", opts.subtreeSigner.Name(), err)
+	}
+
+	if err := checkOriginSignerName(opts.origin, opts.subtreeSigner.Name()); err != nil {
+		return nil, fmt.Errorf("checkOrgiginSignerName: %v", err)
```
roger2hk:
```suggestion
		return nil, fmt.Errorf("checkOriginSignerName: %v", err)
```

roger2hk (approved):

cmd/mtc/log/mtc.go
```diff
@@ -398,13 +445,26 @@ func NewMTCLog(ctx context.Context, a *tessera.Appender, opts *Options) (*MTCLog
 		return nil, fmt.Errorf("failed to initialise landmark publisher: %v", err)
 	}
 
+	logCosignerID, err := mtcproof.ParseCosignerID(opts.subtreeSigner.Name())
+	if err != nil {
+		return nil, fmt.Errorf("invalid subtree signer name %q: %w", opts.subtreeSigner.Name(), err)
+	}
+
+	if err := checkOriginSignerName(opts.origin, opts.subtreeSigner.Name()); err != nil {
+		return nil, fmt.Errorf("checkOrginSignerName: %v", err)
```
roger2hk:
```suggestion
		return nil, fmt.Errorf("checkOriginSignerName: %v", err)
```

There is still a typo...

cmd/mtc/log/mtc.go
```diff
@@ -174,13 +186,48 @@ func (o *Options) WithMaxCertLifetime(duration time.Duration) *Options {
...
+// The same origin MUST be used with Tessera's WithOrigin() option.
+// TODO: think of whether we could enforce this matching.
+func (o *Options) WithOrigin(origin string) *Options {
+	o.origin = origin
+	return o
+}
+
+// WithSubtreeSigner configures the signer used to produce subtree signatures for MTC proofs.
+// signer MUST not be nil, otherwise valid() will fail.
+// The same signer MUST be used with Tessera's WithCheckpointSigner() option.
```
roger2hk, line 200: Is this enforced somewhere (e.g. `valid()`)?

roger2hk (approved):

cmd/mtc/log/mtc.go
phbnf: No, we can't at the moment since Tessera and the MTC Log are defined independently. `valid()` is only an MTC thing. I thought about reading the checkpoint at check that it's been signed by the same signer... but checkpoints are signed by multiple signers, so it's not straight forward.

There's a TODO to enforce the see if we can enforce matching origins later to get back to all of this.

PR #1130, opened 2026-08-20
[MTC] Lazily cosign subtrees on demand with log signer

phbnf:
Towards #945.

This PR signs subtrees and caches signatures for all `AddTBS()` requests to use.

This is a starting point, and this can be improved later with:
 - better retries if need be
 - returning and exporting the `subtree` structure (let me know if you want that, and I'll do it in this or a followup PR, it made the diff too hard to consume)
 - moving some functions over to t-dev/merkle

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -285,8 +285,10 @@ func subtreeNodes(start, end uint64) (proof.Nodes, error) {
 }
 
 // isSubtreeValid returns whether a subtree covers a valid range.
+// copied over from t-dev/merkle for the time being.
 func isSubtreeValid(start, end uint64) error {
 	bitCeil := func(n uint64) uint64 {
+		// copied over from t-dev/merkle for the time being.
 		if n <= 1 {
 			return 1
 		}
@@ -303,6 +305,7 @@ func isSubtreeValid(start, end uint64) error {
 		}
 		return nil
 	}
+
 	bc := bitCeil(l)
 	if start%bc != 0 {
 		return fmt.Errorf("start %d not a multiple of bit_ceil(end - start) = %d", start, bc)
--- a/client/client.go
+++ b/client/client.go
@@ -21,6 +21,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"math/bits"
 	"sync"
 
 	lru "github.com/hashicorp/golang-lru/v2"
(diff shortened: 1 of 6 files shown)


cmd/mtc/log/internal/checkpoint/reader.go
```diff
@@ -86,26 +96,57 @@ func (r *Reader) Checkpoint(ctx context.Context) ([]byte, error) {
...
 // pushSubtreeLocked adds a subtree and potentially removes old ones, capping the number of elements.
 // When capacity is exceeded, the oldest delta subtree is merged into subtrees[0], keeping subtrees[0]
 // covering [0, subtrees[1].end) before shifting so that the log start is always anchored at 0.
 // r.mu MUST be locked when calling this method.
-func (r *Reader) pushSubtreeLocked(st subtree) {
+func (r *Reader) pushSubtreeLocked(start, end uint64, rawCp []byte) {
+	st := r.makeSubtree(start, end, rawCp)
 	if len(r.subtrees) >= maxSubtrees {
-		r.subtrees[0].end = r.subtrees[1].end
+		r.subtrees[0] = r.makeSubtree(0, r.subtrees[1].end, rawCp)
```
roger2hk, line 149: Can the `r.subtrees[0].signatures` be cached to call less `makeSubtree()`?

client/client.go
```diff
@@ -249,6 +250,66 @@ func (pb *ProofBuilder) SubtreeInclusionProof(ctx context.Context, index, start,
...
+	}
+	return nil
+}
+
+func bitCeil(n uint64) uint64 {
+	if n <= 1 {
+		return 1
+	}
+	return 1 << (64 - bits.LeadingZeros64(n-1))
+}
```
roger2hk: Potential integer overflow in this function.

client/client_test.go
```diff
@@ -351,6 +351,65 @@ func TestNodeFetcherAddressing(t *testing.T) {
...
+		t.Fatalf("NewProofBuilder: %v", err)
+	}
+
+	tests := []struct {
+		name    string
+		start   uint64
+		end     uint64
+		want    []byte
+		wantErr bool
+	}{
```
roger2hk, line 388: Add one more test for valid range with non-zero start?

cmd/mtc/log/mtc.go
```diff
@@ -425,10 +426,31 @@ func NewMTCLog(ctx context.Context, a *tessera.Appender, opts *Options) (*MTCLog
...
 
-	cpReader, err := checkpoint.NewReader(ctx, opts.reader.ReadCheckpoint)
+	logCosignerID, err := mtcproof.ParseCosignerID(opts.subtreeSigner.Name())
+	if err != nil {
+		return nil, fmt.Errorf("invalid subtree signer name %q: %w", opts.subtreeSigner.Name(), err)
+	}
+
+	if err := checkOriginSignerName(opts.origin, opts.subtreeSigner.Name()); err != nil {
+		return nil, fmt.Errorf("checkOriginSignerName: %v", err)
+	}
```
roger2hk, line 436: This check could be moved to option validation in `opts.valid()`.

roger2hk (approved):

cmd/mtc/log/internal/checkpoint/reader.go
```diff
@@ -86,26 +96,57 @@ func (r *Reader) Checkpoint(ctx context.Context) ([]byte, error) {
...
+	)
+	return subtree{
+		start: start,
+		end:   end,
+		// signatures lazily retrieves and caches subtree signatures on first successful
+		// call. Errors are not cached to allow subsequent callers to retry.
+		//
+		// TODO: Consider using a detached or background context so that one caller's
+		// cancellation does not cancel in-flight signature fetching that other concurrent
+		// or future callers could benefit from.
```
roger2hk: Include transient signer failures (e.g. HSM/KMS timeout) handling in the TODO as well?

cmd/mtc/log/internal/checkpoint/reader.go
phbnf: I don't think we can do that because this defines a new subtree that has never been seen before, so its signature would has to be computed. Note that it shouldn't be too much of a problem since this signature will _only_ be computed if an `AddTBS()` requires it.

client/client.go
phbnf: I've moved this function inside isSubtreeValid where there's already a check. I expect that these functions will be removed once we use the t-dev/merkle ones.

cmd/mtc/log/mtc.go
phbnf: We can't put the `logCosignerID` over there, so I thought it would make sense to have them all together since they are about the same parameters. I've added a comment.

roger2hk (approved):

PR #1131, opened 2026-08-20
[MTC] Ensure earliest subtree in checkpoint.Reader always covers [0, start)

phbnf:
Towards #945.

In a followup PR, we'll cache subtree signatures in `Reader.subtrees`. Therefore, we also need to store a subtree starting at 0 to cache its signatures.

### DIFF
--- a/cmd/mtc/log/internal/checkpoint/reader.go
+++ b/cmd/mtc/log/internal/checkpoint/reader.go
@@ -100,10 +100,13 @@ func (r *Reader) Checkpoint(ctx context.Context) ([]byte, error) {
 }
 
 // pushSubtreeLocked adds a subtree and potentially removes old ones, capping the number of elements.
+// When capacity is exceeded, the oldest delta subtree is merged into subtrees[0], keeping subtrees[0]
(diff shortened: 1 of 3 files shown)


cmd/mtc/log/internal/checkpoint/reader.go
```diff
@@ -139,6 +141,7 @@ func (r *Reader) SubtreeForIndex(index uint64) (start, end uint64, ok bool) {
 		}
 	}
 
-	// If not found in retained subtrees, it must precede the earliest retained subtree.
-	return 0, r.subtrees[0].start, true
+	// Since subtrees[0] always covers [0, ...), any index < LatestSize() should
+	// have been matched in the loop above.
+	return 0, 0, false
```
roger2hk: These lines should be unreachable. Should this be replaced with a panic or error log?

roger2hk (approved):

cmd/mtc/log/internal/checkpoint/reader.go
phbnf: It now returns an error!

roger2hk (approved):

PR #1132, opened 2026-08-20
[MTC] Fetch subtree cosignatures and populate AddTBS responses with them.

phbnf:
Towards #945.

With this PR, an MTC log requests subtree cosignatures when processing `AddTBS()` requests, and inserts them accordingly in the `MTCProof`.

For the time being, only accept subtree cosignatures from witnesses that have also cosigned a checkpoint.

This design allows two things:
 - enforcing that mirrors MUST have signed a checkpoint before signing a subtree
 - allows to re-use checkpoint policy checking code, without re-implementing policy checking for subtree signatures

I expect that some of this implementation will change in the near future, as we iterate on the mirror client with request deduplication against multiple instances, and maybe providing a way to pass _all_ checkpoint signatures to the MTC log.

### DIFF
--- a/cmd/mtc/log/internal/subtreewitness/gateway.go
+++ b/cmd/mtc/log/internal/subtreewitness/gateway.go
@@ -1,0 +1,237 @@
+// Copyright 2026 The Tessera authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package subtreewitness provides a gateway for subtree cosigning with witnesses.
+package subtreewitness
+
+import (
+	"context"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"log/slog"
+	"maps"
+	"net/http"
+	"net/url"
+	"slices"
+	"sync"
+
+	f_note "github.com/transparency-dev/formats/note"
+	"github.com/transparency-dev/tessera"
+	"github.com/transparency-dev/tessera/cmd/mtc/log/internal/mtcproof"
+	wc "github.com/transparency-dev/witness/client/http"
+	"golang.org/x/mod/sumdb/note"
+)
+
+// ErrPolicyNotSatisfied is returned when witness responses do not satisfy the required policy.
+var ErrPolicyNotSatisfied = errors.New("witness policy was not satisfied")
+
+type witnessKey struct {
+	name    string
+	keyHash uint32
+}
+
+type witness struct {
+	client     SubtreeWitnessClient
+	verifier   f_note.SubtreeVerifier
+	cosignerID []byte
+}
+
+// SubtreeWitnessClient defines the interface for calling a witness's sign-subtree endpoint.
+type SubtreeWitnessClient interface {
+	SignSubtree(ctx context.Context, start, end uint64, subRoot []byte, proof [][]byte, rawCp []byte) ([]byte, error)
+}
+
+// Gateway manages concurrent requests to subtree witnesses and evaluates policy satisfaction.
+type Gateway struct {
+	witnesses map[witnessKey]witness
+	policy    tessera.WitnessGroup
+}
+
+// New creates a new subtree witness Gateway.
+// It only creates witness endpoints which implement f_note.SubtreeVerifier, i.e. which
+// use ML-DSA signatures.
+// SPEC: [DRAFT] Chrome Quantum-resistant Root Program Policy, Version 0.3.0, Section 3.1.
+// "Mirroring Cosigner Keys MUST be ML-DSA-44"
+func New(httpClient *http.Client, policy tessera.WitnessGroup) (*Gateway, error) {
+	if httpClient == nil {
+		httpClient = http.DefaultClient
+	}
+
+	witnesses := make(map[witnessKey]witness)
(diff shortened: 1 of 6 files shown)


cmd/mtc/log/internal/subtreewitness/gateway.go
```diff
@@ -0,0 +1,237 @@
...
+			}
+			keyHash := s.Hash
+			sigBytes := raw[4:]
+
+			k := witnessKey{name: s.Name, keyHash: keyHash}
+			// SPEC: draft-ietf-plants-merkle-tree-certs section 6.2.
+			// "An MTCProof parser MUST reject the input if there are duplicate cosigner_id values"
+			if _, ok := verifiedSubtreeSigs[k]; ok {
+				continue
+			}
```
roger2hk, line 197: The logic here verifies the witness keys instead of cosigner ID. Is the witness key always one-to-one mapped to the cosigner ID? If multiple keys are mapped to the same cosigner ID due to key rotation, the rejection check requirement from the specification seems to be wrong.

cmd/mtc/log/internal/subtreewitness/gateway.go
```diff
@@ -0,0 +1,237 @@
...
+// will be met once sufficient cosignatures from cosigners who have signed a corresponding
+// checkpoint have been collected.
+// TODO: implement subtree cosignature policy matching directly.
+func (gw *Gateway) CosignSubtree(ctx context.Context, origin string, start, end uint64, subRoot []byte, consProof [][]byte, rawCp []byte) ([]mtcproof.SubtreeSignature, error) {
+	if len(gw.witnesses) == 0 {
+		if gw.policy.Satisfied(rawCp) {
+			return nil, nil
+		}
+		return nil, ErrPolicyNotSatisfied
+	}
```
roger2hk, line 119: Is this part of the TODO? 

If a witness group consists of non-subtree verifiers, the `len(gw.witnesses)` is `0`, the logic will then go to return `nil, nil`. It is possible to bypass the policy.

cmd/mtc/log/internal/subtreewitness/gateway.go
```diff
@@ -0,0 +1,237 @@
...
+
+	// Open the checkpoint without verifying it to extract all signatures.
+	_, err := note.Open(rawCp, note.VerifierList())
+	var unverified *note.UnverifiedNoteError
+	if !errors.As(err, &unverified) {
+		return nil, fmt.Errorf("failed to parse checkpoint note: %v", err)
+	}
+	n := unverified.Note
+
+	// reconstructCp is used for policy checking.
```
roger2hk: Is it `reconstructedCp` or `reconstructCp`?

cmd/mtc/log/mtc.go
```diff
@@ -540,6 +564,15 @@ func (l *MTCLog) AddTBS(ctx context.Context, tbs TBSCertificateLogEntry) (*AddTB
...
 		return nil, fmt.Errorf("failed to get subtree signatures: %v", err)
 	}
+	if numSigs := len(subtreeSigs); numSigs < 2 {
+		// TODO: this is not strictly enforced for now, for development purposes. Consider
+		// enforcing it.
+		// SPEC: [DRAFT] Chrome Quantum-resistant Root Program Policy, Version 0.3.0, Section 2.4.5.
+		// "Standalone certificates MUST have at least 2 cosignatures. One of these
+		// MUST be from the MTC CA Operator, and one MUST be from a Mirroring
+		// Cosigner recognized by Chrome and not operated by the MTC CA Operator."
+		slog.WarnContext(ctx, "collected less than 2 subtree signatures", slog.Int("num_sigs", numSigs))
```
roger2hk:
```suggestion
		slog.WarnContext(ctx, "Collected less than 2 subtree signatures", slog.Int("num_sigs", numSigs))
```

cmd/mtc/log/internal/subtreewitness/gateway.go
```diff
@@ -0,0 +1,237 @@
...
+	"errors"
+	"fmt"
+	"log/slog"
+	"maps"
+	"net/http"
+	"net/url"
+	"slices"
+	"sync"
+
+	f_note "github.com/transparency-dev/formats/note"
```
roger2hk, line 30: In Go, package names must be concise and use only lowercase letters and numbers (e.g., k8s, oauth2). Multi-word package names should remain unbroken and in all lowercase (e.g., tabwriter instead of tabWriter, TabWriter, or tab_writer).

https://google.github.io/styleguide/go/decisions.html#package-names

cmd/mtc/log/internal/subtreewitness/gateway.go
phbnf: `s.Name()` is the cosigner ID, I've added a comment. We already verify that it is a valid cosigner ID on line 44.

cmd/mtc/log/internal/subtreewitness/gateway.go
phbnf: You are correct, and I think it is fine for now. I've added a TODO.

For now, it is the responsibility of log operators to set a valid policy for their operations with ML-DSA verifiers. 

Further down the line, I don't think it should be possible to start an MTC log with a policy that lists non-subtree verifiers. It shouldn't be possible to start one with a policy that doesn't  comply with Chrome requirements either. In the meantime, I've put TODOs in `main.go`. I don't really know where we'll enforce that, maybe this will be a special policy parser.  There's also a TODO and warning if not enough signatures have been collected.

cmd/mtc/log/internal/subtreewitness/gateway.go
phbnf: It's `reconstructedCp`, fixed, thanks.

cmd/mtc/log/internal/subtreewitness/gateway.go
phbnf: I'm following the pattern we've used in this repo so far, https://github.com/search?q=repo%3Atransparency-dev%2Ftessera+f_note&type=code. Maybe we can change this in a followup PR?

roger2hk (approved):

PR #1136, opened 2026-08-21
storage/posix: use actual checkpoint file modTime in tests

roger2hk:
Fixes test failures in `storage/posix/files_test.go` on high-performance environments by replacing `time.Now()` approximations with the actual checkpoint modification time via `s.stat(layout.CheckpointPath)`.

### DIFF
--- a/storage/posix/files_test.go
+++ b/storage/posix/files_test.go
@@ -336,11 +336,15 @@ func TestPublishTree(t *testing.T) {
 				s:           s,
 				entriesPath: opts.EntriesPath(),
 			}
-			pubAt := time.Now() // Good approximation of the checkpoint's future modTime.
 			appender, lr, err := s.newAppender(ctx, logStorage, opts)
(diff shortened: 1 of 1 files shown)


storage/posix/files_test.go
```diff
@@ -384,7 +388,11 @@ func TestPublishTree(t *testing.T) {
 				}
 				if !bytes.Equal(cpOld, cpNew) {
 					updatesSeen++
-					pubAt = time.Now()
+					cpInfo, err := s.stat(layout.CheckpointPath)
```
phbnf, line 391: nit: create a pubAt function that does all of this?

phbnf (approved):
Maybe

storage/posix/files_test.go
roger2hk: It's only 4 lines of code. It should be good to keep it as-is for now.

Issue #1152, opened 2026-09-07
Migrate to use formats/policy


AlCutter:
**Description**

Tessera used its own implementation & policy file for configuring witnesses (and later, mirrors). 

Now there's a common implementation in `formats/policy`, Tessera should be migrated to use this instead.

PR #1160, opened 2026-09-08
Reimplement Witness/WitnessGroup in terms of TLogPolicy

AlCutter:
This PR starts the process of migrating Tessera over to use the policy support implemented in `github.com/transparency-dev/formats/policy`.

The witness policy c'tors are re-implemented in `TLogPolicy`, and helper functions are provided to convert between the two representations.

Towards #1152


witness.go
```diff
@@ -38,158 +36,82 @@ type policyComponent interface {
...
+		for _, m := range g.Members {
+			if w, ok := witnesses[m]; ok {
+				members = append(members, w)
+			} else if grp, ok := groups[m]; ok {
+				members = append(members, grp)
+			} else {
+				return WitnessGroup{}, fmt.Errorf("invalid policy: member %q not defined", m)
+			}
+		}
+		wg := NewWitnessGroup(int(g.Threshold), members...)
```
roger2hk, line 96: The `NewWitnessGroup(n, children...)` panics if `n < 0 || n > len(children)`. Do we need any verification here and return error?

witness_test.go
```diff
@@ -212,3 +664,129 @@ func BenchmarkWitnessGroupSatisfaction(b *testing.B) {
...
+	v1, err := f_note.NewVerifierForCosignatureV1(vKey1)
+	if err != nil {
+		t.Fatalf("NewVerifierForCosignatureV1: %v", err)
+	}
+	v2, err := f_note.NewVerifierForCosignatureV1(vKey2)
+	if err != nil {
+		t.Fatalf("NewVerifierForCosignatureV1: %v", err)
+	}
+	w1 := Witness{vkey: vKey1, Key: v1}
+	w2 := Witness{vkey: vKey2, Key: v2}
```
roger2hk: Use `NewWitness` as `w1` and `w2` are instantiated without setting `witName`.

witness.go
```diff
@@ -253,11 +185,89 @@ func NewWitnessGroup(n int, children ...policyComponent) WitnessGroup {
...
+			if err != nil {
+				return "", fmt.Errorf("failed to parse witness URL %q: %w", c.URL, err)
+			}
+			if witName == "" {
+				witName = fmt.Sprintf("anonWit-%d", anonGroupNameCounter.Add(1))
+			}
+			alreadyAdded := false
+			for _, existing := range p.Witnesses {
+				if existing.Name == witName {
+					if existing.Verifier == c.Key && (existing.VKey == c.vkey || c.vkey == "") {
```
roger2hk: Should the comparison be using the key identity (i.e. key name, key hash) instead of the interface pointer?

witness.go
```diff
@@ -253,11 +185,89 @@ func NewWitnessGroup(n int, children ...policyComponent) WitnessGroup {
...
+						alreadyAdded = true
+						break
+					}
+					witName = fmt.Sprintf("%s-%d", witName, anonGroupNameCounter.Add(1))
```
roger2hk: By renaming the identical key to `-1`, one single witness signature would be treated as more than one quorum, potentially bypassing the quorum threshold.

witness.go
AlCutter: I could add the check, but that would have had to have been an invalid policy which somehow made it through parsing to get to here, in which case panic is probably reasonable?

witness.go
AlCutter: Good spot, although actually I think it's probably fine simplify and just check `vkey` for equality.

witness.go
AlCutter: This is where the key is different (although same name), but actually I think this is also only possible with an invalid policy, so I'll just disallow it.

witness_test.go
AlCutter: Done

roger2hk (approved):

witness.go
```diff
@@ -253,11 +189,88 @@ func NewWitnessGroup(n int, children ...policyComponent) WitnessGroup {
...
+func (wg WitnessGroup) ToPolicy() (policy.TLogPolicy, error) {
+	return wg.toPolicy()
+}
+
+func (wg WitnessGroup) toPolicy() (policy.TLogPolicy, error) {
+	if wg.N == 0 || len(wg.Components) == 0 {
+		return policy.TLogPolicy{
+			Quorum: "none",
+		}, nil
+	}
```
roger2hk: Is there a chance that we need to return optional groups when `N = 0`?

witness.go
```diff
@@ -253,11 +189,88 @@ func NewWitnessGroup(n int, children ...policyComponent) WitnessGroup {
...
+	return grpName, nil
+}
+
+// ToPolicy converts a [WitnessGroup] to a [policy.TLogPolicy].
+func (wg WitnessGroup) ToPolicy() (policy.TLogPolicy, error) {
+	return wg.toPolicy()
+}
+
+func (wg WitnessGroup) toPolicy() (policy.TLogPolicy, error) {
+	if wg.N == 0 || len(wg.Components) == 0 {
```
roger2hk: Should we return an error for `WitnessGroup{N: 1, Components: nil}`?

witness.go
AlCutter: Good spot, updated.

witness.go
AlCutter: Sure, probably overkill given this is all being deprecated, but why not :)

witness.go
```diff
@@ -253,11 +189,91 @@ func NewWitnessGroup(n int, children ...policyComponent) WitnessGroup {
...
 	return WitnessGroup{
+		grpName:    fmt.Sprintf("anonGrp-%d", anonGroupNameCounter.Add(1)),
 		Components: children,
 		N:          n,
 	}
 }
 
+// populatePolicy recursively populates a policy.TLogPolicy from a WitnessGroup.
+// It returns the name of the group and an error if any part of the population fails.
+func populatePolicy(p *policy.TLogPolicy, wg WitnessGroup) (string, error) {
```
roger2hk, line 200: The logic here doesn't check the duplicate subgroup name, so it is possible to have the same name appearing more than once in `p.Groups`. Is that allowed?

```go
sub := NewWitnessGroup(1, w1, w2);
root := NewWitnessGroup(2, sub, sub)
```

witness.go
```diff
@@ -253,11 +189,91 @@ func NewWitnessGroup(n int, children ...policyComponent) WitnessGroup {
...
+			if witName == "" {
+				return "", fmt.Errorf("witness with vkey %q has no name", c.vkey)
+			}
+			u, err := url.Parse(c.URL)
+			if err != nil {
+				return "", fmt.Errorf("failed to parse witness URL %q: %w", c.URL, err)
+			}
+			alreadyAdded := false
+			for _, existing := range p.Witnesses {
+				if existing.Name == witName {
```
roger2hk, line 227: The current logic doesn't return an error when multiple witnesses use the different name but the same public key. Although this sounds impossible, it's a violation in the tlog-policy specification.

https://github.com/C2SP/C2SP/blob/main/tlog-policy.md#public-key-representation

witness.go
AlCutter: That's not allowed, added a check for that.

This code is all rather temporary though, I'm trying to deprecate and remove all this code which operates in terms of `NewWitness` and `NewWitnessGroup`.

PR #1162, opened 2026-09-08
Migrate append_lifecycle to TLogPolicy

AlCutter:
This PR migrates the append lifecycle code to use the `formats/policy` package.

Towards #1152

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ and
 and performs the following steps:
   1. Create a new Checkpoint and sign it with the signer provided by [WithCheckpointSigner](https://pkg.go.dev/github.com/transparency-dev/tessera#AppendOptions.WithCheckpointSigner)
   2. Contact witnesses and collect enough cosignatures to satisfy any witness policy configured by [WithWitnessPolicy](https://pkg.go.dev/github.com/transparency-dev/tessera#AppendOptions.WithWitnessPolicy)
-  3. If the witness policy is satisfied, make this new Checkpoint public available
+  3. If the witness policy is satisfied, make this new Checkpoint publicly available
 
 An entry is considered published once it is committed to by a published Checkpoint (i.e. a published Checkpoint's size is larger than the entry's assigned index).
 Due to the nature of append-only logs, all Checkpoints issued after this point will also commit to inclusion of this entry.
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ and
 [WithCheckpointRepublishInterval](https://pkg.go.dev/github.com/transparency-dev/tessera#AppendOptions.WithCheckpointRepublishInterval))
 and performs the following steps:
   1. Create a new Checkpoint and sign it with the signer provided by [WithCheckpointSigner](https://pkg.go.dev/github.com/transparency-dev/tessera#AppendOptions.WithCheckpointSigner)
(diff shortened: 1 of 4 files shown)


append_lifecycle.go
```diff
@@ -850,14 +837,13 @@ func (o AppendOptions) CheckpointPublisherContext(ctx context.Context, lr LogRea
...
 func (o AppendOptions) witnessGateway(ctx context.Context, lr LogReader, httpClient *http.Client) (*witness.WitnessGateway, error) {
 	witnesses := []witness.Witness{}
-	for uStr, vs := range o.witnesses.WitnessEndpoints() {
-		u, err := url.Parse(uStr)
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse witness URL: %w", err)
+	for _, w := range o.witnessPolicy.Witnesses {
+		if w.URL == nil {
+			return nil, fmt.Errorf("invalid witness policy: witness %q has no URL", w.Name)
 		}
```
roger2hk, line 843: Should we also add the `nil` check for `w.Verifier` just in case it is `nil`?

```go
if w.Verifier == nil {
    return nil, fmt.Errorf("invalid witness policy: witness %q has no verifier", w.Name)
}
```

append_lifecycle.go
roger2hk, line 857: Should we return an error here if `o.witnessPolicy.Quorum` is not `none` but the `len(witnesses)` is `0`?

roger2hk (approved):

README.md
```diff
@@ -139,7 +139,7 @@ and
 [WithCheckpointRepublishInterval](https://pkg.go.dev/github.com/transparency-dev/tessera#AppendOptions.WithCheckpointRepublishInterval))
 and performs the following steps:
   1. Create a new Checkpoint and sign it with the signer provided by [WithCheckpointSigner](https://pkg.go.dev/github.com/transparency-dev/tessera#AppendOptions.WithCheckpointSigner)
-  2. Contact witnesses and collect enough cosignatures to satisfy any witness policy configured by [WithWitnesses](https://pkg.go.dev/github.com/transparency-dev/tessera#AppendOptions.WithWitnesses)
+  2. Contact witnesses and collect enough cosignatures to satisfy any witness policy configured by [WithWitnessPolicy](https://pkg.go.dev/github.com/transparency-dev/tessera#AppendOptions.WithWitnessPolicy)
   3. If the witness policy is satisfied, make this new Checkpoint public available
```
roger2hk:
```suggestion
  3. If the witness policy is satisfied, make this new Checkpoint publicly available
```

append_lifecycle_test.go
```diff
@@ -930,3 +975,28 @@ func mustNewWitness(t *testing.T, vkey, urlStr string) Witness {
 	}
 	return wit
 }
+
+type keyUrl struct {
```
roger2hk: Use `keyURL`.

append_lifecycle.go
AlCutter: Sure, done.

append_lifecycle.go
AlCutter: Sure, added, but just wanted to point out that this (URL) one is required because URLs are technically [optional in tlog-policy](https://c2sp.org/tlog-policy#defining-a-witness) (although, obviously, not for a policy used by a log when signing checkpoints), so we could end up here even though the parser succeeded.

PR #1180, opened 2026-09-16
[POSIX] Fix incorrect mirror partial tile reconstruction

### LINKED ISSUE #1178: [POSIX] MirrorWriter publishes truncated tiles/bundles to full-tile/bundle paths
**Description**

`MirrorWriter.UpdateCheckpoint` can write an under-sized hash tile to a *full* tile/bundle
path. Example: integrate 66000 entries, then publish a checkpoint of size 65537:

```
tile/1/001    32 bytes   # 1 node, but a full tile must have 256
```

https://github.com/transparency-dev/tessera/blob/cf6ceed1640f0a79a76d7ea8685f64ed4c73940d/storage/posix/files.go#L1397-L1410

**Reproduction Example**
For `cpSize = 65537` and `treeSize = 66000`:

At `l = 1, idx = 1`: `treeP = 1, cpP = 0`.
The tile has 0 nodes at `cpSize`, but `ensurePartialTile` reads the 1-node partial tile (`treeP = 1`) and writes it untrimmed to `tile/1/001` (the full-tile path).

**Impact**
- Full-tile paths are immutable and CDN-cacheable. Clients fetching full tiles receive corrupt data, breaking Merkle proof verification.
- Subsequent integration up to full tile capacity cannot overwrite the file via `createIdempotent`.

https://github.com/transparency-dev/tessera/blob/cf6ceed1640f0a79a76d7ea8685f64ed4c73940d/storage/posix/files.go#L1512-L1514

AlCutter:
This PR addresses an issue with the mirror writer lifecycle implementation in the POSIX driver where it would create an invalid out-of-bounds tile under certain circumstances.

Fixes #1178


storage/posix/files.go
roger2hk, line 1404: What if you short-circuit earlier here?

```suggestion
cpP := layout.PartialTileSize(l, idx, cpSize)
if cpP == 0 {
	continue
}
treeP := layout.PartialTileSize(l, idx, treeSize)
```

storage/posix/files_test.go
```diff
@@ -864,14 +849,38 @@ func TestMirrorWriter_UpdateCheckpointGeometry(t *testing.T) {
...
+					// haven't incorrectly created one.
+					if fi, err := os.Stat(filepath.Join(s.cfg.Path, fullTilePath)); err == nil {
+						t.Fatalf("Level %d tile %d: full tile %s should not exist for size %d (found %d bytes)", level, idx, fullTilePath, test.size, fi.Size())
+					} else if !errors.Is(err, os.ErrNotExist) {
+						t.Fatalf("Failed to stat %s: %v", fullTilePath, err)
+					}
+				} else {
+					tile := mustReadTile(t, lr, uint64(level), idx, uint8(wantNodes))
+					if len(tile.Nodes) != wantNodes {
+						t.Errorf("Expected %d nodes in partial tile %d/%d.%d, got %d", wantNodes, level, idx, wantNodes, len(tile.Nodes))
```
roger2hk: Should the warning message use `uint8(wantNodes)` instead of `wantNodes` in case it's the full tile? Otherwise, it may show `.256`.

storage/posix/files.go
```diff
@@ -1388,7 +1388,7 @@ func (m *MirrorWriter) UpdateCheckpoint(ctx context.Context, fn func(old []byte)
 // If an implied partial resource is not already present, this function will attempt to create
 // it from a strictly larger resource whose presence is implied by treeSize.
```
roger2hk, line 1389: Should we add one more line to the code comment to include the new change?

"When cpSize == treeSize the resources are guaranteed present by integration, so there's nothing to do."

storage/posix/files_test.go
```diff
@@ -928,3 +937,22 @@ func BenchmarkMarshalTlogEntryBundle(b *testing.B) {
 	}
 }
 
+func generateBundles(t *testing.T, totalEntries int) iter.Seq2[*api.EntryBundle, error] {
+	t.Helper()
+
+	return func(yield func(*api.EntryBundle, error) bool) {
+		remaining := totalEntries
+		dummyEntry := []byte("a")
```
roger2hk: nit: Using `[]byte("a")` doesn't try to catch whether there is any offset. Consider using something like `fmt.Appendf(nil, "entry %d", base+i)`?

roger2hk (approved):

storage/posix/files.go
AlCutter: I think that's implied on the method comment, but I've added it as an in-line comment on the condition check.

PR #1183, opened 2026-09-17
Bump formats & witness, along with fixups

AlCutter:
This PR:
- bumps the `formats` and `witness` deps to bring in recent changes to `SignSubtree`.
- updates the locations where the MTC code interacts with `SignSubtree`.


cmd/mtc/log/internal/mtcproof/mtcproof.go
```diff
@@ -58,15 +60,37 @@ type SubtreeSignature struct {
...
+	cID, err := ParseCosignerID(name)
+	if err != nil {
+		return SubtreeSignature{}, fmt.Errorf("invalid cosigner ID: %v", err)
+	}
+
 	return SubtreeSignature{
-		CosignerID: cosignerID,
-		Signature:  s,
+		CosignerID: cID,
+		Signature:  sigRaw[8:],
```
phbnf: Maybe add a comment to explicitly say we're stripping the timestamp for now.

phbnf (approved):

cmd/mtc/log/internal/mtcproof/mtcproof.go
AlCutter: There's no change from the old behaviour in that respect, so I think the function comment on L52 probably does the job?

I have made the discarding of the timestamp more obvious, though.

PR #1188, opened 2026-09-18
Tighten queue cancel

AlCutter:
This PR tightens up the queue semantics when it's being closed down.

This should help avoid any deadlocks or ambiguity around whether leaves sent via `Add()` were accepted or errored.


storage/internal/queue.go
```diff
@@ -49,7 +52,9 @@ type FlushFunc func(ctx context.Context, entries []*tessera.Entry) error
 // for maxAge, or the size of the queue reaches maxSize.
 func NewQueue(ctx context.Context, maxAge time.Duration, maxSize uint, f FlushFunc) *Queue {
 	q := &Queue{
-		inputs: make(chan queueItem, maxSize),
+		queueCtx: ctx,
+		// inputs is unbuffered so that Add() returns immediately when the queue is shutting down.
+		inputs: make(chan queueItem),
```
roger2hk: The benchmark shows +47% throughput regression after removing the buffer from `inputs`. What if we keep the buffer and drain the `inputs`?

storage/internal/queue.go
```diff
@@ -58,12 +63,21 @@ func NewQueue(ctx context.Context, maxAge time.Duration, maxSize uint, f FlushFu
...
 
-		var items []queueItem
+		items := make([]queueItem, 0, maxSize)
 
 		// Initialise a timer with an arbitrarily large value, and immediately
 		// call Stop() to avoid a spurious trigger on the first iteration.
 		timer := time.NewTimer(time.Hour)
 		timer.Stop()
+		// Ensure timer is stopped once ctx is done and we return.
+		defer timer.Stop()
```
roger2hk: nit: This seems to be a no-op since Go 1.23.

https://go.dev/doc/go1.23#timer-changes

storage/internal/queue.go
```diff
@@ -133,6 +143,10 @@ func (q *Queue) Add(ctx context.Context, e *tessera.Entry) tessera.IndexFuture {
 		return func() (tessera.Index, error) {
 			return tessera.Index{}, ctx.Err()
 		}
+	case <-q.queueCtx.Done():
+		return func() (tessera.Index, error) {
+			return tessera.Index{}, q.queueCtx.Err()
+		}
```
roger2hk, line 147: If `q.queueCtx` is cancelled but the batching goroutine is still active, it's still possible to get an entry into the shutting-down queue.

storage/internal/queue_test.go
```diff
@@ -154,6 +154,76 @@ func TestNotify(t *testing.T) {
...
+		{
+			name:      "partial batch waiting in items",
+			maxSize:   10,
+			numItems:  5,
+			blockWork: false,
+		},
+		{
+			name:      "worker blocked in flush and batches full",
+			maxSize:   2,
+			numItems:  2 + 2 + 2 + 1, // 2 in doFlush (flushFunc below), 2 in batches chan, 2 blocked in flush() (on batches<-), 1 blocked in Add() (on inputs<-).
```
roger2hk: The test here doesn't fail (not flaky) but it is not always the same as what it describes.

The `a.Add` on L202 may not finish adding the entry when `cancel()` on L212 is called.

```go
addWg.Go(func() {
	futures[i] = q.Add(t.Context(), tessera.NewEntry(fmt.Appendf(nil, "item %d", i)))
})
```

roger2hk (approved):

storage/internal/queue.go
AlCutter: That's true, but only if the batching goroutine is in `<-q.inputs` on L111 (`inputs` is not buffered), so either that's impossible because it already detected the ctx cancelled, or as soon as it processes the recv and goes around the loop it'll detect the cancelled context and call `cancelItems()`.

storage/internal/queue_test.go
AlCutter: Gnash, gnash. Ok should be better now.

storage/internal/queue.go
AlCutter: Yeah, you're right - I was worried about a spurious tick but we have to have come out of here for the defer to work anyway.

Removed!

PR #1190, opened 2026-09-18
Tighten client nodecache

AlCutter:
nodeCache was very relaxed in terms of the node IDs it
would serve and the tiles it would accept from a fetcher.

This PR tightens both, and removes a potential correctness issue:
* GetNode explicitly rejects out-of-range nodes.
* fetchTileNodes now asserts that a tile is either the precise partial
  size requested, or a full tile (i.e. the fetcher fell-back to
  requesting a full tile)
* Removed call to GetRootHash. This was unnecessary, but
  also incorrectly populated the cache with ephemeral nodes.


client/client.go
```diff
@@ -528,7 +531,22 @@ func (n *nodeCache) fetchTileNodes(ctx context.Context, tileLevel, tileIndex uin
...
+		}
+		switch gotLen := len(tile.Nodes); gotLen {
+		case wantSize:
+			// We got the exact size tile we asked for, nothing extra to do.
+		case layout.TileWidth:
+			// Must have asked for a partial tile and got a full tile back.
+			// Trim the full tile down to the size of the requested partial tile.
+			tile.Nodes = tile.Nodes[:wantSize]
+		default:
+			return nil, fmt.Errorf("invalid tile: expected %d or %d nodes, got %d", wantSize, layout.TileWidth, gotLen)
```
roger2hk, line 546: If `wantSize` is `256` and `p = 0`, the error message would become:

```
invalid tile: expected 256 or 256 nodes, got 255
```

Having `256` twice seems confusing.

client/client.go
roger2hk, line 526: The change in L489 is ineffective because the `os.ErrNotExist` is not returned here.

```suggestion
			return nil, fmt.Errorf("failed to fetch tile: %w", err)
```

client/client_test.go
```diff
@@ -198,28 +198,100 @@ func TestCheckLogStateTracker(t *testing.T) {
 	}
 }
 
-func TestNodeCacheHandlesInvalidRequest(t *testing.T) {
+func TestNodeCache(t *testing.T) {
 	ctx := context.Background()
```
roger2hk, line 202:
```suggestion
	ctx := t.context()
```

roger2hk (approved):

PR #1191, opened 2026-09-18
Fix mirror padding pickup

AlCutter:
This PR fixes a situation where the mirror lifecycle could miscalculate the size of the partial bundle required to left-pad the first chunk of an add-entries request.


mirror_lifecycle.go
```diff
@@ -291,33 +291,41 @@ func (mt *MirrorTarget) AddEntries(ctx context.Context, uploadStart, uploadEnd u
...
+		if startPad := start % layout.EntryBundleWidth; startPad != 0 {
+			bIdx := start / layout.EntryBundleWidth
+			// The tlog-mirror spec allows an upload to start some way below the current tree size, so the
+			// stored bundle we need to read could well contain more entries than we're going to use for padding.
+			// Figure out which size of tile we can fetch, given the size of the _actual_ tree, and then fetch that
+			// and trim as necessary.
+			storedPartial := layout.PartialTileSize(0, bIdx, integratedSize)
+			br, err := mt.reader.ReadEntryBundle(ctx, bIdx, storedPartial)
 			if err != nil {
 				yield(nil, fmt.Errorf("failed to read bundle containing uploadStart (%d): %v", start, err))
```
roger2hk: (Not related to this pull request but while we're here)

Use `%w` to return `os.ErrNotExist`.

```suggestion
				yield(nil, fmt.Errorf("failed to read bundle containing uploadStart (%d): %w", start, err))
```

mirror_lifecycle.go
```diff
@@ -291,33 +291,41 @@ func (mt *MirrorTarget) AddEntries(ctx context.Context, uploadStart, uploadEnd u
...
+			// Parse and clip, since the bundle we read may contain more entries than we need for padding.
 			b := &api.EntryBundle{}
 			if err := b.UnmarshalText(br); err != nil {
 				yield(nil, fmt.Errorf("failed to unmarshal bundle containing uploadStart (%d): %v", start, err))
 				return
 			}
-			if l := len(b.Entries); l < int(p) {
-				yield(nil, fmt.Errorf("POTENTIAL CORRUPTION: partial bundle at index %d.%d has only %d entries", start/layout.EntryBundleWidth, p, l))
+			if l := len(b.Entries); l < int(startPad) {
+				yield(nil, fmt.Errorf("POTENTIAL CORRUPTION: bundle at index %d (requested p.%d) has only %d entries, want at least %d", bIdx, storedPartial, l, startPad))
```
roger2hk: If `storedPartial` is `0`, the error message would be misleading (i.e. `(requested p.0)`).

mirror_lifecycle.go
```diff
@@ -291,33 +291,41 @@ func (mt *MirrorTarget) AddEntries(ctx context.Context, uploadStart, uploadEnd u
 //
 // Yielded entry bundles are always aligned to bundle boundaries. Specifically, this means that if the provided start is _not_ bundle aligned, then we will
 // fetch entries from the bundle at start/256 and use those entries to left-pad the first yielded bundle.
-func (mt *MirrorTarget) bundleIterator(ctx context.Context, next func() (*MirrorPackage, error), start uint64, pendingCP *log.Checkpoint) func(func(*api.EntryBundle, error) bool) {
+func (mt *MirrorTarget) bundleIterator(ctx context.Context, next func() (*MirrorPackage, error), start, integratedSize uint64, pendingCP *log.Checkpoint) func(func(*api.EntryBundle, error) bool) {
```
roger2hk: The `integratedSize` seems to be exactly the same as `mt.reader.IntegratedSize(ctx)`.

roger2hk (approved):

mirror_lifecycle.go
AlCutter: Updated.

mirror_lifecycle.go
AlCutter: That is where the value `integratedSize` holds comes from, we could change to make another call to `IntegratedSize()` here instead of passing the value we already have through, but that's another potential RPC or read from storage - is there an issue you have in mind which would make that preferable to just using the value we already have?

roger2hk: It was just for readability. LGTM.

roger2hk (approved):

